PR #32, opened 2018-09-10
Add WebView Flow types

empyrical:
This pull request adds Flow proptypes for the iOS and Android WebViews based off of the React PropTypes in the original repo.

`.flowconfig` files for iOS and Android have been added, and two scripts have been added to package.json: `flow` and `flow-android`, for running `flow check` against the two platforms. The only difference between the two configs are the ignored suffixes at the very top. When Win10's WebView is added, a new flowconfig will need to be created for it as well.

The React Native version has also been bumped as well, there are a few differences (like missing annotations in some files in 0.56 that are now present in 0.57 that were causing `check` to fail) that made having a flowconfig that supported both a bit awkward.

In the original components' PropTypes, there were some Android-only props that were in iOS' PropTypes and vice-versa. Only Android had a few PropTypes that weren't in iOS', and I pretty much mirrored how they were originally. Should the platform-specific types be split up further?


js/WKWebView.ios.js
```diff
@@ -15,14 +15,20 @@ import { requireNativeComponent } from 'react-native';
 
 const RCTWKWebView = requireNativeComponent('RCTWKWebView');
 
-class WKWebView extends React.Component {
-  componentWillReceiveProps(nextProps) {
+type RCTWKWebViewProps = {
```
TheSavior: Types, especially for Props should try to be read only and exact:

```
type RCTWKWebViewProps = $ReadOnly<{|
  allowsInlineMediaPlayback?: boolean,
  mediaPlaybackRequiresUserAction?: boolean,
  dataDetectorTypes?: boolean,
|}>;
```
This keeps people from accidentally passing in additional props or misspelling props. It also ensures that the component implementation doesn't do things like `this.props.foo = 4;`

js/WKWebView.ios.js
```diff
@@ -15,14 +15,20 @@ import { requireNativeComponent } from 'react-native';
...
+  allowsInlineMediaPlayback?: boolean,
```
TheSavior: Something we've started doing more of (but are still not at 100% on) internally is making sure that optional props are also nullable:

```
  allowsInlineMediaPlayback?: ?boolean,
```

The reason for this is so that you can have a parent component which can spread in it's props:

```
const MyWrappedComponent(props: RCTWKWebViewProps) => (
   <WKWebView {...props} />
)
```

The above will error with the types you have right now because it might be passing the prop but setting it to null having come from the wrapper. If you make the webview props both optional and nullable then this will work fine.

js/WebView.android.js
```diff
@@ -41,10 +49,52 @@ const defaultRenderLoading = () => (
...
 
+type State = {|
+  viewState: WebViewState,
+  lastErrorEvent: ?WebViewErrorEvent,
+  startInLoadingState: boolean,
+|};
+
+type WebViewPropsAndroid = $ReadOnly<{|
+  ...WebViewSharedProps,
+  onNavigationStateChange?: (event: WebViewEvent) => any,
```
TheSavior: The return type of the callback should be `mixed` instead of `any`: https://flow.org/en/docs/types/functions/#toc-function-type

js/WebView.android.js
```diff
@@ -55,7 +105,7 @@ class WebView extends React.Component {
 
   state = {
     viewState: WebViewState.IDLE,
-    lastErrorEvent: null,
+    lastErrorEvent: (null: ?WebViewErrorEvent),
```
TheSavior: I'm surprised this type annotation is necessary

js/WebView.android.js
```diff
@@ -94,11 +145,11 @@ class WebView extends React.Component {
       webViewStyles.push(styles.hidden);
     }
 
-    const source = this.props.source || {};
-    if (this.props.html) {
-      source.html = this.props.html;
-    } else if (this.props.url) {
-      source.uri = this.props.url;
+    let source = this.props.source || ({}: WebViewSource);
```
TheSavior: Thoughts on:

```
let source : WebViewSource = this.props.source || {};
```

js/WebView.ios.js
```diff
@@ -442,16 +447,8 @@ class WebView extends React.Component {
   }
 }
 
-const RCTWebView = requireNativeComponent(
-  'RCTWebView',
-  WebView,
-  WebView.extraNativeComponentConfig,
```
TheSavior: I don't know where these were being specified but you can probably delete these too.

js/WebViewTypes.js
```diff
@@ -0,0 +1,332 @@
...
+import type {ViewStyleProp} from 'StyleSheet';
+import type {ViewProps} from 'ViewPropTypes';
+
+export type WebViewErrorEvent = {
+  domain: any,
+  code: any,
+  description: any,
+};
+
+export type WebViewEvent = Object;
```
TheSavior: It would be great to get this typed more accurately

js/WebViewTypes.js
```diff
@@ -0,0 +1,332 @@
...
+ */
+
+'use strict';
+
+import type {Node, Element} from 'react';
+import type {EdgeInsetsProp} from 'EdgeInsetsPropType';
+import type {ViewStyleProp} from 'StyleSheet';
+import type {ViewProps} from 'ViewPropTypes';
+
+export type WebViewErrorEvent = {
```
TheSavior: ReadOnly and exact

js/WebViewTypes.js
```diff
@@ -0,0 +1,332 @@
...
+  | 'link'
+  | 'address'
+  | 'calendarEvent'
+  | 'trackingNumber'
+  | 'flightNumber'
+  | 'lookupSuggestion'
+  | 'none'
+  | 'all';
+
+export type WebViewSourceUri = {|
```
TheSavior: ReadOnly too

js/WebViewTypes.js
```diff
@@ -0,0 +1,332 @@
...
+  /**
+   * The HTTP body to send with the request. This must be a valid
+   * UTF-8 string, and will be sent exactly as specified, with no
+   * additional encoding (e.g. URL-escaping or base64) applied.
+   * NOTE: On Android, this can only be used with POST requests.
+   */
+  body?: string,
+|};
+
+export type WebViewSourceHtml = {|
```
TheSavior: ReadOnly

js/WebViewTypes.js
```diff
@@ -0,0 +1,332 @@
...
+  baseUrl?: ?string,
+|};
+
+export type WebViewSource = WebViewSourceUri | WebViewSourceHtml;
+
+export type WebViewNativeConfig = $ReadOnly<{|
+  /*
+   * The native component used to render the WebView.
+   */
+  component?: ?any,
```
TheSavior: A component that is passed in appears that it must also take the same props as the existing WebView, so this should probably be something like
```
React.ComponentType<WebViewSharedProps>
```

js/WebViewTypes.js
```diff
@@ -0,0 +1,332 @@
...
+
+  /**
+   * Function that returns a loading indicator.
+   */
+  renderLoading: () => Element<any>,
+
+  /**
+   * Function that is invoked when the `WebView` has finished loading.
+   */
+  onLoad: (event: WebViewEvent) => any,
```
TheSavior: These should all be mixed

TheSavior:
I'm not exactly sure what you mean by 
>In the original components' PropTypes, there were some Android-only props that were in iOS' PropTypes and vice-versa. Only Android had a few PropTypes that weren't in iOS', and I pretty much mirrored how they were originally. Should the platform-specific types be split up further?

but what you have right now seems reasonable if both ios and android pass Flow.

empyrical:
> I'm not exactly sure what you mean

I mean, should the iOS only props go to a `WebViewPropsIOS` type, and the rest of the Android only props go to `WebViewPropsAndroid`?

js/WebView.android.js
empyrical: That works, and looks a lot better too!

js/WebView.android.js
empyrical: I copied that over from the main repo - flow passes without it, so I removed it

js/WebViewTypes.js
empyrical: Had a look at the native code and made a more exact type looking at Android's `createWebViewEvent` and iOS RCTWebView's `baseEvent`

empyrical:
Cleaned things up a little and filled out the error/event types. I also notice looking at the native code on the android side that `navigationType` doesn't seem to be implemented yet, may be worth creating an issue for after this PR is done.

package.json
```diff
@@ -5,12 +5,20 @@
...
   "version": "0.1.0",
   "homepage": "https://github.com/react-native-community/react-native-webview#readme",
+  "scripts": {
+    "flow": "flow check",
+    "flow-android": "flow check --flowconfig-name .flowconfig.android"
+  },
   "peerDependencies": {
     "react": "^16.0",
-    "react-native": "^0.56"
+    "react-native": "~0.57.0-rc.4"
```
jamonholmgren: 0.56 does work with this repo, is there a reason to lock to rc 4?

empyrical: I will see if I can make `flow` pass on 0.56

jamonholmgren: If it's a lot of work, we can require 0.57+.

empyrical:
Re-organized the platform-specific types into their own types and applied them to `WebViewSharedTypes`.

Also added a note about the types to the readme.

package.json
empyrical: So I discovered why I was having so much trouble with 0.56's JS files, as opposed to 0.57's: 0.56 was tested against Flow 0.75, and [Flow 0.78 introduced](https://github.com/facebook/flow/blob/master/Changelog.md#0780) the `--flowconfig-name` flag which allowed flow to use the `.flowconfig.android` file in the main repo and test the flow files - previously, it seems flow was not ran against android platform files.

0.57, now being tested against Flow 0.80, has a bunch of comments like this in the Android files:

```js
/* $FlowFixMe(>=0.78.0 site=react_native_android_fb) This issue was found when
 * making Flow check .android.js files. */
```

Without them, on 0.56, I get all these errors:

https://gist.github.com/empyrical/70e7cceb87c6e8f89c5f01c120d4d9d3

So I suppose if you want to support 0.56 on this repo, Flow could only be ran against 0.57+ but other tests could be ran on 0.56

jamonholmgren: I'd like to support 0.56 if we can -- I'm totally fine if Flow is not run against 0.56 at all and we just manually test it.

empyrical: I'll bump the version back down to 0.56, and add a note that flow will currently only pass against 0.57

empyrical:
Restored the min. React Native version to 0.56

Titozzz:
Just to be sure, as I'm not flow expert (at all) @TheSavior is it all good ?

TheSavior:
Looks good to me

jamonholmgren:
Thank you @empyrical, awesome work! ❤️

PR #34, opened 2018-09-11
try reuse Android tooling config from React Native

dulmandakh:
Try to use Android tooling config from React Native

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -18,5 +18,5 @@ android {
 }
 
 dependencies {
-    compile 'com.facebook.react:react-native:+'
+    implementation 'com.facebook.react:react-native:+'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -18,5 +18,5 @@ android {
 }
 
 dependencies {
-    implementation 'com.facebook.react:react-native:+'
+    compile 'com.facebook.react:react-native:+'
 }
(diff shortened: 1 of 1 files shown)


jamonholmgren:
Thanks for submitting this, @dulmandakh ! Can you explain more what this PR will change and what it affects?

dulmandakh:
It'll try to use tools and their versions that is used by React Native project or app, which landed with https://github.com/facebook/react-native/commit/0a3055d98a36e49746144e883edc7e20afec4fcb. So the advantage is you don't have to bump tools versions every time RN bumps them, also users don't have to download other versions of Android tools just to compile this module. 

For example, RN master landed SDK 27 and uses build tools 27.0.3 with https://github.com/facebook/react-native/commit/044b399e6590d84065a9b186750f77bc9d851aac. If we don't have this configuration RN developers have to download SDK 27 for RN and SDK 26 for this module.

Titozzz:
Should we default to 27 ? what do you think @dulmandakh ?

This PR is indeed needed and we will merge it :)

dulmandakh:
RN 0.57 use SDK 26, so it's safe to use it as default. SDK 27 landed in master, and will be a part of RN 0.58.

Titozzz:
If you check there: https://github.com/react-native-community/react-native-releases/pull/40/files it says it will be 27 for 0.57, is this a typo in the changelog?

jamonholmgren:
Ready to merge if @Titozzz gives approval.

jamonholmgren:
By the way, I'm okay dropping support for 0.56 if it's causing issues, @Titozzz .

PR #42, opened 2018-09-13
Import/0c576ef allow file access android

Titozzz:
https://github.com/facebook/react-native/commit/0c576ef84a1c7f79b228f205cc687ab1b945bda1


Titozzz:
Ran flow check

PR #54, opened 2018-09-19
[FEATURE]: Added Android overscroll property

Titozzz:


Titozzz:
Should this be unified with https://facebook.github.io/react-native/docs/webview#bounces?

js/WebView.android.js
```diff
@@ -61,6 +61,7 @@ type State = {|
  */
 class WebView extends React.Component<WebViewSharedProps, State> {
   static defaultProps = {
+    overScrollMode: 0,
```
jamonholmgren: Let's add some constants for these modes:

```
OVER_SCROLL_ALWAYS = 0
OVER_SCROLL_CONTENT = 1
OVER_SCROLL_NEVER = 2
```

So then this would be something like

```js
static OVER_SCROLL_ALWAYS = 0
static OVER_SCROLL_CONTENT = 1
static OVER_SCROLL_NEVER = 2
static defaultProps = {
  overScrollMode: OVER_SCROLL_ALWAYS,
  // ...
```

Titozzz: Do you want to expose thoses values to the user ? or are you just trying to make the internal code clearer?

js/WebViewTypes.js
```diff
@@ -218,6 +218,16 @@ export type AndroidWebViewProps = $ReadOnly<{|
...
 
+  /**
+   * Sets the overScrollMode. Values can be found at
+   * https://developer.android.com/reference/android/view/View#OVER_SCROLL_NEVER
+   * 0 = always (default)
+   * 1 = if content scrolls
+   * 2 = never
+   * @platform android
+   */
+  overScrollMode?: ?number,
```
empyrical: Flow type looks good! :+1:

js/WebView.android.js
jamonholmgren: Good question -- I feel we need them either way, but exposing to the user seems like a good idea. Your thoughts?

js/WebViewTypes.js
empyrical: Actually - perhaps this would be better:

    overScrollMode?: 0 | 1 | 2,

js/WebView.android.js
empyrical: What if it was changed to use a string, rather than a number? Like:

```js
    overScrollMode?: 'always' | 'content' | 'never',
```

And on the android side, it converted the strings back to their appropriate numbers

jamonholmgren: I really like that idea @empyrical, and it would fit with some other properties, like this one:

https://github.com/react-native-community/react-native-webview/blob/e70a18466e0fd76a042795d7e76fe725a98006a6/js/WebViewTypes.js#L30

Titozzz:
Updated to string, WDYT?

js/WebView.android.js
```diff
@@ -61,6 +61,7 @@ type State = {|
  */
 class WebView extends React.Component<WebViewSharedProps, State> {
   static defaultProps = {
+    overScrollMode: 'always',
```
satya164: Maybe `content` makes more sense as default? It's basically 'auto'.

Titozzz: Well android default is always, I think we should not alter the current behavior

jamonholmgren:
@Titozzz Can you add a note in the API reference in `./docs`?

Titozzz:
Will sure do

recepkocur:
This job, okay? overscroll feature is not working

PR #57, opened 2018-09-21
Example Project

ashishmusale:
This is initial example project. We can add separate webviews to demonstrate specific properties


Titozzz:
Hello, thanks for your contribution ! I'm trying to think about the whole value added by this, as the example here is still quite simple.

Do you think we could bring the same quantity of information (or more) in a simpler design ? Maybe like very detailed docs ?

I'm not sure the webview has the necessary complexity to need an example repo.

I'm not closed to the idea, but I'd love to know a bit more about the pain you are trying to solve :)

andreipfeiffer:
I think that an example project might be very useful, especially in the context of, let's say supporting File Uploads. This comes with a set of permissions that you need to specify in native code, both for Android and iOS.

Android has even more settings required, like `FileProvider` etc, and having that example project (that would be excluded from npm when published) might be helpful to figure out why your integration is not working.

It could also help development, because we could test regressions (automated or not).
We could publish on Github Pages html pages with various setups that maintainers and users could use as test beds.

ashishmusale:
3 reasons why I would like to have an example project are:
1) example help understand the usage better
2) If someone wants to try out the library before start using it, they can test it out there.
3) It helps the contributor to test features quickly without writing the boilerplate code to try out their changes.

jamonholmgren:
How about having a sample repo here in the react-native-community org, which we update with every release? That way you're not cloning down more than you need.

Something like:

https://github.com/react-native-community/example-webview

or

https://github.com/react-native-community/webview-demo

ashishmusale:
If we keep it in a separate repo, it will go out of sync very quickly and hard to maintain as well.

andreipfeiffer:
@jamonholmgren I too believe that keeping the example in this repo is the best solution, especially because it is needed only for this package.

Users won't download it when using `npm install` because it can be set to be ignored, using `.npmignore`.

What I think is crucial is to be able to easily make the example project use the package outside the example project, the one that you will be working with in development (not the one in `/example/node-modules`).

Otherwise, it's a long development loop, you'll have to:
* push to Github any changes you want to test
* install the latest version in the example project
* clear watchman
* re-compile

@ashishmusale do you think you can give it a shot?

jamonholmgren:
So I'm still somewhat ambivalent to this. I think until there's a very clear value proposition, I'd rather not add a lot more to the repo at this time. I'd rather spend time improving the `docs/Guide.md` to include more examples.

I'll close for now; however, I'm willing to hear counterarguments and reconsider. Decisions like this are always "until new information arises". And thanks to @ashishmusale for submitting the PR, I hope you continue to contribute!

PR #59, opened 2018-09-24
Proposal to add onOverrideUrlLoading for Android

maxammann:
This is only a proposal and we need to discuss how we implement this properly. **Maybe we also want to streamline this with iOS.**

Currently there is no way to block the navigation within an `WebView` completly and handle it on your own. E.g. to call `Linking.openUrl`.

This allows you to override all urls with `urlOverridingEnabled` and listen for events with `onOverrideUrlLoading`.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -50,6 +50,7 @@ import com.facebook.react.uimanager.annotations.ReactProp;
 import com.facebook.react.uimanager.events.ContentSizeChangeEvent;
 import com.facebook.react.uimanager.events.Event;
 import com.facebook.react.uimanager.events.EventDispatcher;
+import com.facebook.react.uimanager.events.RCTEventEmitter;
 import com.reactnativecommunity.webview.events.TopLoadingErrorEvent;
 import com.reactnativecommunity.webview.events.TopLoadingFinishEvent;
 import com.reactnativecommunity.webview.events.TopLoadingStartEvent;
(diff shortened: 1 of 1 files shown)


maxammann:
@Titozzz I'm required to handle some links specially. For example I need to open a internal PDF viewer for PDF urls. I also need internal linking between screens of my navigator.

The OriginWhiteList is too limited for me because it always sends an intend.

Titozzz:
Do you know if this is doable with the WKWebView for iOS?

jamonholmgren:
@maxammann Thanks for submitting this proposal! One thing I'm not understanding is why doesn't `onShouldStartLoadWithRequest` work on Android?

maxammann:
@jamonholmgren probably because it was simpler to implement on iOS. Or the original developer just didn't need the functionality.

jamonholmgren:
Yeah, I'd agree with @Titozzz -- let's make the original API work properly! Would this break backwards compatibility? I don't imagine it would.

maxammann:
Will come back to this PR soon. Could need some tips for easier development though. Currently I use Android Studio for the android/ project and IntelliJ for the JavaScript part.

How do you test a react-native library quickly on a device?

GitHubTracey:
I can give you all the use case... In our web app we allow clients to edit certain data with html, but there may be several other fields on the form. When they are done, and we go to show it on mobile, I take that html and it stays in the same position in the page. Instead of opening it as a separate link, I use it inline. It works very well on ios... but android with the onShouldStartLoadWithRequest not working, it trys to open the links in the middle of my scrollview... not very effective for my use case !

jamonholmgren:
@maxammann I'm hoping to make the development experience better in the coming months, including documentation about testing on device. Until then, just hack away best you can. :)

@GitHubTracey Good case study!

maxammann:
So I looked yesterday into the day it works on iOS. The only way to use the return value of `onShouldStartLoadWithRequest` to cancel or continue with the loading is by cancelling it always and using `webview.loadUrl()` to navigate further.
(The reason is that the communication between native and react-native is async)

I'll test if this works properly on android or causes any glitches.

PR #60, opened 2018-09-25
Fix #33 Support Android file upload

### LINKED ISSUE #33: Support Android file upload
There are already a lot of forks of the official webview to support that feature, there is no reason not to add it to that repo.

Let review all of them and we can discuss about implementation here.

andreipfeiffer:
Fixes #33 

I could really use some help from an Android developer on this one, because I just "made it work", don't know how to "make it work good".

Some things that should be reviewed:

- [ ] validate Android 5.0 devices (my emulator work, but outputs some weird sounds; a Galaxy 4 I tested on crashes)
- [ ] validate Android 5.1 devices (emulator works, couldn't find a real device)
- [ ] how to handle File Extensions? (https://www.w3schools.com/tags/att_input_accept.asp)

I'm sure that there's more refactoring to be done, so any help and advice would be appreciated.


Titozzz:
Thank you for the fast work 🚀 ! I've linked your PR to a few people to get more reviews so we can get this right

jamonholmgren:
I'm out of town until this weekend and won't have time until then -- but will take a look when I can!

Titozzz:
So about the FileProvider, it's great I think as it is the recommended way to do that by Google.

I'm not an android expert to judge the rest, but it seems fine to me.

One thing tho: I know that the code you used is only working on Android 5 or more, that is fine to me, as supporting older versions means using unofficial/undocumented APIs, but we should document that.

lucasferreira:
Hi @Titozzz,

I my soft experience when I implemented the same support in [react-native-android-webview](https://github.com/lucasferreira/react-native-webview-android) it's that you never could "cover" all the Android devices around there. This could be something to warn in a future documentation about this file uploading feature.

But I will try to look and help with the @andreipfeiffer good work ;)

andreipfeiffer:
@Titozzz It's true that I haven't tested on versions older than 5, but from my best guess, it should work (or more precisely, we could implement support), maybe with fewer features.

Currently, there's a `@TargetApi(Build.VERSION_CODES.LOLLIPOP)` set on the `createViewInstance()` method in the `RNCWebViewManager` that was there previously, which is preventing the `onShowFileChooser()` also.

andreipfeiffer:
In regard to the `FileProvider` which was added in sdk `22`, I have no idea what happens on sdk 21, which is LOLLIPOP 5.0. I'll have to check that.

lucasferreira:
[Here](https://github.com/lucasferreira/react-native-webview-android/blob/master/android/src/main/java/com/burnweb/rnwebview/RNWebViewModule.java#L77) we have a "sort of" implementation for Android 4.1+ without _multifile_ support.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -413,6 +414,10 @@ public boolean onConsoleMessage(ConsoleMessage message) {
       public void onGeolocationPermissionsShowPrompt(String origin, GeolocationPermissions.Callback callback) {
         callback.invoke(origin, true, false);
       }
+
+      public boolean onShowFileChooser(WebView webView, ValueCallback<Uri[]> filePathCallback, FileChooserParams fileChooserParams) {
```
Titozzz: This does not gets triggered below 5.0

There are others function signature that allow unofficial support but I'm not sure there is a point to support this below 5.0

Titozzz:
If we were to support that feature only on 5.0+ devices that would be fine to me.

We need to verify however that we can still use the webview as before on older devices

andreipfeiffer:
@Titozzz I could take a look on how to support 4.1+. Since RN sets `minSdkVersion` to `16` by default, I guess users would expect that for any community package as well.

The only question is that if you think I should tackle this on this PR, or not.
Since this is not production ready, I don't see an issue to merge this PR and add support for older SDKs later.

lucasferreira:
The worst thing about 4.1+ old API it's that we can't have support for multifile html5 inputs.

And @andreipfeiffer it's right, RN `minSdkVersion` it's `16` and I really don't know if it good to support (or not) older Android versions bellow 5.0.

Titozzz:
You could make it work for 4.1-4.3 but never for 4.4, and that would mean using undoc APIs so I'm not that hyped about it.

We are trying to keep that repo as clean as possible even if it is PRE-release as it is already linked by the 0.57 changelog and some people already use it

jamonholmgren:
@lucasferreira I would support that.

andreipfeiffer:
I would implement basic file upload (without multiple) for sdk 16+ and progressively enhance it with supported features: `accept` attr, `multiple` attr.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
andreipfeiffer: 4.1-4.4 (SDK 16-19) which means JellyBean & KitKat cover more than 10% of market share.
If it's easy to cover that, I think it's definitely worth it.
I would like to give it a shot, at least.

Titozzz:
@andreipfeiffer I really believe there is no way to get 4.4 to work and without this 4.1 to 4.3 has a really low market share.

I'd go with 5+ only and @lucasferreira suggestion.

However if you guys prefer the other way that's fine by me but we need to document the compatibility features for <5 so that once react-native drops support for Android < 5 migration is easy

andreipfeiffer:
I'm fine with either approach.

@Titozzz @lucasferreira  My knowledge is kinda limited, so could you please explain why is it not possible to implement a basic file upload for 4.4?

Titozzz:
https://github.com/delight-im/Android-AdvancedWebView/issues/4#issuecomment-70372146

andreipfeiffer:
Woaw, didn't know about that. Thanks for the insight @Titozzz 
In that case, I would support @lucasferreira 's proposal also, which 2 things to clarify:

1. how should that method be implemented? by verifying the version? or is feature detection possible?
2. how should we name the method? I would prefix it with something that suggests a boolean value, like `isFileUploadSupported()`

Titozzz:
I would not bother and just check the version.
I like prefixing bool values with is too.
Do not forget to add it on iOS too then

lucasferreira:
I agree with you guys @Titozzz and @andreipfeiffer.

Good method name `isFileUploadSupported ` and check `SDK_VERSION` for a simple comparison.

andreipfeiffer:
Will look into it in the following days.

jamonholmgren:
On iOS it would just return `true`, right?

jamonholmgren:
Really appreciate your work on this, @andreipfeiffer !!! ❤️

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
```diff
@@ -1,22 +1,320 @@
...
+  private RNCWebViewPackage aPackage;
+
+  private static final int REQUEST_CAMERA = 1;
+  private static final int SELECT_FILE = 2;
+
+  private ValueCallback<Uri[]> filePathCallback;
+  private Uri outputFileUri;
+  private String intentTypeAfterPermissionGranted;
+
+  // @todo this could be configured from JS
```
lucasferreira: When you say "this could be configured from JS" it's from RN app code or from JS executed with HTML in WebView?

IHMO that just from RN app code will be enough to this kind of personalizations and translations strings.

What do you think?

andreipfeiffer:
@jamonholmgren sorry I was so quiet these last days. For iOS I was thinking the same thing, to always return `true`, at least in this first implementation.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
andreipfeiffer: Sorry for the long silence @lucasferreira 
I was thinking of the same thing, to be passed from the RN app.

Regarding the default accepted types, I would default it to `*/*` in the Java Module, so you could select any type of file, including: pdf/doc/etc (similar to how web works). You should be able override this using the `accept` attribute from the web page.

Regarding the translations for strings, one solution would be to be able to override them from RN, passed as props.

Titozzz:
Wow, that is great work, I'll review asap

Titozzz:
Okay I have a few question.

Would it be possible to still use half the functionality (eg take picture or pick file) if only one of the permissions is defined in plist or manifest. This would allow people to use file upload without asking for not needed permissions.

What happens if you don't need file upload at all and don't want to specify the permissions?

Lastly what happens if someone did not add the permissions but still tries to use file upload ? We don't want to crash and I think we might want to display an error in development mode while doing nothing in production

andreipfeiffer:
I see 2 major use-cases here:

1. you don't need File Upload: this means that you won't display web content with `<input type="file" />` so you don't need to add any extra permissions

2. you need File Upload support: which means that you'll have to add some extra code to make it work

Since all the logic is written in native code, I have no idea how to send any error/warning to the JS thread. There is some error handling on the native side for Android. Also, iOS will tell you that you don't have the proper permissions specified, but all of this will be available only in the native consoles.

To get any message to the JS thread, I guess we would need to implement an EventEmitter, which will send messages to the JS thread, but still, the consumer would need to listen for those Events, which don't serve any purpose beyond telling you that you didn't specify the proper permissions.

CharlesStover:
For whatever it's worth, I've tested this, and it's working on my device.

This is definitely a needed feature. WebView already supports file uploads on iOS out of the box. Android is the outlier by not supporting it.

andreipfeiffer:
@Titozzz I've just checked, and the permissions for Android are used only to take photos / record videos. Selecting files from device works without any permissions.

The only thing I could add, is to display the `Take photo` and `Record video` only when you have permission to do that.

andreipfeiffer:
@Titozzz also, are there any blocking issues to get this PR merged?

Titozzz:
Thank you @CharlesStover for testing this. It is however a really huge feature and I want to make this right by not rushing it. I need that feature to be able to replace my custom webview with this repo so don't worry, it'll be merged ASAP.

@andreipfeiffer Ok so I'm fine with https://github.com/react-native-community/react-native-webview/pull/60#issuecomment-426667436 , if you made sure that the feature is opt-in only and won't break if you don't want to use it.

Last thing I hope before we can merge this is the displaying of the alert to pick a file. We need to either be able to localize string, or to render it with a function given to the JS side.

I can see upsides and downsides with both. 
Using https://developer.android.com/guide/topics/resources/localization would be probably the easiest to setup but for apps having a language settings to override phone's default on the JS side (using intl or other), you might end up with the app in the custom language and the alert in the phone language.

lucasferreira:
Hi @Titozzz and @andreipfeiffer 

About this localization/intl thing this will be working only for android right?

So we have to *prefix* some prop to people be aware that this "terms" will be needy only for android use?

Something like that:

```jsx
<Webview
  androidFileUploadStrings={{ // pt-br
    TAKE_PHOTO: "Capturar uma foto",
    TAKE_VIDEO: "Gravar um vídeo",
    CHOOSE_FILE: "Escolha um arquivo existente",
    CANCEL: "Cancelar"
  }}
  ... />```

**Comment by Titozzz** [comment]
@lucasferreira yes if we choose to use string to define this
@andreipfeiffer 

> The only thing I could add, is to display the Take photo and Record video only when you have permission to do that. This applies to Android.

Are you sure this does not require the file_provider.xml ? Or you could only use file provider and not permissions if you need half the capabilities ? It would be nice I think.

**Comment by andreipfeiffer** [comment]
@Titozzz 

__Regarding permission fallback:__
afaik: capturing media (photo/video) uses FileProvider to temporary store the files, without explicitly requiring you to set storage permissions at app level.

One thing I could do is to create a temp file, and delete if afterwards. This could tell me if you have configured the FileProvider or not, so I won't display the options in the Dialog.

__Regarding localization:__
I would't actually add a different prop for Android only. I would stick to the same approach as on iOS.
I don't really think any dev would want to control what is displayed there, I would expect it to work out of the box in the "correct language".

What is the correct language?
* default to phone's settings
* accept a language code to override from app, like "fr" or "ja"

The translations would be stored in the native package and could be expended as needed.
I (kinda) know how to do that for Android, I have no idea for iOS.

**Comment by Titozzz** [comment]
@andreipfeiffer In your docs, you ask to add:
```
  <uses-permission android:name="android.permission.CAMERA" />
```
So is it needed (or maybe for compat issues)? and could it be used instead of creating a temp file?

About localization, do we really want to define all locale inside this package instead of giving the choice to the user? I'm relatively neutral but I'd say that we need to stay consistent between iOS and Android

**Comment by andreipfeiffer** [comment]
@Titozzz 

__Permissions__
My initial implementation didn't use any permissions, only the FileProvider. I got a PR from somebody else that needed the permissions + runtime check.

Not being an Android dev, I have no idea if you need these permissions or not, if you need `camera` or `camera2` for the versions that need to be supported. That's why I need help from an Android dev that knows shit (I've mentioned this in my initial comment on this PR).

If I remove the permission from the `manifest` and the runtime checks, it works as expected on my emulators and on device. But that's not a comprehensive test.

I will try to get feedback from an Android "guy" that I know, and see what he suggests.

__Localization__
I would also stay consistent in behavior. That's why I've suggested the not so easy approach, to handle the translations ourselves, and not put the effort on the consumers (which I think it would be easier to implement).

Or, we would use icons instead of text :)

**Comment by Titozzz** [comment]
Alright, I'll try to get an android guy to check this PR too.
Icons could work, that's an idea!
I'm also fine with doing trads in the package as people can PR if they need to add more languages

**Comment by jamonholmgren** [comment]
I'm definitely fine with building in some localization support out of the box.

**Comment by Titozzz** [comment]
Hello, @ankur-sardar Opened up another PR to support android up to 4.1, as this seems to be a needed feature, I though we might want to discuss supporting it with that PR directly. I know that @andreipfeiffer was motivated to support it too.

**Comment by andreipfeiffer** [comment]
@Titozzz I have discussed today with an Android guru with a lot of experience with camera / permissions / etc, and I don't need to ask any camera permissions, so that code should go away.

@ankur-sardar 
Regarding your implementation, afaik if you don't use the FileProvider, you need to ask for storage permissions explicitly, and also handle them at runtime (for sdk 23+) because storage is so called "dangerous permissions".
Is that correct?

The FileProvider approach, shouldn't require any permissions.
However, FileProvider was added in sdk 22

So I guess, for full support, we need to combine the 2 approaches: 21 & below should use your approach, 22+ use the FileProvider.

Also, for 16-21 we don't need to handle the runtime permissions, because they were added only in 23.

That's my current understanding of the problem, so far.

**Comment by ankur-sardar** [comment]
Yes I agree, in my solution, I need to ask for Storage Permission.
Let's decide first how to handle all the different versions of sdk. After that I think it would be better to combine the solutions.

**Comment by Titozzz** [comment]
Great to see you guys joining forces into building a better solution for everyone 🔥

**Comment by andreipfeiffer** [comment]
@ankur-sardar `createChooser` does sounds like a better option, indeed.

I see the handling of the sdk versions as follows:
- 16-21: your implementation, using File + storage permissions
- 18-19: alert or toast that upload is not supported
- 22+: FileProvider

**Comment by ankur-sardar** [comment]
@andreipfeiffer All right, I have time available next week, so you can add your code to my PR. Then we can submit for a code review.

**Comment by Titozzz** [comment]
Please ping me when you need a review 👍

**Comment by Titozzz** [comment]
FYI: https://beebom.com/google-to-soon-end-chrome-support-on-android-jelly-bean-devices/

**Comment by andreipfeiffer** [comment]
Sorry for the silence, was a busy week.

@ankur-sardar Adding my stuff to your stuff was a bit difficult to me, due to other stuff that I've implemented in my PR. So, it was a bit easier to implement your approach into my PR.

@Titozzz @jamonholmgren what's been added:
- support Android 4.1 - 4.3 (reflected in `isFileUploadSupported()` as well)
- on Android 4.4 it does nothing
- replace Dialog based selection to Intent Chooser (which doesn't require translations also)

Been testing on Emulators api 16, 18, 19, 21, 23 and on real device OnePlus api 28.

Would greatly appreciate any feedback and testing, so we can get this PR merged so that people can benefit from it.

Thanks a lot.

**Comment by ankur-sardar** [comment]
@andreipfeiffer All right, no problem. Well, I was waiting for your implementation in my PR also. Or else I would have implemented in this PR. Anyway, It is the same.

**Comment by Titozzz** [comment]
This is great work. I'll will test it soon. Don't worry about the delays man you've done already plenty of work to help everyone so no need to apologize :)

**Comment by finalquest** [comment]
Hi guys. Awesome work.
I just want to point something that maybe will be troublesome for some people.
I applied this PR on my fork, and used in a project that has already have been using with a fileProvider.
So I have this config in manifest:
```
<provide
android:name="android.support.v4.content.FileProvider"
android:authorities="${applicationId}.provider"
android:grantUriPermissions="true"
tools:replace="android:authorities"
android:exported="false">
<meta-data
android:name="android.support.FILE_PROVIDER_PATHS"
android:resource="@xml/filepaths" 
tools:replace="android:resource"
/>
</provider>
```
notice that in our project we are using, FileProvider as is, and the ahutorities is applicationID.provider.

here is where the issues began.
RCWebview needs a fileProfiver, but ones that has an authorities of applicationID.**fileProvider**.
So if I use 
```
<provider
            android:name="android.support.v4.content.FileProvider"
            android:authorities="${applicationId}.provider"
            android:grantUriPermissions="true"
            tools:replace="android:authorities"
            android:exported="false">
            <meta-data
                android:name="android.support.FILE_PROVIDER_PATHS"
                android:resource="@xml/filepaths" 
                tools:replace="android:resource"
            />
        </provider>

<provider android:name="android.support.v4.content.FileProvider"
            android:authorities="${applicationId}.fileProvider"
            android:grantUriPermissions="true"
            tools:replace="android:authorities"
            android:exported="false">
            <meta-data
                android:name="android.support.FILE_PROVIDER_PATHS"
                android:resource="@xml/filepaths" 
                tools:replace="android:resource"
            />
        </provider>
```
the app wont compile, because we cant have to providers with the same name.
I'm not an android developer, but I think that or RCWebview package implements a custom fileprovider, only to allow us to declare in the manifest. Or the module should accept a param for the auhorities string.

I dont know if I explained my self with clarity. So, just let me know what you guys think.

regards

**Comment by andreipfeiffer** [comment]
@finalquest nice catch. I found a solution for this problem, will try it out tomorrow and update my PR.

**Comment by Titozzz** [comment]
Great catch I would not have seen that. ❤️ Community

**Comment by andreipfeiffer** [comment]
@finalquest can you please try to integrate the package again? See docs/Getting-Started.md for the manifest setup.

**Comment by andreipfeiffer** [comment]
I have fixed a merging conflict in `RNCWebViewManager.java`

**Comment by Titozzz** [comment]
I'll wait for @finalquest review and then I'll do testing before merging this :)

**Comment by finalquest** [comment]
I will need a couple of hours. We are under heavy fire in my job. :)

**Comment by Titozzz** [comment]
No probs, no rush!

**Comment by andreipfeiffer** [comment]
Thank you @finalquest for your input.
Regarding RequiresApi, I really have no idea about all the differences between it and TargetApi.
Tomorrow and on Friday I will meet with some Android guys for a code review on this, so hopefully they will clarify some things.

**Comment by Titozzz** [comment]
cc @elbis from #99 

> Problem 1.

>On every phone i've tested after PR#60, i can see options Camera/Video/Documents >for upload type
>On my own phone, (Galaxy A3 - 5.0.2) camera is opening properly,but on Android 7 and 8(Other phones) it's not.

**Comment by Titozzz** [comment]
I think we should ask for permission on >6 directly within this PR when the users tries to use them instead of crashing, no ?

**Comment by andreipfeiffer** [comment]
This package itself doesn't require CAMERA permissions, it doesn't USE the camera itself.
What it does, it tells another application (the default Camera App) to capture the photo/video and store it in a specific file.

It only requires STORAGE permissions, which should be added to manifest for versions below 5.
Above 5.0, the package uses the FileProvider, which doesn't require any permissions at all.

So, if the application that you use this package in, requires the CAMERA permissions for another feature, then they should be handled at the App level.

**Comment by finalquest** [comment]
> This package itself doesn't require CAMERA permissions, it doesn't USE the camera itself.
> What it does, it tells another application (the default Camera App) to capture the photo/video and store it in a specific file.
> 
> It only requires STORAGE permissions, which should be added to manifest for versions below 5.
> Above 5.0, the package uses the FileProvider, which doesn't require any permissions at all.
> 
> So, if the application that you use this package in, requires the CAMERA permissions for another feature, then they should be handled at the App level.
> 
> That's how I understand it.

You are absolutly correct.
I was thinking with iOS in mind, where camera description should be present on info.plist.

I cant get an android < 6.0, so, without a stacktrace I cant really start to look where the issue is.
On emu, the camera picker doesnt show up

**Comment by andreipfeiffer** [comment]
Ok, I had a talk with an Android guru this morning.

1. Apparently, I don't need the `FileProvider` at all, because I don't share the media with other apps.
Well, currently I do, because I store them on the External Storage, which is not needed actually.

2. I should store the media on the Internal Storage temporary, which doesn't require any kind of permissions at all, because we don't really want to persist the image/video.

3. Regarding the Camera Intent, the way it is implemented now, it basically delegates to another App to capture the media. However, we don't know if that App handles everything properly, what result will it send back. So, on specific devices, this might actually fail.

If we don't want to rely on external apps, we would need to implement our own Camera, with autofocus & shit (which is a lot of overhead).

**Comment by andreipfeiffer** [comment]
UPDATE:

1. I wasn't able to ditch the `FileProvider`. I guess that's because an external app cannot write to the private folder of our app. So, I actually need the `FileProvider` after all.

2. Currently media is written in the External Directory. Could play around with this, but wouldn't change much.

3. I don't think we should implement the Camera feature, it's a lot of overhead.

4. I will try to move all the FileProvider inside the package, so consumers won't need to make their own setup. Not sure if it works, but apparently it should be possible.

**Comment by Titozzz** [comment]
About delegating the camera to another app, I personaly prefer it when it like that cause I have several apps to take pictures and I like to be able to chose!

_TLDR: I think doing (3) is overkill, but we need (1) & (2) as it would drastically simplify the implementation right?_

**Comment by andreipfeiffer** [comment]
@Titozzz did you see my latest comment? I've posted it just a few seconds before yours, so you might have missed it.

This weekend I will try to tackle (2) and (4).
(1) leaded me to a dead end.
(3) as you said, is overkill.

Hopefully I will also get a Android 5.0 phone, so I can test on a real device.

**Comment by Titozzz** [comment]
Haha I really missed it 😅 . So to update my point of view: 

_(1) Ouch, thanks for trying 😛_ 
(2) There is no special need to do that, as whenever is use and app on android to take pictures, google photos always ask me if I want to backup the app folder so I guess it stays on my phone somewhere. So implement if you want but not necessary.
_(3) Let's forget about that_
**(4) This would be a big game changer to the users, so I'd like to have it**

**Comment by andreipfeiffer** [comment]
ok, my latest work is completed:

- I've removed the `FileProvider`, unify the way files are created for all API levels
- However, the files are still stored in the External Public Directory (was the only one that worked on lower APIs)
- if needed, I can put back the `FileProvider` for >6.0, with all setup inside the library, so no extra configuration is needed for consumers (I've tested it and it works)

Feedback is appreciated.

File: docs/Getting-Started.md [hunk 3]
```diff
@@ -20,7 +20,63 @@ $ react-native link react-native-webview
 
 _NOTE: If you ever need to uninstall React Native WebView, run `react-native unlink react-native-webview` to unlink it._
 
-#### 3. Import the webview into your component
+#### 3. Add support for File Upload
```
jamonholmgren: This is a really valuable section! In my opinion, this section works better in the `Guide.md`. I like having the Getting Started cover the most basic use case (which wouldn't include file upload). Then the "In-Depth Guide" has more specific use-cases like this one.

andreipfeiffer: Makes sense. Will update the docs.

Titozzz:
@andreipfeiffer Wow, 🔥 you are killing it dude.

I remember that I wanted to use Fileprovider for the reasons from this article:
https://drivy.engineering/android-fileprovider/

> When you need to share a file with other apps, the easiest way could be to use the external storage as a temporary place where to save this file. For example, if you need to take a picture with a camera app, you need to specify a file where the camera app will save the picture, and using external storage might be tempting.
> 
> However this solution has many drawbacks:
> 
> You lose control of your file
> Because you put your file in a public directory, you can’t safely delete it nor control which app can read and modify it.
> 
> You have to ask WRITE_EXTERNAL_STORAGE permission
> This could make many users afraid and may lead to a bad UX with runtime permissions.
> 
> External storage quickly becomes a mess
> Since you can’t safely delete your shared files, you let them in the external storage forever.
> 
> 

So is it possible to use Fileprovider so that even on android 16-18 we wouldn't need any setup either ? If not I'd say let's leave this as you did now.

andreipfeiffer:
@Titozzz I was thinking about the same thing over the weekend. I have no problem putting back the `FileProvider`. Otherwise, I guess I would have to handle runtime permissions for STORAGE.

Will be back with the (hopefully) winning solution.

andreipfeiffer:
@Titozzz I was thinking about the same thing over the weekend. I have no problem putting back the `FileProvider`. Otherwise, I guess I would have to handle runtime permissions for STORAGE.

Will be back with the (hopefully) winning solution.

Titozzz:
Arg github IS DYING. tried to delete your 5 sames posts (because of the lag) and I think I got all of them

Titozzz:
Arg github IS DYING. tried to delete your 5 sames posts (because of the lag) and I think I got all of them

andreipfeiffer:
@Titozzz I was thinking about the same thing over the weekend. I have no problem putting back the `FileProvider`. Otherwise, I guess I would have to handle runtime permissions for STORAGE.

Will be back with the (hopefully) winning solution.

francisco-sanchez-molina:
I test this pr and works well, but testing in my mobile without free space crash without message

```
11-14 18:30:23.175  2592  2727 E MiniThumbFile: couldn't save mini thumbnail data for 40587;
11-14 18:30:23.175  2592  2727 E MiniThumbFile: java.io.IOException: No space left on device
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at sun.nio.ch.FileDispatcherImpl.pwrite0(Native Method)
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at sun.nio.ch.FileDispatcherImpl.pwrite(FileDispatcherImpl.java:70)
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:89)
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at sun.nio.ch.IOUtil.write(IOUtil.java:51)
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at sun.nio.ch.FileChannelImpl.writeInternal(FileChannelImpl.java:760)
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at sun.nio.ch.FileChannelImpl.write(FileChannelImpl.java:745)
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at android.media.MiniThumbFile.saveMiniThumbToFile(MiniThumbFile.java:262)
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at com.android.providers.media.MediaThumbRequest.execute(MediaThumbRequest.java:247)
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at com.android.providers.media.MediaProvider$6.handleMessage(MediaProvider.java:755)
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at android.os.Handler.dispatchMessage(Handler.java:105)
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at android.os.Looper.loop(Looper.java:156)
11-14 18:30:23.175  2592  2727 E MiniThumbFile:         at android.os.HandlerThread.run(HandlerThread.java:61)
11-14 18:30:23.176  2592  2727 W MediaProvider:
11-14 18:30:23.176  2592  2727 W MediaProvider: java.io.IOException: No space left on device
11-14 18:30:23.176  2592  2727 W MediaProvider:         at sun.nio.ch.FileDispatcherImpl.pwrite0(Native Method)
11-14 18:30:23.176  2592  2727 W MediaProvider:         at sun.nio.ch.FileDispatcherImpl.pwrite(FileDispatcherImpl.java:70)
11-14 18:30:23.176  2592  2727 W MediaProvider:         at sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:89)
11-14 18:30:23.176  2592  2727 W MediaProvider:         at sun.nio.ch.IOUtil.write(IOUtil.java:51)
11-14 18:30:23.176  2592  2727 W MediaProvider:         at sun.nio.ch.FileChannelImpl.writeInternal(FileChannelImpl.java:760)
11-14 18:30:23.176  2592  2727 W MediaProvider:         at sun.nio.ch.FileChannelImpl.write(FileChannelImpl.java:745)
11-14 18:30:23.176  2592  2727 W MediaProvider:         at android.media.MiniThumbFile.saveMiniThumbToFile(MiniThumbFile.java:262)
11-14 18:30:23.176  2592  2727 W MediaProvider:         at com.android.providers.media.MediaThumbRequest.execute(MediaThumbRequest.java:247)
11-14 18:30:23.176  2592  2727 W MediaProvider:         at com.android.providers.media.MediaProvider$6.handleMessage(MediaProvider.java:755)
11-14 18:30:23.176  2592  2727 W MediaProvider:         at android.os.Handler.dispatchMessage(Handler.java:105)
11-14 18:30:23.176  2592  2727 W MediaProvider:         at android.os.Looper.loop(Looper.java:156)
11-14 18:30:23.176  2592  2727 W MediaProvider:         at android.os.HandlerThread.run(HandlerThread.java:61)
11-14 18:30:23.180  2592  6674 I ctaifs<Almacenamiento multimedia>[Almacenamiento multimedia][com.android.providers.media]: [openFile] 打开文件```

**Comment by jamonholmgren** [comment]
@francisco-sanchez-molina Is it only if there's no free space? Is that something we should be detecting and handling in this package?

**Comment by francisco-sanchez-molina** [comment]
Hi @jamonholmgren 
On my test, the app crash on press ```<input type="file">```, after remove a some files, the app works well and I can upload a file without a problem. 

In the logcat I can see ` java.io.IOException: No space left on device` and this `MiniThumbFile: couldn't save mini thumbnail data for 40587;` it looks like something is trying to write a thumbnail, but I do not know what causes it. I checked the PR but I did not see anything.

**Comment by andreipfeiffer** [comment]
Hey @francisco-sanchez-molina 
Yes, I need to create an empty file before actually capturing it using the external CameraApp.
As far as I know, I handle the exception when the file cannot be created, for some reason (however, I don't display this error in the UI).

The exception is indeed self explanatory.
I'm curious if (and how) I can reproduce this error.

**Comment by francisco-sanchez-molina** [comment]
Hi @andreipfeiffer 

For some reason the exception is not handled and the app crash :/

To reproduce this, I think the easiest way you can do is:

1-fill emulator freespace from adb
``` dd if=/dev/zero of=myfile.dat bs=1048576 count=100```

2-try to upload a file


if I can help something...

thanks

**Comment by andreipfeiffer** [comment]
Will try it this weekend and get back to you.

BTW: what device do you use? If you happen to have (or able to borrow) an Android 5.0 or 5.1 device, it would help a lot to test it on those devices. Those are the versions that I'm least confident of this PR.

Thanks

**Comment by francisco-sanchez-molina** [comment]
My device is a huawei p9 with android 7.0

**Comment by andreipfeiffer** [comment]
@francisco-sanchez-molina have you tried to capture an image with your default Camera App, with your disk space full?

The error doesn't look to be coming from the RN app.
I also cannot reproduce on a 8.1 emulator.
I've also looked at the code and I only create an empty file, which could cause an exception, but it handled.

This plugin doesn't handle the media capturing/selection.
It delegates to the default apps you have on your device.
I'm curious to see if the Camera App on your device handles "disk full" exceptions.

**Comment by jamonholmgren** [comment]
@francisco-sanchez-molina and @andreipfeiffer I don't think that bug should hold up merging this PR. We can address this edge case in the future if it persists.

What else remains before we can merge and publish this? cc @Titozzz

**Comment by Titozzz** [comment]
@andreipfeiffer I'd also release a first version and iterate, this has been opened for too long ? Is everything you wanted to push inside that PR ?

**Comment by francisco-sanchez-molina** [comment]
Yes, I also think it's a case edge case and this shouldn't lock this feature. I wrote it to have the adb trace. ;)

**Comment by andreipfeiffer** [comment]
@jamonholmgren @Titozzz 
I have updated the first comment of the PR.
My main 2 concerns are in regard to Android 5.0 & Android 5.1.

I agree that we should release this as is.
It's not perfect, but it's more than what's currently available.

**Comment by vitorreis** [comment]
Me waiting for this merge

File: android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java [hunk 4]
```diff
@@ -1,22 +1,309 @@
...
+  private Intent getFileChooserIntent(String[] acceptTypes, boolean allowMultiple) {
+    Intent intent = new Intent(Intent.ACTION_GET_CONTENT);
+    intent.addCategory(Intent.CATEGORY_OPENABLE);
+    intent.setType("*/*");
+    intent.putExtra(Intent.EXTRA_MIME_TYPES, getAcceptedMimeType(acceptTypes));
+    intent.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, allowMultiple);
+    return intent;
+  }
+
+  private Boolean acceptsImages(String types) {
```
Titozzz: So about that. @andreipfeiffer we did not change this. So upload won't work on my website. I feel like we should support all the types from https://www.w3schools.com/tags/att_input_accept.asp,
file_extension is currently not supported.

Should this block the first release ?

andreipfeiffer: Well, I have no idea how to handle this, especially when to enable Photo/Video capture.

IE: if you pass `accept=".gif, .png"` and your Camera outputs `.jpeg`, what should I do? Same applies for movies.

I find that this is a non-trivial implementation.

If you have any suggestions on how to handle the file extensions, please share.
Otherwise, I would implement this for the File Selection only.

Titozzz: Alright we should probably merge this for now, agreed ?

andreipfeiffer: I see it as a non-blocker, since it should enable you to select existing files from the file system.

Titozzz:
@allcontributors[bot] please add @andreipfeiffer for code, review, ideas

PR #67, opened 2018-09-29
add hideKeyboardAccessoryView option

marcelkalveram:
Fixes https://github.com/react-native-community/react-native-webview/issues/38 for the `WKWebView`, not the `UIWebView` (even though a patch there should be doable as well).

The PR mainly does the same as the function from [`react-native-wkwebview-reborn`](https://github.com/CRAlpha/react-native-wkwebview/blob/master/ios/RCTWKWebView/RCTWKWebView.m#L177), except the webview here is null when the function first runs (due to the way initialisation works) so the value gets stored and the function run again once the webview has been assigned.

The initial version is based on a solution provided on Stack Overflow: https://stackoverflow.com/questions/19033292/ios-7-uiwebview-keyboard-issue/19042279#19042279

Tested on iOS.


jamonholmgren:
Very cool @marcelkalveram! A couple things:

1. How can we make sure this also works on Android? Since `react-native-webview` is cross-platform, it's important to try to achieve cross-platform compatibility. Does Android even need this?
2. Make sure to update the `docs/Reference.md` file with the new property in this PR

marcelkalveram:
Thanks for the feedback @jamonholmgren. I'll look into 1. but doubt that it's a quick fix, since the tweak is based on the iOS-specific https://github.com/CRAlpha/react-native-wkwebview#compatibility-with-uiwebview but I'll investigate.

As for 2., I'll amend this PR accordingly.

marcelkalveram:
@jamonholmgren I've added the prop to the reference.

I did some research. For Android, there is no option like this. It's a iOS-specific option. That's what the RN docs say as well: https://facebook.github.io/react-native/docs/inputaccessoryview.

Titozzz:
Will merge soon when I'm done setting up semantic release

PR #70, opened 2018-10-05
Android webview file uploader (Support android 4+ devices) 

### LINKED ISSUE #33: Support Android file upload

ankur-sardar:
Fixes #33 

I found one similar PR  https://github.com/react-native-community/react-native-webview/pull/60
But in that implementation, it does not support android 4.x devices. 
in my solution, it supports android 4+ all devices. Please have a look

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -62,6 +62,7 @@ import java.util.Map;
 import javax.annotation.Nullable;
 import org.json.JSONException;
 import org.json.JSONObject;
+import android.annotation.TargetApi;
 
 /**
  * Manages instances of {@link WebView}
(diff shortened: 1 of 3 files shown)


Titozzz:
Hello @ankur-sardar, have you read through the other PR discussion ? There is a very long and detailed list of reviews and reasons for every decision. I think it would be easier to keep everything in one thread as the code here is very similar aside this:
```
      protected void openFileChooser(ValueCallback<Uri> uploadMsg, String acceptType) {
        module.openFileChooserView(uploadMsg, acceptType);
      }
       protected void openFileChooser(ValueCallback<Uri> uploadMsg) {
        module.openFileChooserView(uploadMsg, null);
      }
       protected void openFileChooser(ValueCallback<Uri> uploadMsg, String acceptType, String capture) {
        module.openFileChooserView(uploadMsg, acceptType);
      }
       @TargetApi(Build.VERSION_CODES.LOLLIPOP)
      @Override
      public boolean onShowFileChooser (WebView webView, ValueCallback<Uri[]> filePathCallback, WebChromeClient.FileChooserParams fileChooserParams) {
        return module.openFileChooserViewL(filePathCallback, fileChooserParams);
      }
```
That code depends on private undocumented APIs and I've read somewhere that is works for android [3.X-4.3], not for 4.4, and >5.0

ankur-sardar:
@Titozzz  I read in some discussion that Android 4.4 has some issue with the File uploader from android webview. So in Android 4.4 devices, it does not work in any web browsers(I tested from different web browsers)
So, In that case I think we can not do anything if the bug is coming from Android.
But as React native support 4.1,4.2,4.3 devices also, so I thought it would be convenient if we provide solution for those devices. For Android >5.0 , `onShowFileChooser ` method will work fine.

Titozzz:
The other PR has some advantages since we already did some iterations on it and it's using android file provider. But I started a discussion to decide if we should support <5. 👍 Do you mind if I close this PR and you join the discussion on the other one?

PR #71, opened 2018-10-09
Fix Android WebView when using UTF-8 characters (garbled output)

gianpaj:
The fix code is taken from the last line of https://stackoverflow.com/a/10831462/728287

I'm testing Ukrainian Cyrilic characters

| Before        | After           |
| ------------- |:-------------:|
|    | 

You can use this to test:
```jsx
<WebView
  // source={{ uri: 'https://hostmaster.ua/' }} // no currently an issue with websites
  source={{
  html: `<html>
    <head><meta name="viewport" content="initial-scale=1.0, maximum-scale=1.0"></head>
    <body>
      <button id="btnSubmit">Створити картку</button>
    </body></html>`,
  }}
/>
```


Titozzz:
Thanks for the fix 👍

PR #95, opened 2018-10-17
Add typescript definitions

doomsower:
Typescript definitions are based on flow
Tested them a bit in my project
I think couple of event types are screwed in flow definitions (synthetic/non-synthetic are mixed), tried my best to test which events are really received.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "react-native-webview",
   "description": "React Native WebView component for iOS, Android, and Windows 10 (coming soon)",
   "main": "index.js",
+  "typings": "typings/index.d.ts",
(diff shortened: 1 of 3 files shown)


jamonholmgren:
Awesome @doomsower ! Thanks for doing this.

Is there a way to automatically generate updated typings when we update the Flow types?

doomsower:
> Is there a way to automatically generate updated typings when we update the Flow types?

There are some projects, I haven't tried any of them myself. And it looks like none of them is mature enough to work without further human post-processing.

Titozzz:
That is cool 🔥. 

Makes me think that since most of the devs use VScode anyway I have the feeling it would make sense to replace flow with TS ? What do you guys think. I felt like using flow was kinda painful

PR #96, opened 2018-10-18
Update Debugging.md

jamonholmgren:
The new debugging guide is great but the filename is wrong. Also added a screenshot. cc @ryanlntn

### DIFF
--- a/docs/Debugging.md
+++ b/docs/Debugging.md
@@ -25,7 +25,7 @@ Settings -> Safari -> Advanced -> Web Inspector
 
 It's possible to debug WebView contents in the Android emulator or on a device using Chrome DevTools.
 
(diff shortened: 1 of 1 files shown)


Titozzz:
@jamonholmgren Are you sure the screenshot really has value? Maybe it would be better to show how to access remote devices

jamonholmgren:
@Titozzz I find it's helpful to give people a preview of what they can expect when they do this. I can make it smaller if it's too overpowering?

Titozzz:
Maybe it's too big... also we could add smaller screenshots of every steps in addition too 😄 ?

Titozzz:
We need to merge this to fix the docs, we will iterate on it later

PR #97, opened 2018-10-18
Add 'allowsBackForwardNavigationGestures' property to the iOS WKWebview

bae-unidev:
Fixes [issue](https://github.com/facebook/react-native/issues/21336) for the WKWebView.

1. Add 'allowsBackForwardNavigationGestures' property to the iOS WKWebview

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -44,6 +44,7 @@ This document lays out the current public properties and methods for the React N
 - [`url`](Reference.md#url)
 - [`html`](Reference.md#html)
 - [`hideKeyboardAccessoryView`](Reference.md#hidekeyboardaccessoryview)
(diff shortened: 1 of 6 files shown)


Titozzz:
Ouch I did break cocoapods when setting up semantic release then :( . Can you split your PR for more visibility ? I'll merge ASAP

Titozzz:
Hello, I updated master so that semantic release is now publish the version to package.json. Were you using a git url to pull the code by the way instead of npm registry ? Else I don't understand how you got that bug 🤔

bae-unidev:
I needed this feature so I was using forked git url. If this commit merged and package is published, we will fix it and you can ignore that errors.

Titozzz:
Okay this is why you had no version inside package.json. It makes sense now :)

PR #98, opened 2018-10-18
Add license to podspec.

bae-unidev:
Fix cocoapods install error and warning.
```
Fetching podspec for `react-native-webview` from `../node_modules/react-native-webview`
[!] The `react-native-webview` pod failed to validate due to 1 error:
    - WARN  | attributes: Missing required attribute `license`.
    - ERROR | version: The version of the spec should be higher than 0.
    - WARN  | license: Missing license type.
    - WARN  | source: Git sources should specify a tag.
```

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,0 +1,14 @@
+{
+  "plugins": [
+    "@semantic-release/commit-analyzer",
+    "@semantic-release/release-notes-generator",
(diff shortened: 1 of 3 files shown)


Titozzz:
Well the issue there is not only the license (warning) but the version. lemme see what I can do

Issue #66, opened 2018-09-29
[DISCUSS] Cross platform API for RN<>WebView 2-way RPC/message passing


fungilation:
Referencing (my) prior discussions regarding this, for context and so we don't repeat here. In reverse chronological order:

- https://github.com/alinz/react-native-webview-bridge/issues/109#issuecomment-425669058
- https://github.com/react-native-community/discussions-and-proposals/pull/3#issuecomment-414858049
- https://github.com/facebook/react-native/pull/9762#issuecomment-260583009

This consistent, robust, cross platform API is important for any app that blends native and webviews. I list [mine](https://medium.com/wonderswipe/rethink-mobile-search-10-100x-faster-introducing-wonderswipe-6f2ff0d0e667) as one. A requirement of this API is that an injected `postMessage ` for purpose of RN message passing, not be overridden by a website's JS in the WebView. Which is currently the case with RN core's WebView, and presumably this forked react-native-webview. It could be achieved (and have been attempted in prior discussions and in RN core's WebView) by loading RN's injected `postMessage` last, but that's unreliable should the website create after onLoad a function of the same name, alongside iframe issues. This is not hypothetical as it does get overridden for a fact, when loading google.com in WebView for example.

A simple function rename that doesn't use the web standard of `postMessage` would sidestep what is effectively a name collision issue. Call it `rnPostMessage` or anything else. Open to suggestions on a better alternative to renaming that actually has a guarantee for not being overridden by non-locally loaded JS within WebView.

This new API should consider/port code (if tested as reliable) from [react-native-wkwebview](https://github.com/CRAlpha/react-native-wkwebview), which is iOS only. And implement the same API for other platforms. There is also working RN<>WebView bridge API at [react-native-webview-bridge](https://github.com/alinz/react-native-webview-bridge) that's well established (old but fully functional, with google.com loaded in webview, and for Android and iOS), but that's based on UIWebView and not WKWebView on the iOS side.

jamonholmgren:
Thanks @fungilation ! 

One other interesting API I found when perusing Github was this one:

https://github.com/pinqy520/react-native-webview-invoke

It exposes an API where you can call an async function directly, as long as you control both the web and mobile side of the connection. Under the hood, it still uses `postMessage`, but I thought I'd bring this up in case it was interesting as a potential API.

fungilation:
Interesting. Although requiring below on the Web side is no good. New API should support any WebView src, not only local js:

```js
[...]

PR #104, opened 2018-10-20
Revamp of postMessage implementation

KoenLav:
This PR:
- Achieves consistent naming of the global ReactNative communication channel in the WebView between Android and iOS;
- Only calls the linkBridge functions in both Android and iOS when overwritePostMessage is set to true, as a result this package no longer clutters the window.postMessage by default;
- Renames MessageHanderName to MessageHandlerName.

Feedback is much appreciated!

Addresses:
https://github.com/react-native-community/react-native-webview/issues/23
https://github.com/react-native-community/react-native-webview/issues/66
https://github.com/react-native-community/react-native-webview/issues/79

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "prettier.singleQuote": true,
+    "prettier.trailingComma": "all"
+}
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -95,7 +95,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   protected static final String HTML_ENCODING = "UTF-8";
   protected static final String HTML_MIME_TYPE = "text/html";
   protected static final String WEBVIEW_MESSAGE_HANDLER = "window.webkit.messageHandlers.";
-  protected static final String BRIDGE_NAME = "ReactNativeBridge";
+  protected static final String HANDLER_NAME = "ReactNative";
 
   protected static final String HTTP_METHOD_POST = "POST";
 
@@ -266,7 +266,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
       /**
        * This method is called whenever JavaScript running within the web view calls:
-       *   - window.webkit.messageHandlers[BRIDGE_NAME].postMessage
+       *   - window.webkit.messageHandlers[HANDLER_NAME].postMessage
        */
       @JavascriptInterface
       public void postMessage(String message) {
@@ -326,9 +326,9 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
       messagingEnabled = enabled;
 
       if (enabled) {
-        addJavascriptInterface(createRNCWebViewBridge(this), WEBVIEW_MESSAGE_HANDLER + BRIDGE_NAME);
+        addJavascriptInterface(createRNCWebViewBridge(this), WEBVIEW_MESSAGE_HANDLER + HANDLER_NAME);
       } else {
-        removeJavascriptInterface(WEBVIEW_MESSAGE_HANDLER + BRIDGE_NAME);
+        removeJavascriptInterface(WEBVIEW_MESSAGE_HANDLER + HANDLER_NAME);
       }
     }
 
@@ -378,7 +378,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
         "window.originalPostMessage = window.postMessage;" +
         "window.postMessage = function(data, origin) {" +
           "window.originalPostMessage(data, origin);" +
-          WEBVIEW_MESSAGE_HANDLER + BRIDGE_NAME + ".postMessage(String(data));" +
+          WEBVIEW_MESSAGE_HANDLER + HANDLER_NAME + ".postMessage(String(data));" +
         "}" +
       ")");
     }
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -94,7 +94,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
   protected static final String HTML_ENCODING = "UTF-8";
   protected static final String HTML_MIME_TYPE = "text/html";
-  protected static final String WEBVIEW_MESSAGE_HANDLER = "window.webkit.messageHandlers";
+  protected static final String WEBVIEW_MESSAGE_HANDLER = "window.webkit.messageHandlers.";
   protected static final String BRIDGE_NAME = "ReactNativeBridge";
 
   protected static final String HTTP_METHOD_POST = "POST";
@@ -249,6 +249,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   protected static class RNCWebView extends WebView implements LifecycleEventListener {
     protected @Nullable String injectedJS;
     protected boolean messagingEnabled = false;
+    protected boolean overwriteWindowPostMessage = false;
     protected @Nullable RNCWebViewClient mRNCWebViewClient;
 
     protected class RNCWebViewBridge {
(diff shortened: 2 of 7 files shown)


jamonholmgren:
@KoenLav Very cool, thank you for taking this on!

I'd like someone to provide feedback who is more well-versed in the specifics of `postMessage` and the bridge.

I assume this breaks compatibility with older WebViews? What will app developers need to do to migrate?

(We'll also want documentation in the `docs` folder if we get that far with this.)

KoenLav:
@Titozzz in my opinion it is easiest to spark some discussion with some code samples, hence the PR.

Also, I think the discussion in #66 (and other issues) noted that the original implementation (clobbering window.postMessage) is unnecessary at best and causes errors at worst.

This warrants a breaking change. We could however provide an option which would allow users to reinstate the old functionality, by adding a property overWriteWindowPostmessage?

In my opinion the breaking change is the only thing standing in the way of this getting merged, and it seems it is warranted here, but I'd happily discuss that!

With regard to the other repo (bridged-webview) I would like to ask you to also take a look at that, as this PR is merely laying out some groundwork and removing past inconsistencies.

KoenLav:
Updated this PR to reflect new insights from: https://github.com/react-native-community/react-native-webview/issues/66#issuecomment-432648318

KoenLav:
@Titozzz @jamonholmgrem please review the latest changes.

It covers the entire discussion in the referenced issue. 

Implementing a breaking change in the location of the postMessage code in the Android and iOS Webviews and providing a prop to overwrite window.postMessage for those who were relying on this (faulty) behavior and not wanting to update their code just yet.

We will need to write docs informing users to migrate. 

I didn't include any of the code which is used in the BridgedWebview repository, I will present these changes as a separate PR.

Let's get things moving! ;)

Titozzz:
Thanks I will review this tomorrow :)

jamonholmgren:
@KoenLav Apologies for the delay on this, both @Titozzz and I have been extremely busy with our day jobs.

I've pinged @axe-fb in Slack to check on one thing -- I'm worried about the `_ReactNativeBridge` module being used here. If he gives the go-ahead I think we're ready to merge and publish.

One other question (you may have said this before but I didn't see it with a quick scan) -- does this PR preserve backwards compatibility with `window.postMessage` so we aren't breaking apps who upgrade? I understand it's a simple fix, but avoiding breaking it would be desirable if possible.

Once again, thanks so much for your help.

empyrical:
Is it possible for the React Native messaging stuff to be in a different JavaScript "realm" from the JS that runs on the main page, like how Chrome content scripts work?

KoenLav:
@jamonholmgren this PR adds a property called overwritePostMessage.

If you enable that you get the old experience. 

I did improve on that as well, as before it would not call the original window.postMessage, which it does now.

Titozzz:
@jamonholmgren  any update from @axe-fb ? Refactoring post message is our next big step after Android upload :)

KoenLav:
?

Titozzz:
I'll ping him more directly !

jamonholmgren:
No updates from @axe-fb , unfortunately. We may need to find someone else at FB to review this.

KoenLav:
I'm for for being so blunt, but the code changes aren't dramatic; it might be worthwhile to take a look and move things ahead :)

Titozzz:
I really want to, right now I'm just overloaded with work and I have really only a few minutes of free time here and there. I'll do my best to improve my time to do open source next year 😉

KoenLav:
@Titozzz I thought the point of migrating the WebView to a community package was to allow development to speed up.

Maybe you should ask others to chime in on maintaining?

jordansexton:
@KoenLav there's a small problem in practice with this code on Android.

Instead of defining the expected iOS object path at `window.webkit.messageHandlers.ReactNative`, the code as written instead creates an object at `window['window.webkit.messageHandlers.ReactNative']`.

[This line](https://github.com/KoenLav/react-native-webview/blob/b4078fd1d62db0f7bd3faf95a3bda7ce6a2b86f2/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L326) is the proximate cause, but the underlying cause is that `addJavascriptInterface` expects to define a global on the `window` object directly, not at a nested path.

I also think the obvious workaround (adding something to the injection to first define `window.webkit.messageHandlers.ReactNative` if it doesn't exist) is not ideal. Defining `window.webkit.messageHandlers` in a non-WebKit browser to avoid using a global is a code smell and risks conflict with other libraries, as does defining `window.webkit.messageHandlers.ReactNative` (nonspecific to this library). The implementation would also be overly complicated (have to check for the existence of `window.webkit`, `window.webkit.messageHandlers`, and `window.webkit.messageHandlers.ReactNative` to just define a method).

What about just defining a global `ReactNativeWebView_postMessage` function, and on

- Android: calling `window.ReactNativeWebView_postMessage` directly
- iOS: defining a reference or wrapper between it and `window.webkit.messageHandlers.ReactNativeWebView.postMessage`

This is unambiguous, won't conflict with anything, and provides a way to call `addJavascriptInterface` without complexity.

KoenLav:
@jordansexton thanks for your feedback!

I agree with the direction proposed.

Would you be willing to fork my repository/branch and create a new PR (if possible rebased against the current master branch)?

I'd be happy to review and/or improve upon your suggested changes!

jordansexton:
I'm happy to help, but unfortunately because of #198 I'm not sure I can run the latest version. I have your code merged against v2.8.0 for testing in my app. I'm using this code inside the webview as a workaround:

```js
let ReactNative;
if (
    (window.webkit != null) &&
    (window.webkit.messageHandlers != null) &&
    (window.webkit.messageHandlers.ReactNative != null) &&
    (typeof window.webkit.messageHandlers.ReactNative.postMessage === 'function')
) {
    ReactNative = window.webkit.messageHandlers.ReactNative;
}
else if (
    (window['window.webkit.messageHandlers.ReactNative'] != null) &&
    (typeof window['window.webkit.messageHandlers.ReactNative'].postMessage === 'function')
) {
    ReactNative = window['window.webkit.messageHandlers.ReactNative'];
}

if (ReactNative != null) {
    ReactNative.postMessage(/* ... */);
}
```

I appreciate your implementation and it works. Since the topic of a bridge has come up in #66 and elsewhere, it would be nice to see some consensus from the maintainers on whether they like this direction. I'll help update the code for merging if so.

Titozzz:
I'm working on fixing #198 for you right, now and then this PR will be the next one to get onboard the repo I hope (finaly huh @KoenLav 😅 !).

KoenLav:
@jordansexton great to hear, thanks!

@Titozzz much appreciated; do let me know if I can help out in any way.

ps my previous comment about asking others to chime in wasn't a personal one, really just a suggestion if things are too busy at the moment :)

Titozzz:
Well there are many things that needs attention, it's hard to follow everywhere but I keep doing my best. I've fixed @jordansexton issue so hopefully we can get going

jordansexton:
@Titozzz I've locally caught this PR by merging in the master branch, resolving the conflicts, and manually reviewing the full diff of the result against master.

How should I proceed with getting that added to this PR? Or should I open a new PR from a new branch in my fork? All @KoenLav's commits have be retained.

I'm working now on making the changes from https://github.com/react-native-community/react-native-webview/pull/104#issuecomment-452132323

jordansexton:
Also, much of the implementation of the PR here is understandably concerned with overwriting/not overwriting `window.postMessage`. However, I think the implementation in master already breaks the function by changing its signature, losing the origin, and not calling the overwritten function anyway. This has major implications, for example, to Stripe's JS embed (which uses iframes and `postMessage`).

IMO we should never be messing with this global at all, so even with an upgrade path provided by the prop `overwriteWindowPostMessage` (which defaults to `false` in this PR), this would be a breaking change anyway (it changes the current default behavior), and would be yet another breaking change later to remove that prop.

Are we comfortable getting rid of this (not touching `window.postMessage`) and just making updating now a breaking change?

KoenLav:
@jordansexton you should be able to merge the changes from this branch into your fork and make a new PR. Or you could provide a PR with your fixes to my branch. Whichever your prefer!

ps I completely agree that the old behavior (which screws up any proper usage of window.postMessage) was faulty and I can't really imagine anyone would want to (keep) relying on the old situation (as it's probably really, really easy to migrate to the new situation).

I'm perfectly fine with removing the overwriteWindowPostMessage prop, as I don't think it should have ever worked like this in the first place.

@Titozzz your thoughts?

Titozzz:
I'm fine with removing any use of window.postMessage and not adding a prop for backward compatibility if we document how to update so that the least experimented users won't have any issues understanding why / how.

We are using semver so it should be fine.
Thanks for the help @KoenLav @jordansexton .

jordansexton:
Awesome, I'll work on that tomorrow.

KoenLav:
Closing this in favor of: https://github.com/react-native-community/react-native-webview/pull/243

PR #107, opened 2018-10-22
Support onShouldStartLoadWithRequest on Android

maxammann:
This PR adds support for `onShouldStartLoadWithRequest` on android.

The initial PR was https://github.com/react-native-community/react-native-webview/pull/59

The issue for this PR is: https://github.com/react-native-community/react-native-webview/issues/106


malonguwa:
@maxammann @jamonholmgren @Titozzz Hi guys~ thanks for the hard work! Just want to ask when we can get this PR merged in~~~ really need this feature~ 👍

maxammann:
@Titozzz @jamonholmgren can you take a look at this?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -72,12 +78,14 @@
...
+ *  - LOAD_URL
  *
  * {@link WebView} instances could emit following direct events:
  *  - topLoadingFinish
  *  - topLoadingStart
  *  - topLoadingStart
  *  - topLoadingProgress
+ *  - topShouldStartLoadWithRequest
  *
  * Each event will carry the following properties:
```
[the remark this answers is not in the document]
Titozzz: Please create an issue about that, we need to fix this

js/WebView.android.js
```diff
@@ -131,6 +131,37 @@ class WebView extends React.Component<WebViewSharedProps, State> {
 
     let NativeWebView = nativeConfig.component || RNCWebView;
 
+    const compiledWhitelist = [
```
Titozzz: I think this code does not belong here.
It is a duplicate of iOS code, so it should be shared inside WebViewShared.js as much a possible !

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -144,52 +153,10 @@ public void onPageStarted(WebView webView, String url, Bitmap favicon) {
...
-        }
-      }
-
-      if (mOriginWhitelist != null && shouldHandleURL(mOriginWhitelist, url)) {
-        return false;
-      }
-
-      launchIntent(view.getContext(), url);
+      dispatchEvent(view, new TopShouldStartLoadWithRequestEvent(view.getId(), url));
       return true;
```
Titozzz: Are we sure that this won't affect our current users ? I mean we are going to interrupt all the loads and communicate with the JS side

maxammann: I tested the behavior and it seems to be ok. We're doing it now the same way as iOS does it. I did some simple performance tests and it seems to have no impact.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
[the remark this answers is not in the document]
maxammann: Done: https://github.com/react-native-community/react-native-webview/issues/126

js/WebView.android.js
maxammann: I moved the code to WebViewShared. Hope it is still simple enough.

maxammann:
@Titozzz oh yeah sure forgot that one :/ Will do it tomorrow evening. Anything else I should change before merge?

Titozzz:
So I digged more into your implementation and iOS implementation.
While you return true instantly and then call loadURL, wouldn't it be more interesting to do exactly as iOS does, which is wait for JS for maxTime (0.25) and then decide to stop the call or not ?

maxammann:
@Titozzz Hm, I'm not sure if its good to block the UI thread. Also I think there is no real benefit. But yeah that would be an other way.

Titozzz:
Do you think we should update iOS then ?

maxammann:
I think that is not needed. Probably there was a reason to implement it that way because the current implementation is not easy.

Also I can not do it. Can't even read Objective-C :D

Titozzz:
Ok I'll have a look right now I have one big hour to code :)

maxammann:
Thank you for fixing it! I would test it now a last time on ios and android. Can we merge this afterwards?

js/WebView.android.js
```diff
@@ -161,7 +157,6 @@ class WebView extends React.Component<WebViewSharedProps, State> {
         domStorageEnabled={this.props.domStorageEnabled}
         messagingEnabled={typeof this.props.onMessage === 'function'}
         onMessage={this.onMessage}
-        onShouldStartLoadWithRequest={onShouldStartLoadWithRequest}
```
maxammann: You still need to pass this. Else react native can not call this function.

maxammann:
Android: Working after adding this line again `onShouldStartLoadWithRequest={onShouldStartLoadWithRequest}`

maxammann:
iOS: Still working, yey!

Titozzz:
@maxammann For me it's not working as expected.
It never get called on the first rendering (unless the web server you are querying redirects you). You can test it on simple web pages such as:
`source={{ uri: 'https://adrien.harnay.me' }}`
Do you know if there's a workaround ?

Also there was an error in the code were the ! was missing here:
 ```
if (!passesWhitelist(compileWhitelist(originWhitelist), url)) {
      Linking.openURL(url);
    }
```
So I have no idea how it worked on your end when you tested 🤔 

Once the last issue is solved we should be good, but right now I don't have an idea

maxammann:
Hm, forgot to test the whitelist :/

About the problem of the first rendering: There is a method [onLoadResource](https://developer.android.com/reference/android/webkit/WebViewClient.html#onLoadResource(android.webkit.WebView,%20java.lang.String))

But there is no way to cancel the loading :/ But is this really something we need to support. I mean the behaviour on iOS and Android must be the same, but I can not image a usecase where you would start a webview and cancel the first load.

Titozzz:
@maxammann Well if we don't get the event on the first load, the code won't go through the whitelist, and that mean you could load a website not inside the whitelist so that is my first concern.

Also, if I encountered that bug it's because it's happening in my codebase, and since the iOS behavior is working the other way it seems hard to ignore the difference :/

maxammann:
@Titozzz Alright I think I get it now. I already had to include `originWhitelist={['*']}` on iOS but not on android. So this behavior already existed before this PR. I agree that we have to fix it. Will look into it whether we can fix it for android or if we maybe have to change the iOS behavior because its not possible on android. 

Should we do this in an other PR because the problem already existed before this PR?
An other option would be not to extend the `onShouldStartLoadWithRequest` but deprecate it and implement a new way to handle *only* navigation events (and not the initial load). Also whitelist would only work for navigation.

Also I do not really get why calling `loadUrl` should be a problem according to the note.

Titozzz:
I need to do some before / after tests to make sure we are not doing breaking change without knowing it!

I'll come back to you asap, hopefully this weekend.

jamonholmgren:
@maxammann Apologies for the delay, both @Titozzz and I have been extremely busy with our day jobs.

I believe we're just waiting on some additional testing before we merge and publish.

Once again, thanks so much for your contributions here, it's very much appreciated!

dyihoon90:
@maxammann @jamonholmgren @Titozzz Thanks for the work guys! My team is watching out for this feature as well.

maxammann:
Very busy right now too. Building a react-native team right now. Our app heavily depends on this feature so I'll try to work in this this week. 

Maybe we can improve the API a little bit more. Also I think we could introduce breaking changes because the package split from react-native.

Titozzz:
I have no issues with breaking changes, as long as everything is well documented.

wayne214:
very good

maxammann:
@Titozzz I merged in master and tested again. Seems to work fine. Can we merge this?

Titozzz:
Can you confirm this ?

Without this PR:
Android whitelist origin is checked on first and other loads
Android has no access to onShouldStartLoadWithRequest
iOS whitelist origin is not checked on first loads but is checked on other loads
iOS has access to onShouldStartLoadWithRequest

With the PR:
[BREAKING CHANGES] - Android whitelist origin is not checked on first loads but is checked on other loads 
Android has access to onShouldStartLoadWithRequest
iOS whitelist origin is not checked on first loads but is checked on other loads
iOS has access to onShouldStartLoadWithRequest

Am I right ?

Titozzz:
good catch

maxammann:
Currently evaluating what you said and the mac update takes ages

maxammann:
Oke this would be the Changelog imo:

iOS: no changes
* Whitelist is checked on user initiated loads
* onShouldStartLoadWithRequest is called on the initial load -> Whitelist origin is checked on initial load
* onShouldStartLoadWithRequest gets called and you can block the loading by returning false
Android: 
* Whitelist is checked on user initiated loads
* onShouldStartLoadWithRequest is NOT called on the initial load -> Whitelist origin is not checked on initial load
* onShouldStartLoadWithRequest gets called and you can block the loading by returning false

So here is my idea:
I would leave it for now because this feature is usable. In a second PR I would:
* Remove originWhitelist
* Rename onShouldStartLoadWithRequest to shouldOverrideUrlLoading (This has nothing to do with individual requests)
* shouldOverrideUrlLoading would only get called for user initiated loads

Also I would look into onLoad, onLoadStart, onLoadEnd and onNavigationStateChange because this is very confusing.

Titozzz:
We do agree that before the PR, android was checking the whitelist on first load ? So I need to release this as breaking ?

maxammann:
@Titozzz I'm very sure that it did not check the whitelist on first load. Androids `shouldOverrideUrlLoading` does not get called on the first load.

So no breaking changes.

Titozzz:
Ok I'm very sorry, I thought that calling client.setOriginWhitelist was somehow doing it and since we removed that code I imagined it was breaking. However client.setOriginWhitelist was just setting a class variable that is now useless and so you are right.

That being said, can you clean mOriginWhitelist from the code since it's useless now ?

maxammann:
@Titozzz I'm very sure that it did not check the whitelist on first load. Androids `shouldOverrideUrlLoading` does not get called on the first load.

So no breaking changes.

maxammann:
@Titozzz I added jcenter() because gradle did not find trove4j.

maxammann:
Thank you very much! I'll create a Issue soon to cleanup the parts which are a bit confusing!

Titozzz:
sounds good :)

oliverdolgener:
Thank you very much for this PR. It solves almost all the problem we had. To some extend iOS and Android still act a little differently on initial loading. For now we can work around them but it would be nice if they get addressed like you said before:

> So here is my idea:
I would leave it for now because this feature is usable. In a second PR I would:

> Remove originWhitelist
Rename onShouldStartLoadWithRequest to shouldOverrideUrlLoading (This has nothing to do with individual requests)
shouldOverrideUrlLoading would only get called for user initiated loads
Also I would look into onLoad, onLoadStart, onLoadEnd and onNavigationStateChange because this is very confusing.

Thanks again for your hard work!

longzhi:
Great job! Just want to ask when we can get the PR merged. I really need this feature.

PR #109, opened 2018-10-25
Add `userScript`,`userScriptForMainFrameOnly` prop

ifsnow:
As you know `WKWebView` has [WKUserScript feature](https://developer.apple.com/documentation/webkit/wkuserscript?language=objc). It provides the function to run JavaScript on every frame(iframe) of a web page. This feature can be useful in many cases.

Please review it positively.

Test Plan:
----------
I checked if the code below works. If `userScript` is injected into all frames, the color of the YouTube header title changes to red.

```js
render() {
  const userScript = 'var injectionNode = document.createElement("style"); injectionNode.innerHTML = ".html5-video-player a { color: red; }"; document.body.appendChild(injectionNode);';

  return (
    <View>
      <WebView
        source={{
          uri: 'https://www.youtube.com/embed/8qCociUB6aQ',
        }}
        userScript={userScript}
      />
    </View>
   );
}  
```

### Normal title

default title

### Title changed by `userScript`

changed title

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -45,6 +45,8 @@ This document lays out the current public properties and methods for the React N
 - [`html`](Reference.md#html)
 - [`hideKeyboardAccessoryView`](Reference.md#hidekeyboardaccessoryview)
 - [`allowsBackForwardNavigationGestures`](Reference.md#allowsbackforwardnavigationgestures)
+- [`userScript`](Reference.md#userScript)
+- [`userScriptForMainFrameOnly`](Reference.md#userScriptForMainFrameOnly)
 
 ## Methods Index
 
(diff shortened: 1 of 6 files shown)


Titozzz:
Thanks for the contrib, is there a way to achieve the same effet on android ?

Titozzz:
By the way, userScript used with MainFrameOnly true, would probably overlaps with injectedJavascript, so we might want to do something to not have twice the same functionality

ifsnow:
@Titozzz Yes, I have a way to work on Android. Do you think this is a useful feature? This is because implementation is more complex than iOS. :) I will work on it if necessary. 

I was also worried about what you said. I think there are two approaches.

1. Add `injectedScriptOnAllFrames` property instead of `userScript` and `userScriptForMainFrameOnly`.
2. Show warning when `userScript` and `injectedScript` are used with `userScriptForMainFrameOnly` true.

Which approach do you prefer? In my opinion, the first approach seems to be clearer. Although, `injectedScript` keyword is ambiguous.

ifsnow:
@Titozzz Do you have any ideas about what I mentioned? :)

Titozzz:
Okay so the goal here is to be non-breaking if there is no win to break stuff.

So we currently have injectedJavascript prop that is basically just the injectJavascript function ran once on the first successful load.

There already are a few issues with the current implem that we might want to solve when implementing this. I'll get back to you with a more detailed plan 😄

ifsnow:
@Titozzz I hope your work is done successfully. I'll be back here then.

ifsnow:
@jamonholmgren @Titozzz I fully understand you have a busy day. It's not easy to maintain the open source. I can wait forever until a decision is made. :)

PR #112, opened 2018-10-26
Add 'userAgent' property to the iOS WKWebView.

bae-unidev:
Add 'userAgent' property to the iOS WKWebView. 

The property to set the userAgent in the iOS WKWebview is [customUserAgent](https://developer.apple.com/documentation/webkit/wkwebview/1414950-customuseragent). But to make the interface the same, I set userAgent props to WKWebview customUserAgent.

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -342,11 +342,11 @@ Boolean value to enable third party cookies in the `WebView`. Used on Android Lo
 
 ### `userAgent`
 
-Sets the user-agent for the `WebView`.
+Sets the user-agent for the `WebView`. If you are using iOS WKWebView, it works as customUserAgent.
 
-| Type   | Required | Platform |
-| ------ | -------- | -------- |
-| string | No       | Android  |
+| Type   | Required |
+| ------ | -------- |
+| string | No       |
 
 ---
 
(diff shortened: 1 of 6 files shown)


KoenLav:
@Titozzz aren't you vouching for removing UIWebView completely soon?

bae-unidev:
@Titozzz 

UIWebview does not officially change the `User-Agent`. Instead, there are three hacks:

- Modify window.navigator.userAgent in the js context with the stringByEvaluatingJavaScriptFromString API.
- Replace the `User-Agent` Header for all requests in the `loadRequest`.
- Modify `NSUserDefaults` to override the settings for the entire app.

Mostly not modern, just HACK. It seems good to add that docs do not support it, is that what you think?

johanforssell:
What about this text in `Reference.md`:
```
### `userAgent`

Sets the user-agent for the `WebView`. This will only work for iOS if you are using WKWebView (see https://developer.apple.com/documentation/webkit/wkwebview/1414950-customuseragent), not UIWebView.

| Type   | Required |
| ------ | -------- |
| string | No       |
```

Titozzz:
@KoenLav You are right, I'll remove it soon, too tired to keep maintaining it.

But I the meantime I like to keep everything clean and up to date ^^'

xclidongbo:
I want to get useragent in wkwebview. How can I do it?

PR #114, opened 2018-10-29
Modified the document for `injectJavaScript`

ifsnow:
I think `injectJavaScript` is a method, not a props.

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -8,7 +8,6 @@ This document lays out the current public properties and methods for the React N
 
 - [`source`](Reference.md#source)
 - [`automaticallyAdjustContentInsets`](Reference.md#automaticallyadjustcontentinsets)
-- [`injectJavaScript`](Reference.md#injectjavascript)
 - [`injectedJavaScript`](Reference.md#injectedjavascript)
 - [`mediaPlaybackRequiresUserAction`](Reference.md#mediaplaybackrequiresuseraction)
 - [`nativeConfig`](Reference.md#nativeconfig)
@@ -53,6 +52,7 @@ This document lays out the current public properties and methods for the React N
 - [`goBack`](Reference.md#goback)
 - [`reload`](Reference.md#reload)
 - [`stopLoading`](Reference.md#stoploading)
+- [`injectJavaScript`](Reference.md#injectjavascriptscript)
 
 ---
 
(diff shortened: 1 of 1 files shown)


gianpaj:
Not really. While yes, you will pass the JavaScript you want to be run, it's converted a string.
The JavaScript needs to be a converted to a string.

It's the same as the built-in WebView in react-native:
https://facebook.github.io/react-native/docs/webview#injectedjavascript

For example:
https://stackoverflow.com/a/47061864/728287

This is how I do it, in order to keep the JavaScript syntax higlighting, etc. on IDE:
```jsx
function JStoInject() {
  alert('injected');
}

class MyWebView extends Component {
  render() {
    return (
      <WebView
        source={{
          html: `<html>
            <head><meta name="viewport" content="initial-scale=1.0, maximum-scale=1.0"></head>
            <body><p>hello fellow webview RN dev</p>
            </body></html>`,
        }}
        injectedJavaScript={`(${JStoInject.toString()}());`}
      />
    );
  }
}

**Comment by ifsnow** [comment]
@gianpaj `WebView` has two features. `injectedJavaScript` is what you said. Another `injectJavaScript` is a function, not a property.

In the code below, You can see how these two are different.

1. "Hello" is displayed on the screen. (`injectedJavaScript`)
2. When you touch the button, it changes to "Hello react-native-webview" (`injectJavaScript`)

```js
import React, { PureComponent } from 'react';
import { View, Button } from 'react-native';

export default class WebviewTest extends PureComponent {
  _webviewRef = React.createRef();

  _onPress = () => {
    const script = 'document.getElementById("hello").innerText = "Hello react-native-webview"';
    this._webviewRef.current.injectJavaScript(script);
  }

  render() {
    const injectedJavaScript = 'document.getElementById("hello").innerText = "Hello"';

    return (
      <View style={{ flex: 1, paddingTop: 40 }}>
        <Button title="Change a text" onPress={this._onPress} />
        <WebView
          ref={this._webviewRef}
          source={{
            html: `<html>
              <head><meta name="viewport" content="initial-scale=1.0, maximum-scale=1.0"></head>
              <body><p id="hello"></p>
              </body></html>`,
          }}
          injectedJavaScript={injectedJavaScript}
        />
      </View>
    );
  }
}
```

gianpaj:
That's true my bad.

Just more notes:
1. I wouldn't call it a "script". It's a function
    > Script will be passed to the WebView and executed immediately as JavaScript.
2. I would live the table to say that's a function _and_ it's not required.
3. I wouldn't remove the fact that a string needs to be the argument of the function otherwise people might overlook and think you can a pass a normal JS function.

ifsnow:
@gianpaj I agree with your idea. So, I changed the expression a little more clearly.

PR #121, opened 2018-10-31
refactor/android-tooling-update-and-kotlin-tests

Titozzz:
Updated gradle so that it will auto detect react-native when working from android studio plus smaller changes
Setup kotlin basics and tested on events class


android/build.gradle
```diff
@@ -1,22 +1,91 @@
...
+}
+
+repositories {
+    mavenCentral()
+    maven {
+        url 'https://maven.google.com/'
+        name 'Google'
+    }
+
+    // Stolen from react-native-firebase, thanks dudes!
```
Salakar: ❤️

PR #123, opened 2018-11-01
Set encoding to UTF-8 in Android.

bendangelo:
Test Plan:
----------
1. Load the HTML below into a webview on Android:
```
let html = `<html lang="fr">
<head>
</head>
<body>
commandité
</body>`;

return (<WebView
        originWhitelist={['*']}
        source={{ html: html }}
      />);
```

2. The French "e" should have the proper accent.

Release Notes:
--------------
[Android][BUGFIX][WebView] - Set encoding to UTF-8 in Android.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -93,7 +93,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   protected static final String REACT_CLASS = "RNCWebView";
 
   protected static final String HTML_ENCODING = "UTF-8";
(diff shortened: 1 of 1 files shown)


Titozzz:
I'm not sure I understand why #71 did not address any issue left with HTML encoding

Titozzz:
Are you sure you are up to date ?

```
        if (source.hasKey("baseUrl")) {
          view.loadDataWithBaseURL(
              source.getString("baseUrl"), html, HTML_MIME_TYPE, HTML_ENCODING, null);
        } else {
          view.loadData(html, HTML_MIME_TYPE + "; charset=" + HTML_ENCODING, null);
        }
```

bendangelo:
Sorry it's not a priority to check this at the moment. I originally made this fix in the react-native repo and was told to back port it here, so there is a good chance this fix is not needed.

PR #129, opened 2018-11-06
Fixed for supporting mediaPlaybackRequiresUserAction option under iOS 10.

ifsnow:
There is a problem that the `mediaPlaybackRequiresUserAction` option does not work under iOS 10.

I hope this helps.


jamonholmgren:
I've tested this and it does fix the issue with iOS 10. Thank you @ifsnow !

PR #133, opened 2018-11-07
Updates Android implementation for urlPrefixesForDefaultIntent

lewnelson:
Updates Android implementation for urlPrefixesForDefaultIntent by providing [shouldOverrideUrlLoading](https://developer.android.com/reference/android/webkit/WebViewClient.html#shouldOverrideUrlLoading(android.webkit.WebView,%20android.webkit.WebResourceRequest)) method in place of deprecated method with the same name [shouldOverrideUrlLoading](https://developer.android.com/reference/android/webkit/WebViewClient.html#shouldOverrideUrlLoading(android.webkit.WebView,%20java.lang.String)).

I've left the deprecated method to support devices < API level 24. I've also updated the reference docs to add documentation for this prop.


Titozzz:
This will conflict with #107 so I'll wait a bit

PR #134, opened 2018-11-11
Add fix for keyboard dismiss leaving viewport shifted in iOS 12

iamtommcc:
There's currently a pretty major bug with WKWebView on iOS 12 that leaves the viewport awkwardly shifted upwards after the keyboard is dismissed. Makes using WebViews with forms an extremely janky experience.
https://openradar.appspot.com/radar?id=5018321736957952

The bug exists on any iOS 12 app compiled with Xcode 10 (which is almost all of them 😅)

This PR adds a keyboard listener that manually animates the content offset, shrinking it on keyboard dismiss. The solution is borrowed from this Ionic PR: https://github.com/ionic-team/cordova-plugin-ionic-webview/pull/201/files

The fix works seamlessly, however, I am unsure of what impact it may have elsewhere (if any).

### DIFF
--- a/ios/RNCWKWebView.m
+++ b/ios/RNCWKWebView.m
@@ -11,6 +11,7 @@
 
 #import "objc/runtime.h"
 
+static NSTimer *keyboardTimer;
 static NSString *const MessageHanderName = @"ReactNative";
 
 // runtime trick to remove WKWebView keyboard default toolbar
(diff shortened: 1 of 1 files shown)


jamonholmgren:
So this did end up having an adverse side effect, listed here: https://github.com/react-native-community/react-native-webview/issues/201#issuecomment-446062183

We're going to revert this until we can find a better fix.

PR #141, opened 2018-11-16
[wkwebview] create api to allow clients to present a client credential for authentication

mjhu:
In order for TLS Mutual Auth to work for webviews, the caller must present a credential. Expose a setter that can be called to set a credential.


jamonholmgren:
Thanks for the PR @mjhu !

* Is this only applicable to iOS, not Android?
* Can you update the `./docs` with the new setter?
* What's a good way for us to test if this works on our machines?

Titozzz:
FYI: https://github.com/facebook/react-native/commit/8911353c47af018f78c1cff59dfab05b975e39ed This has been merged into the core. But i'd still like that we wait for the answers to above questions :)

mjhu:
Thanks for the comments @jamonholmgren and @Titozzz!

* This is only applicable to iOS for now.
* Will update the docs with the new setter 
* Our use case is that we pass in a string label for the identity that we stored. We have another RN module that takes a string label and calls the RCTWKWebView static method to store the fetched credential object. Finally, any opened webviews that requires the client credential will pass auth checks .

mjhu:
@jamonholmgren and @Titozzz sorry for the delay! I updated the docs for usage of the new setter.

docs/Custom-iOS.md
```diff
@@ -132,6 +132,16 @@ Once these are exposed, you can reference them in your custom web view class.
 }
 ```
 
+### Setting Client Certificate Authentication Credential
+
+If you open webpages that needs a Client Certificate for Authentication, you can create a credential and pass it to the webview:
+
+```
+[RNCWKWebView setClientAuthenticationCredential:credential];
```
jamonholmgren: @mjhu  Can you explain where you would put this code, normally? Would it be in the `AppDelegate.m` file?

mjhu: It doesn't have to be in AppDelegate.m. For our use case, we have this call inside our auth initiation stack that runs as part of app start. The only requirement is that call is made before using any webviews. I'll update the comment with this.

mjhu:
@jamonholmgren Regarding testing, this was done on an internal app and I verified the webview presents the client cert correctly and is able to access the test site.

Titozzz:
As this landed on the core webview, I'll merge it here too.

jaulz:
@mjhu do you have some code examples by any chance how you use it? 😊

PR #147, opened 2018-11-19
Add "Frame load interrupted" error handling for OAuth for RNCWKWebView

oNaiPs:
Adds the same code that is on `RNCUIWebview` https://github.com/react-native-community/react-native-webview/blob/master/ios/RNCUIWebView.m#L271 for the same purpose.

I came across this problem while implementing google SAML support with the RNCWKWebView.

### DIFF
--- a/ios/RNCWKWebView.m
+++ b/ios/RNCWKWebView.m
@@ -402,6 +402,13 @@ static NSString *const MessageHanderName = @"ReactNative";
       return;
     }
 
+    if ([error.domain isEqualToString:@"WebKitErrorDomain"] && error.code == 102) {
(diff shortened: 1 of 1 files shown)


jamonholmgren:
Thanks for the PR @oNaiPs ! How would I go about testing this?

oNaiPs:
> Thanks for the PR @oNaiPs ! How would I go about testing this?

@jamonholmgren not sure, I came across this error while developing with google SAML for an internal service.

tigo0:
This works in our case with the authentication flow. Thanks

Titozzz:
Thank @tigo0 for testing this, I'll merge into master

PR #148, opened 2018-11-19
Add incognito prop to iOS WKWebview

oNaiPs:
Allows the webview to be opened with an ephemeral data storage.

Looks like there is no out-of-the-box support for `UIWebView` so I didn't implement that.
I think there should be a relatively simple way to implement this on Android as well.


Andreyco:
FYI, using this fork in production app, works well!

Titozzz:
@Andreyco Thanks for your testing, I'll merge this

Andreyco:
Superb!

PR #152, opened 2018-11-20
[feat] Add `cacheEnabled` prop

Salakar:
This PR 'fixes' https://github.com/react-native-community/react-native-webview/issues/128 

cc @Titozzz @jamonholmgren as per linked issue above. Let me know your thoughts and if anything needs changing. I've enabled it by default as discussed with @Titozzz  on slack.

### Checklist

- [x] Android support
- [x] iOS support
- [x] Flow Types updated
- [x] Reference Documentation updated

### Release Notes

 - [BREAKING] [FEAT] Added a new ~`enableCache`~ `cacheEnabled` prop to toggle Android & iOS webview caching behavior. This change makes caching **enabled by default** when previously there was no caching behavior which may cause unexpected behaviour changes in your existing implementations.


Salakar:
I need to do a few minor changes for Android still, please don't merge yet. Thanks

https://developer.android.com/reference/android/webkit/WebSettings.html#setAppCacheEnabled(boolean)
https://developer.android.com/reference/android/webkit/WebSettings.html#setAppCachePath(java.lang.String)
Both needed as well.

Salakar:
Right, this should be good to go now, cc @Titozzz @jamonholmgren

thedusty99:
Hey there! Thanks for your work! Is there anything stopping this PR from getting merged (besides the conflicts now)?

Titozzz:
Yeah, it needs testing, because when I tested it I did not feel the expected improvements

munsifhayat:
Any updates on this one ?
Need to have this 'WebCache' enabled property.

Titozzz:
@munsifhayat did you test this PR works for you ?

Titozzz:
Keep me updated

jerolimov:
For my project I can confirm that this PR improves the caching and performance. 👍 

@Salakar @munsifhayat @Titozzz But unfortunately, in the `WebView.ios.js` its required to push the props to the `<NativeWebView...` in line 249 https://github.com/react-native-community/react-native-webview/blob/ac59610f55fbe909b2120dfef029ac4b33811b60/js/WebView.ios.js#L248-L258

Without this _enableCache is always false in `RNCWKWebView.m`. I expect that its similar on Android but doesn't debug it here.

I would also recommend to rename `enableCache` to `cacheEnabled` or `cachingEnabled` (or something similar), to match all the other [props](https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#props-index)

But independently of this, this PR is great and important 👍

Titozzz:
@jerolimov excellent catch about the props naming. We need to do that. 👍

Titozzz:
@Salakar @jerolimov I updated the PR according to the reviews above :)
What do you think ?

jamonholmgren:
Let's ship it!

jerolimov:
@Titozzz Sorry for the delay. Thanks for applying the feedback. Sourcecode looks fine. I tested it also in a live project on iOS and it works as expected. 👍 

(I could not verify this yet on Android because I have other troubles here.)

munsifhayat:
From my side I installed it got an RCTWebview error which I solved by dragging that library into Xcode project. 


Imported the Webivew as required and than : 
<WebView 
                ref="webview" 
                source={{ uri: this.state.url }} 
                onLoadEnd={e => this.webViewLoaded()}
                domStorageEnabled = {true}
                javaScriptEnabled = {true}
                style={styles.container}
                cacheEnable = {true}
           />

Rest are working fine but 'cacheEnable' could be verified once I have live build Soon. Let me know if anything need to change. 
I am using this for installing : 

"react-native-webview": "git+https://github.com/react-native-community/react-native-webview.git"

munsifhayat:
Its working perfectly fine.

NeverFearTheasHere:
I think this `cacheEnabled` prop is missing from the typescript prop type declarations (in`typings/index.d.ts`) so I've opened a PR for that: #335

PR #157, opened 2018-11-21
Better types

ccorcos:


Titozzz:
By the way we want to use TS instead of flow of you are motivated

ccorcos:
Cool. Should be a fairly easy switch.

PR #159, opened 2018-11-21
Fix unhandled promise rejection error

jerolimov:
Fix unhandled promise rejection error when the URL was not whitelisted and this URL could not be opened by `Linking.openURL` as well.

This happen for example when try to open a (relative) file:// url without whitelist the file:// scheme.

Not handling promises results in a red screen in the iOS simulator. This change renders a warning instead.

### DIFF
--- a/js/WebView.ios.js
+++ b/js/WebView.ios.js
@@ -202,43 +202,6 @@ class WebView extends React.Component<WebViewSharedProps, State> {
       webViewStyles.push(styles.hidden);
     }
 
-    const nativeConfig = this.props.nativeConfig || {};
-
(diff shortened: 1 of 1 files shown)


Titozzz:
Please use async / await

jerolimov:
@Titozzz I changed it to async / await and also extract some methods from the render method into (imho a little bit more readable) class methods.

The whitelist checking code was not touched.

jamonholmgren:
@jerolimov Thanks for this PR! Looks really good to me. Can you fix the conflicts and also what's a good way for us to test this before we merge? And will this break backwards compatibility in any way?

Issue #106, opened 2018-10-22
Support onShouldStartLoadWithRequest on Android


alebedev:
Please add `onShouldStartLoadWithRequest` support on Android.  react-native-webview-android has it, but built-in webview never had.  This causes developers to use different WebView libraries on different platforms

**Context**
I use WebView to display some static resources like license agreement, documentation, etc.  It's desirable to open links from those files in system browser rather than inside same webview.  Currently, the best way to do this is to use `onShouldStartLoadWithRequest`, return `false` and use `Linking` to open URL externally

Titozzz:
You mean https://github.com/react-native-community/react-native-webview/pull/59 ..?

maxammann:
Ok shit.. I really have to fix this soon :)

malonguwa:
@maxammann @Titozzz hi guys~ thanks for looking after the project! Just want to ask when we can get the PR merged~~~ really need this feature~~

Titozzz:
I've reviewed the PR!

PR #163, opened 2018-11-21
add targetFrame to nativeEvent on onShouldStartLoadWithRequest

francisco-sanchez-molina:
On my app I have a custom logic to open some external links in the mobile browser. But I have too an iframe with some links that can modify the url of these iframe. For this reason I need to differentiate if the change of url is in the main frame or in another to apply my custom logic.

PD: In android onNavigationStateChange is not fired with iframe links


Titozzz:
We try to keep the events consistent between Android and iOS , is the same feature doable on Android ?

jerolimov:
On Android (since Android >= 7.0, API level 24) this is also supported on Android. See:

https://developer.android.com/reference/android/webkit/WebViewClient#shouldOverrideUrlLoading(android.webkit.WebView,%20android.webkit.WebResourceRequest)

francisco-sanchez-molina:
`onShouldStartLoadWithRequest` callback is only available on ios. 

image

For this reason the event "is consistent". But, I think that another PR can implement this callback for android and send url, navigationType and targetFrame data.

What do you think?

Titozzz:
Sorry there is already a PR which is why I'm mistaken. It's a bit stuck tho, so please come and discuss #106

Titozzz:
@francisco-sanchez-molina Now that the feature also exist on android, I'd like to stay consistent and implement it on both places at once

francisco-sanchez-molina:
Hi @Titozzz, I am going to review it as soon as possible, and I will tell you if I can do it ;)

Titozzz:
Thank you so much for the instant reply ❤️

vivan-niche:
since the android API just returns a boolean, it may make sense to return a boolean `isMainFrame` instead of the main frame name or `other` ?

jamonholmgren:
I'm okay with this change, even if it's iOS-only. However, let's make sure that documentation is updated too. See PR https://github.com/react-native-community/react-native-webview/pull/493 for an example of how to do this.

PR #164, opened 2018-11-21
[android] replace deprecated `WebView.PictureListener`->`onContentSizeChange` impl

Salakar:
### Description

The current implementation of `onContentSizeChange` uses an old deprecated API [setPictureListener](https://developer.android.com/reference/android/webkit/WebView.html#setPictureListener(android.webkit.WebView.PictureListener)) - it was deprecated in API 12 but this library is set to a minimum of API 16 - this API most likely doesn't exist on devices using newer APIs 💥 

I've swapped it out for [WebView#onSizeChanged](https://developer.android.com/reference/android/webkit/WebView#onSizeChanged(int,%20int,%20int,%20int))

### Release Notes

 - [ANDROID] [FIX] `onContentSizeChange` now uses `WebView`s own `onSizeChanged` events instead of the now deprecated API `setPictureListener`.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -295,8 +295,8 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
           this,
           new ContentSizeChangeEvent(
             this.getId(),
-            this.getWidth(),
-            this.getContentHeight()
+            w,
+            h
           )
         );
       }
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -18,7 +18,6 @@ import java.util.Map;
 import android.content.ActivityNotFoundException;
 import android.content.Intent;
 import android.graphics.Bitmap;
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -285,6 +288,20 @@ public void onHostDestroy() {
...
     }
 
+    @Override
+    protected void onSizeChanged(int w, int h, int ow, int oh) {
+      if (sendContentSizeChangeEvents) {
+        dispatchEvent(
+          this,
+          new ContentSizeChangeEvent(
+            this.getId(),
+            this.getWidth(),
```
Titozzz: Can't use any of the variables above ?

Salakar: Potentially I guess, I just wasn't sure which ones they relate to as the former implementation relied on `getContentHeight()` and not sure if this is the same value as provided by `int height`

Thoughts? Can swap it over if needed?

Titozzz: Well I'd rather use the variables provided instead of mesuring again, if possible

Titozzz:
@allcontributors[bot] please add @Salakar for code, review, ideas, tool

PR #165, opened 2018-11-21
[ios] Add 'pagingEnabled' property to the iOS WKWebview

wiscat:
Resolve this [issue](https://github.com/react-native-community/react-native-webview/issues/153)


Titozzz:
Do you know if that is doable on Android ? We like to have as many shared props as possible

wiscat:
@Titozzz I could be wrong, but there is no such option on native Android webview. There is some custom implementation on react native side inside ReactScrollView component, but this package don't use it. Again I'm not totally sure.

PR #166, opened 2018-11-21
More better types

ccorcos:
I considered moving everything over to typescript but I got hung up on importing android.js vs ios.js. Pretty annoying to deal with...


Titozzz:
I want to move to whole repository to typescript. If I setup the project, do you feel okay to do the refactoring ?

PR #167, opened 2018-11-22
Fixing typing.

bruteforcecat:
so happy to see there is typing in the newest release.
I added back WebViewIOSLoadRequestEvent which also exist in react-native typing and fix startInLoadingState which should be boolean.


Titozzz:
@bruteforcecat Are you willing to convert all to TS ?

bruteforcecat:
I think I could do this. so no more flow-typed? but you want to change the setup as well?

Titozzz:
Yeah we will remove flow. Ideally we can generate flow types from TS and export them if people need them

Titozzz:
#168

PR #170, opened 2018-11-24
[ios] Add allowsLinkPreview property to WKWebView

jermaine:
### allowsLinkPreview
Boolean value that determines whether pressing on a link displays a preview of the destination for that link. This is specific to iOS with devices that support 3D Touch.

**More info:**
https://developer.apple.com/documentation/webkit/wkwebview/1415000-allowslinkpreview


Titozzz:
Great PR :)

LeslieOA:
@jermaine @Titozzz Thanks for adding this.
Question: is there currently a way to customise "peek and pop" to use the existing WebView rather than opening the link in  Safari?
[Read this article](https://webkit.org/blog/7016/ios-10-link-preview-api-in-wkwebview/) and wondered.

PR #171, opened 2018-11-25
Migrate to Typescript

bruteforcecat:
Hello this PR is for this [issue](https://github.com/react-native-community/react-native-webview/issues/168)

This PR is mainly removing flow and use typescript. I also remove the dependencies of fbjs.
Haven't done the part of auto-generating flowtype from typescript yet because I am not too familiar with this. Maybe someone can help me on this later PR?

I tested on my project. It seems everything works fine. Please let me know if there is anything wrong I would love to correct them.

And I haven't add tslint yet. Maybe I will leave it to others to decide what linting rule we wants.


bruteforcecat:
sorry the prepare script in package.json seems not working properly as there is the only index.js got transpiled. and I am still looking for reason

Titozzz:
Hello, thank you for starting this, this is already a lot of work 🎉 .

A have a few feedbacks for this PR that will need to be address // simplify the merge.

- [ ] Use the same config for ' instead of " (This makes the PR super hard to review)

- [ ] Use the same config for trailling comas (Same reason)

- [ ] There is no need to build the typescript using TSC (Since react 0.57 uses babel 7 with the typescript preset to strip the typing)

- [ ] We do not want to eradicate flow completely, as some people might use it, so we will need to either do it manualy, either generate flow-typed definitions

- [ ]  (Debatable) I personaly do not use .jsx / .tsx, and I do everything with .js and .ts files

bruteforcecat:
Thx for the feedback. Sorry it's because of my auto format from the code editor and I see there is no any linting tool used right now? anyway I will change it back the `'` and trailing comma.
And also fix those issue hopefully tmr

Titozzz:
I'm still working on the setup of prettier / eslint / typescript and vscode configuration.

Once all that is done, do you want me to push to your branch directly ?

Titozzz:
I updated your branch with the following things in it (and also submitted a PR so you can review the diffs):

VSCode project config
Prettier - code formatting
ESLint - code linting
Typescript - type checking

All tests can be run with yarn ci:test

I tested the code with my own repository and it works great and type integration with VSCode is 👌 :)

Mandatory left to do: 

- [ ] fix the remaining 'any' that are breaking the tests

- [ ] Create _manually_ index.flow.js at the root of the folder.

- [ ] Update readme.md with the new state of the repo (I'll do it once everything else is ready)

empyrical:
From a TypeScript user's perspective, would there be any disadvantage if this project just maintained Typescript stubs alongside the Flow typed files, rather than doing a whole port of the codebase to TypeScript?

Titozzz:
As a user no, as a contributor yes, I find it way easier to do ts than flow (because of the tooling)

_This is of course my point of vue_

empyrical:
Considering how difficult I found it to make flow checks for multiple RN versions work it may be for the best 🤔

Titozzz:
I will do that :)

Titozzz:
Done

index.ts
```diff
@@ -0,0 +1,10 @@
+import { Platform } from 'react-native';
+import WebViewIOS from './src/WebView.ios';
+import WebViewAndroid from './src/WebView.android';
+
+const WebView = Platform.OS === 'android' ? WebViewAndroid : WebViewIOS;
```
LinusU: I might be missing something here, but won't this negate the automatic inclusion of only the correct platform for each bundle, and now always include both the Android and iOS code for both platforms? 🤔 

Would anything be lost by simply setting `"main": "src/WebView"`? 🤔

Also, and this is probably a much broader discussion, but I think that this is the first package in the React Native ecosystem to be published as pure TypeScript, instead of shipping the, by `tsc`, precompiled version. Is this something that we want to do? What are the pros/cons.

One con that I see is that it will probably mess up for people that are using a custom `.babelrc` that doesn't include TypeScript. I actually personally found it a problem that some React Native ships as flow code, which means that you cannot use a custom `.babelrc` that doesn't include flow.

The only pro I can think of is that it simplifies the publishing of this module a little bit, since we don't have to have a `"prepare"` script. Might be other pros that I'm missing though 🤔 

One approach would be to let `tsc` transpile the `src/` dir into `lib/`, and then have two small entry files:

**`index.js`**

```js
import WebView from './lib/WebView'

// We keep this for compatibility reasons
export { WebView }
export default WebView
```

**`index.ts`**

```ts
import WebViewIOS from './lib/WebView.ios'
import WebViewAndroid from './lib/WebView.android'

declare const WebView: WebViewIOS | WebViewAndroid

export { WebView }
export default WebView
```

The `lib/` folder would contain both `*.js` and `*.d.ts` files, which is the default output from `tsc`. We would just have to add `"prepare": "tsc"` to the package.json-scripts.

Titozzz:
Thank you for working on this @bruteforcecat. I'll finish setting everything up this weekend hopefully.

Little update after many discussion with react-native contributors:

> We will ship index.js, index.d.ts and index.js.flow . As much as I'd like to push things forward, there is no need to rush solutions that would not be perfect for now just for the sake of it.

> I really want to set a good example for the RNC, so I'll try to be extra careful with the project setup (eslint/prettier/tsc).

bruteforcecat:
Thanks @Titozzz 💪

bruteforcecat:
hello @Titozzz . any update on this one? is there anything I could help?

Titozzz:
No I'm just super sorry about this, I'm so busy right now, but don't worry I did not give up. I'm actually working with on TS migration inside my own company code too 💪

Titozzz:
I've picked up this subject again. Will try to get it done ASAP

Issue #150, opened 2018-11-20
[WKWebview] iOS 9 - Crash on unmounting


hieuvanlai:
it works perfectly with ios 10.x.x and greater, but in ios 9.1, it makes app crash when Webview unmount

Titozzz:
If you start the app in a simulator you should see the line that crashes. Could help us track the bug down

hieuvanlai:
i am running in iPhone 6S.

hieuvanlai:
<img width="1601" alt="screen shot 2018-11-20 at 3 54 46 pm" src="https://user-images.githubusercontent.com/27544529/48762032-b21ceb00-ecdc-11e8-985e-a4a4e8cdffa2.png">

hieuvanlai:
when it crashes, it shows this message. is it useful for you?

Titozzz:
That's what I asked, but that one is not helping me 😢

biduone:
Unfortunately, I have the same problem.

ReactNative : 0.57.6
ios: 9.3

![image](https://user-images.githubusercontent.com/2018001/49341874-7bf14b00-f68e-11e8-9e1d-eb1a43c57450.png)

Titozzz:
I'd love to get a pull request to address that issue!

AndrewJack:
I think I've seen this in crashlytics (only on ios 9):
```
Crashed: com.apple.main-thread
0  libobjc.A.dylib                0x2190467a objc_retain + 9
1  UIKit                          0x2677c78d -[UIScrollView setDelegate:] + 36
2  WebKit                         0x290bef31 -[WKScrollView _updateDelegate] + 228
3  WebKit                         0x290c89ff -[WKWebView dealloc] + 266
4  libobjc.A.dylib                0x219053a9 (anonymous namespace)::AutoreleasePoolPage::pop(void*) + 388
5  CoreFoundation                 0x2205ef89 _CFAutoreleasePoolPop + 16
6  CoreFoundation                 0x2211006f __CFRunLoopRun + 1582
7  CoreFoundation                 0x2205f229 CFRunLoopRunSpecific + 520
8  CoreFoundation                 0x2205f015 CFRunLoopRunInMode + 108
9  GraphicsServices               0x2364fac9 GSEventRunModal + 160
10 UIKit                          0x26733189 UIApplicationMain + 144
11 Example                     0x97def3 main (main.m:14)
12 ???                            0x21d07873 (Missing)
```

[...]

PR #138, opened 2018-11-15
Add shared process pool so cookies and localStorage are shared across webviews in iOS

kylemantesso:
Add a shared process pool in iOS to allow shared cookies and localStorage between webviews. This brings iOS into line with expected behaviour in Android.

PR #175, opened 2018-11-28
implement sharedCookiesEnabled prop for iOS RNCWKWebView

SebiVPS:
We had the problem on iOS WebViews that local cookies (stored in local HTTPCookieStorage, set with  [react-native-cookie](https://github.com/shimohq/react-native-cookie) ) were not added in loadRequests. On Android the local stored cookies were sent like expected.
This kinda "hacky" solution is the only way we found, that works for us.
The stack overview link is in the code below.
If someone finds a better solution we would very much like to accept that.


Titozzz:
Isn't the issue belonging in the react-native-cookie repo ?

SebiVPS:
@Titozzz No, there is no problem with the cookie repo, the cookies are saved as expected. But they are missing at the requests, which are created by the webview.
The cookies are there:
`[[NSHTTPCookieStorage sharedHTTPCookieStorage] cookies]`

We have to set them "manually" in the WebView Configuration:
`[wkWebViewConfig.websiteDataStore.httpCookieStore setCookie: currentCookie completionHandler: nil];`
 (for iOS 11+)

That was the only way, we found, to set a custom cookie and add it to the WebView's requests.

kylemantesso:
@SebiVPS @Titozzz Please see my PR that already adds this functionality and does not rely on the react-native-cookie repo:
https://github.com/react-native-community/react-native-webview/pull/138

Titozzz:
@kylemantesso this PR is about using react-native-cookies with the webview.
@SebiVPS #138 is about sharing cookies across wkwebviews

ios/RNCWKWebView.m
```diff
@@ -106,7 +106,63 @@ - (void)didMoveToWindow
...
+          NSArray<NSHTTPCookie*>* cookies = [[NSHTTPCookieStorage sharedHTTPCookieStorage] cookies];
+          for(int i = 0; i < (int)[cookies count]; ++i)
+          {
+              NSHTTPCookie* currentCookie = cookies[i];
+              [wkWebViewConfig.websiteDataStore.httpCookieStore setCookie: currentCookie completionHandler: nil];
+          }
+
+          _webView = [[WKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
+
+          // Hotfix: set cookies twice, otherwise they are not sent with first request
```
jamonholmgren: Why does this need to be set twice?

It does look like the `setCookie:completionHandler:` is an async method, so is it because there's a race condition here?

SebiVPS: Yeah you are right. I found another solution here (https://forums.developer.apple.com/thread/97194)
This line of code should do the trick: 
`wkWebViewConfig.websiteDataStore = [WKWebsiteDataStore nonPersistentDataStore];`
Ititialze the websiteDataStore worked for me.

SebiVPS: I updated the PR, do you need further information?
Will you merge the PR into master?

SebiVPS:
Do you need further information to merge this PR into master?

Titozzz:
I've asked @jerolimov @hippotz @PratibhaSomashekhara to try it out, so I'm waiting for their feedback. We do agree that this would enable cookies for WKWebView < iOS 11.0?

jerolimov:
@SebiVPS @Titozzz  the iOS >= 11.0 part works fine. But the script workaround for iOS < 11 doesn't work for me.

Maybe it helps in some other scenarios where the HTML is public available and you need the cookie to fetch some additional data from an API.

In my scenario this doesn't help. Because our server validates the cookies for the URL which was displayed in the webview. But the "user script" was not applied if the server returns just an error. For example a HTTP 403 Forbidden.

SebiVPS:
Thank you @jerolimov!   I added your code to add the cookies for versions < iOS 11, can i delete the "script workaround" for < 11, this seems useless now? 
Could you please test again on other iOS Versions?

Thank you!

SebiVPS:
> @SebiVPS In my test scenario, I also need also your "JavaScript injection workaround". So I think we need both. Can you add this again?

It is still available :) i didn't delete it.

> I could test it later today. I will also check if its possible to escape the cookies when they contain a `'`.

Thank you :)

jerolimov:
@SebiVPS Ah perfect, didn't see it. 👍  Will test it this evening :)

Titozzz:
Keep me updated 👍

jqn:
@SebiVPS how can I test this pull request? I ran into the same cookies issue yesterday and migrating my app from WebView core to WebView community can't happen until this is resolved.

jerolimov:
@SebiVPS @Titozzz Still testing this issue and will give feedback asap. Currently I analyse the issue #150 where my app crashes on iOS 9.3.

Titozzz:
Thank you for the hard work @jerolimov, really means a lot to me. 👍😍❤️🎉😊

jerolimov:
Hi @SebiVPS @Titozzz here is my final feedback. 😄

I extends your branch with just one commit you can find here: https://github.com/jerolimov/react-native-webview/commit/9826d91df90cdb5592ddb100d9272a81cb75ad60 This commit clean ups a little bit the javascript "cookie injection" and solves the problem that cookies with an `'` could not be handled.

I keep some code in this version which could be used to unset also all cookies of the WebView before setting the shared cookies from the app. But I expect that this would make more troubles and confuses more people than it helps. I could understand if you say that that commented code is not clean. In this case I or @SebiVPS or you could remove it later.

I also extracted the code which add all cookies to the NSURLRequest into its own method (`requestForSource`) and also simplify and harden the code with the power of `NSHTTPCookie`.

@SebiVPS For me its fine if you copy and paste my RNCWKWebView.m over yours and commit these changes as you. Or just merge my branch into yours. But because this PR lives a little bit longer, it contains already multiple merge commits. Maybe it would be great to rebase the change into a new branch directly based on the latest master branch. What do you thing?

Sure, all of this is a workaround for older iOS versions. But beside of this, I'm really happy with result.  As said before, I tested this with a own expressjs based webserver. For this I created a new React Native project which you can use if you want to verify this branch (incl. my changes).

You find the project here: [react-native-webview-shared-cookies-test](https://github.com/jerolimov/react-native-webview-shared-cookies-test)

SebiVPS:
@jerolimov  Thank you very much for your work. I cherry-picked your commit to this branch.
@Titozzz I added the `sharedCookiesEnabled` prop to the Reference.md (i hope this is ok, you still can update the text if you wish, or @jerolimov)

jerolimov:
Hi @SebiVPS @Titozzz, from my side this PR is fine. Currently I have nothing to add for the documentation. If I found more time for this, I would create a new PR. Can we ship this? 🚀 

Looking already into another improvement for this project. 😏

ios/RNCWKWebView.m
```diff
@@ -142,6 +142,69 @@ - (void)didMoveToWindow
     wkWebViewConfig.mediaPlaybackRequiresUserAction = _mediaPlaybackRequiresUserAction;
 #endif
 
+    if(_sharedCookiesEnabled) {
+      // More info to sending cookies with WKWebView
+      // https://stackoverflow.com/questions/26573137/can-i-set-the-cookies-to-be-used-by-a-wkwebview/26577303#26577303
+      if (@available(iOS 11.0, *)) {
+        // Set Cookies in iOS 11 and above, initialize websiteDataStore before setting cookies
+        // See also https://forums.developer.apple.com/thread/97194
+        wkWebViewConfig.websiteDataStore = [WKWebsiteDataStore nonPersistentDataStore];
```
Titozzz: This is already defined by cacheEnabled props, are you sure this still belongs here?

SebiVPS:
@Titozzz i refactored the code, now the ci tests fail (i think this does not have something to do with the latest commit?), can you please check this?
Thank you!

Titozzz:
Yeah CI is broken I'm trying to find what broke it... Sorry about that

jamonholmgren:
CI should be fixed when #395 is merged.

SebiVPS:
@Titozzz can you recheck the code and

ios/RNCWKWebView.m
```diff
@@ -127,6 +127,68 @@ - (void)didMoveToWindow
...
+              cookie.expiresDate.timeIntervalSince1970 * 1000
+            ];
+          }
+          [script appendString:@";\n"];
+        }
+        [script appendString:@"})();\n"];
+
+        WKUserScript* cookieInScript = [[WKUserScript alloc] initWithSource:script
+                                                              injectionTime:WKUserScriptInjectionTimeAtDocumentStart
+                                                           forMainFrameOnly:NO];
```
Titozzz: forMainFrameOnly:NO

Isn't this a risk to leak your cookies to anything loaded by your website? I'm not sure 🤔

SebiVPS: Yep indeed, 
From Docs: `A Boolean value indicating whether the script should be injected only into the main frame (true) or into all frames (false).`
It should not be necessary to provide the cookie to all frames. 
@jerolimov could you give this a try? I could not test on devices with iOS < 11 atm. Thank your very much.

SebiVPS: @Titozzz I changed the code to `forMainFrameOnly: YES` seems the logic way. 
I tested on an old iPad with iOS 9.x, seemed to work as expected.

tunegov:
Merge that, please!

SebiVPS:
@Titozzz @jerolimov Can we ship this?

kuzyo:
Build locally and used in my project, but still having issue with sharing in WebView in ios, Android works okay. We also using react-native-cookies package.

tmaly1980:
I'm still not able to delete cookies using CookieManager.clearAll() on iOS. If I call that method, and then call getAll() afterward, it shows nothing, but the WebView still has the cookie somewhere.

Daavidaviid:
> I'm still not able to delete cookies using CookieManager.clearAll() on iOS. If I call that method, and then call getAll() afterward, it shows nothing, but the WebView still has the cookie somewhere.

Me too I'm having the same issue. Because we use the webview for some home made sso.

Dylan0916:
> I'm still not able to delete cookies using CookieManager.clearAll() on iOS. If I call that method, and then call getAll() afterward, it shows nothing, but the WebView still has the cookie somewhere.

+1

tmaly1980:
I think it actually has something to do with react-native-cookies, the npm package seems to be outdated, so I had to import it from github directly, as well as using it with webkit requires passing some parameters. I actually didn't get it working until I called clearAll() two different times/ways, ie :

 ```
   CookieManager.clearAll(true).then((res) => {
      console.log('AFTER CookieManager.clearAll true =>', res);
      CookieManager.clearAll(false).then((res) => {
        console.log('AFTER CookieManager.clearAll false =>', res);
        dispatch(ac.authentication.logout(value));
      })
    })
```

PR #177, opened 2018-11-28
Use AndroidX instead of Support Library

### LINKED ISSUE #176: Use AndroidX instead of Support Library
AndroidX will replace the Support Libraries.

> AndroidX is the open-source project that the Android team uses to develop, test, package, version and release libraries within Jetpack.
>
> AndroidX is a major improvement to the original Android Support Library. Like the Support Library, AndroidX ships separately from the Android OS and provides backwards-compatibility across Android releases. AndroidX fully replaces the Support Library by providing feature parity and new libraries.

See: https://developer.android.com/jetpack/androidx/

wjwarren:
Fixes #176.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewFileProvider.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewFileProvider.java
@@ -1,6 +1,7 @@
 package com.reactnativecommunity.webview;
 
-import android.support.v4.content.FileProvider;
+
+import androidx.core.content.FileProvider;
 
 /**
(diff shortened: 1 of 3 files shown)


Titozzz:
Are you sure this would build without changing the gradle files and it's not only working because of some custom configuration ?

wjwarren:
@Titozzz Not sure what you mean. Could you elaborate?

Titozzz:
So I checked with some people, we shouldn't not use androidX and old support libraries at the same time. So since react-native did not update yet, this will stay on hold. Will merge as soon as RN gets it :)

donnellycolton:
Since react-native has updated to androidx, can we expect to see this merge soon?

Titozzz:
This was released in version 6.X

Issue #68, opened 2018-10-02
Cookies/localStorage is not shared across WKWebView:s


epiceros:
I opened an issue here:
https://github.com/facebook/react-native/issues/21450

Is this something that needs to be addressed here as well?

felixmeziere:
This is blocking us too :/

epiceros:
Hi! Some more info. We have an app with one main webview where, the user logs in. Then at times we overlay that webview with another webview, where the user needs to be logged in (so they are both open at the same time, and they are on the same domain). The webview displays fine, but the user is not logged in, in the overlaying webview. This is only the case for WKWebView, in UIWebView, and on Android, it just works.

Titozzz:
Ok, thanks for the details. I you want to start working on it I can review / merge pull requests 🚀

Titozzz:
Good insight, thanks

finalquest:
Me again.
We have been strugling with cookies errors on wkwebview from the start.
The way wkwebview handles cookies on iOS is pretty rare.
In the old UIWebview, the cookies are shared between all the network requests of the application.
So, for example, I logged to a server using fetch on RN, the webview can keep a session without login, because the cookie session is already setted by the first fetch.

WKWebview dosnt work like this. I've made several attempst to fix that on the react-native-wkwebview repo, but I failed every time.

So, if anyone has any insight on this, I will gladly accept it

Regards

KoenLav:
This should already be fixed here:

https://github.com/CRAlpha/react-native-wkwebview/commit/3e0162987ac7bf4175f94625edc44565d107fe2d

Please consider replacing WKWebView with this better maintained version, per my suggestion here: https://github.com/react-native-community/react-native-webview/issues/105

Issue #25, opened 2018-09-10
MIGRATED: Cookie not working with Payment Geteway. (#20470)


jamonholmgren:
Migrated from https://github.com/facebook/react-native/issues/20470

@abdelhakimkotti says:

---

Environment
React Native Environment Info:
System:
OS: macOS High Sierra
CPU: 2,53 GHz Intel Core i5
Memory: 8 Go 1067 MHz DDR3
IDEs:
Android Studio: Version 3.0
Xcode: Version 9.2

Description
Am implementing party payment gateway in a webview And when the user got redirected to the result page, i got a message saying that "cookies are not enabled in the browser ".
To access the payment pages you need to have enabled cookie.

Code
     <WebView 
          source={{ uri: this.state.url }}
          scalesPageToFit={true} 
          javaScriptEnabled={true}
          thirdPartyCookiesEnabled={true} 
          onNavigationStateChange={this.onNavigationStateChange} 
     />
Reproducible Demo
For iOS:
https://developer.apple.com/documentation/foundation/nshttpcookiestorage/1418390-cookies?language=objc

For Android:
https://developer.android.com/reference/java/net/CookieStore

---

Salakar:
Hey @abdelhakimkotti I'm going to go ahead and close this given that progress is being in #181 for this and the age of this issue.

If you need this to remain open for whatever reason please let me know, otherwise please follow along with the PR mentioned above.

PR #181, opened 2018-11-29
feat(WKWebView): [ios] RNCWKWebView send headers on all page requests

### LINKED ISSUE #4: MIGRATED: WebView does not send custom request header after the first initial page load (#8693)
Migrated from React Native core: https://github.com/facebook/react-native/issues/8693

@chiraggshah says:

---

The source prop for WebView allows for passing request headers. When the webview renders for the first time, the request headers are sent.

But when the links within that page are clicked, the headers are not sent to the server in the request header.
Similarly, when the goBack() method is called on the WebView, the request headers are not being sent.

---

_More context available in the original issue._

mattycas:
This enables the WKWebView implementation to send the headers set on the `source` prop on every page load request rather than just the initial load request.

Fixes #4, #68, and #25


Titozzz:
Hello, thank you for this, I've seen it requested a lot, could the same be done for android ?

mattycas:
Hey @Titozzz thanks for your quick reply. It was my understanding that android would send through the headers for all GET requests. Is that not right?

Understandably this is a fairly controversial change/implementation and a breaking one as well. Should I add a prop to RNCWKWebView to enable this behaviour?

Titozzz:
If you just align iOS functionality to android, it might be a bugfix. If we change behavior, we can do breaking change easily with semver anyway so don't worry too much about it.

jamonholmgren:
Note that Android does _not_ set headers on subsequent page loads. This will need to be done for Android as well.

@Titozzz Do you want to require that we do this for Android before merging or are you okay merging with iOS only?

Secondly, do we want to add a feature flag? I think it's fine to publish without a flag, but it is a breaking change.

Titozzz:
If there is a way to have this on android at the same time it would be awesome, as the difference between the two would be quite annoying I think ?

If we do both at the same time I don't think feature flag is necessary, but if it's only on iOS it would prob be better

ralesiupgrade:
Hey hows it going? I been following along in the chain of PRs and issues and was having an issue in iOS similar to https://github.com/react-native-community/react-native-webview/issues/68 and it looks like this will fix it. Any progress on getting this merged :) My android build works fine but looks like for now for iOS i will have to use `useWebKit={false}` until this is merged. Thanks

smathson:
I'd like to argue for this being opt-in behavior and not the default. Any site that uses Authentication headers and also whitelists third-party domains for valid reasons (think youtube embeds) would be unintentionally leaking authentication tokens by default without a clear way to opt-out unless I'm missing something.

Also since the prop that gets passed is a `source` object with `uri` and `headers`, it seems a bit unclear that the headers would also be implicitly sent along to potentially different hosts. Moving the behavior to an optional prop of headers that get merged into every request (i.e. `defaultHeaders`) and then noting the potential security risks might be a more "secure-by-default" and flexible approach. 

This way someone could attach authentication headers to specific requests by changing the `source` prop and still provide a list of default headers that are fine to send with every request OR go ahead and add authentication headers to the `defaultHeaders` and only whitelist trusted domains.

mattycas:
I agree @smathson . If @Titozzz and @jamonholmgren agree I am happy to put this behind a feature flag prop.
I could also limit it to only user initiated actions on the same host as the source request's host?

Titozzz:
What would be the best way to achieve something that would work for everyone ?

Setting headers on a host basis ? on a request basis ? Apparently not the same header for all requests 😛 .

smathson:
@mattycas That sounds like it makes sense to me as long as the headers are only sent along to the original host. I guess it would be possible for someone to update just the `uri` on the `source` prop object and forget about `headers` still being set, but that seems like an edge case to me. Maybe we could add a note to the docs to be careful to set the entire `source` prop when using authentication headers to avoid leaking tokens.

smathson:
@jamonholmgren I like that. Makes it explicit and makes people think about which headers they are attaching for which requests.

mattycas:
@jamonholmgren this is a nice solution and I'm happy to update the PR.

Titozzz:
Is there a way to achieve the same behavior on android?

jamonholmgren:
On Android, it looks like we can set custom headers by overriding `shouldInterceptRequest` (see [here](https://developer.android.com/reference/android/webkit/WebViewClient.html#shouldInterceptRequest) and the [Stackoverflow answer](https://stackoverflow.com/a/29811280/204044) I saw first). 

We should probably make the behavior the same across iOS and Android.

chiraggshah:
@mattycas : It would be nice to know the reason why this was closed, since this had active discussion around it? Was there a separate PR which caused this to be closed?

jamonholmgren:
I'd love someone to champion this PR and move it forward. Any volunteers?

Issue #187, opened 2018-11-30
JavaScript alert not working in IOS


npm-ued:
Sample Code

```
const html = '<!DOCTYPE html><html><head></head/><script>alert(11);</script><body><h1>This is a heading! test</body></html>';
return (
      <SafeAreaView style={{flex: 1, backgroundColor: '#fff'}}>
        <View style={{flex: 1}}>
          <WebView 
            source = {{html: html}}
            onLoadProgress = {e => console.log(e.nativeEvent.progress)}
          />
        </View>
      </SafeAreaView>
);
```

npm-ued:
Sample Code
`
const html = '<!DOCTYPE html><html><head></head/><script>alert(11);</script><body><h1>This is a heading! test</body></html>';
    return (
      <SafeAreaView style={{flex: 1, backgroundColor: '#fff'}}>
        <View style={{flex: 1}}>
          <WebView 
            source = {{html: html}}
            onLoadProgress = {e => console.log(e.nativeEvent.progress)}
          />
        </View>
      </SafeAreaView>
      
    );
`

PR #188, opened 2018-12-03
Fix JavaScript alert not working in IOS

npm-ued:
fix this bug #187


npm-ued:
@Titozzz

Titozzz:
How would we manage multiple languages?

jamonholmgren:
Before (no alert):

image

After:

image

Issue #80, opened 2018-10-16
Can not download or open any file from HTML download Attribute in android webview 


ankur-sardar:
If we go to this link - https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_a_download
And Click on the image (which is a download attribute HTML tag) , It does not download anything.

ankur-sardar:
Should we support this feature in android webview? Create some Download listener in native side.
We can use `android.webkit.DownloadListener`

jamonholmgren:
@ankur-sardar Could you edit your original issue body and provide an example of what you're trying to do that isn't working?

vitorreis:
Hello there, I am not using this package yet (waiting for https://github.com/react-native-community/react-native-webview/pull/60 be merged), however I have the problem with the default webview in react-native, that nothing happens when the user clicks in any download link in the webview, nothing happens...

It seems that the same behaviour happens with this project as reported by @ankur-sardar ...
Any plans on implementing this feature?

Titozzz:
This is a duplicate of #33 , will be fixed soon hopefully !

ankur-sardar:
@Titozzz This issue is regarding the "Downloading" option for android webview (Not Upload as you referred the issue)
Are you going to support Downloading events also in #33 ?

Titozzz:
Oh my bad, sorry, reopening, however we should wait for #33 to land before starting this I think. It will probably be way easier to implement than upload (I hope 😮)

smathson:
@Titozzz @ankur-sardar see PR #203 re: download support on Android.

smathson:
This has been implemented and released in 2.15.0. @Titozzz or @jamonholmgren, this issue can be closed.

PR #203, opened 2018-12-10
Add file download support for Android

smathson:
Addresses #80.

Caveat: I am not an Android developer. This code comes from a fork of the original RN WebView that we have been using in production for some time, so all credit goes to @Oblongmana: https://github.com/Oblongmana/react-native-webview-file-upload-android.

Setting up a DownloadManager for the WebView is pretty straightforward, as is adding any known cookies to the request. Most of the complication comes from the requirement after SDK 23 to ask the user for the WRITE_EXTERNAL_STORAGE permission. Unfortunately there is no mechanism to suspend the download request until permission is resolved so this code stores off the request and sets up a listener that enqueues the download once permissions are resolved so the user experience is really nice.

I didn't see anything in the way of tests or documentation that needs to be added for this change, so let me know if I missed anything. Thanks!

### DIFF
--- a/docs/Guide.md
+++ b/docs/Guide.md
@@ -105,3 +105,28 @@ WebView.isFileUploadSupported().then(res => {
 
 ```
 
+### Add support for File Download
+
+##### iOS
+
+For iOS, all you need to do is specify the permissions in your `ios/[project]/Info.plist` file:
+
+Save to gallery:
+```
+<key>NSPhotoLibraryAddUsageDescription</key>
+<string>Save pictures for certain activities.</string>
+```
+
+##### Android
+
+Add permission in AndroidManifest.xml:
+```xml
+<manifest ...>
+  ......
+
+  <!-- this is required to save files on Android  -->
+  <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+
+  ......
+</manifest>
+```
--- a/docs/Guide.md
+++ b/docs/Guide.md
@@ -50,7 +50,7 @@ class MyWeb extends Component {
 }
 ```
 
-### Add support for File Upload
+### Add support for File Upload / Download
 
 ##### iOS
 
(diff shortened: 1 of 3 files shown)


jamonholmgren:
Hey @smathson, this looks really cool! I'm cloning down your fork and trying it out now.

jamonholmgren:
This is very cool. Screenshots:

image

image

Titozzz:
@andreipfeiffer Would you mind reviewing this ? You spent a lot of time doing the upload so you probably have useful insights on this! 😄

andreipfeiffer:
Hey guys,

Great work @smathson :D
In this situation, I guess that the user would like the downloaded file to be in a shared folder, that he could open/read with other apps. This means that the `WRITE_EXTERNAL_STORAGE` permission is mandatory here.

I would add in the docs that this permission is required for downloading files.

I will also test this in the next days.

smathson:
@andreipfeiffer Good idea. Do you think we should add it to https://github.com/react-native-community/react-native-webview/blob/master/docs/Guide.md alongside the File Upload support or somewhere else?

andreipfeiffer:
@smathson you're right, it should definitely go in there.

smathson:
Added documentation re: permissions and merged master.

docs/Guide.md
```diff
@@ -50,7 +50,7 @@ class MyWeb extends Component {
 }
 ```
 
-### Add support for File Upload
+### Add support for File Upload / Download
```
andreipfeiffer: I recommend to put the `Download` section separately, because users might get the impression that they also need to add the iOS settings for the Download, as well.

andreipfeiffer:
Hey @smathson,
I added a comment regarding the docs.

Also, is the `download` attr supported in iOS by default?
Is so, it should be specified in the docs.
If some setting is necessary, it should be specified what is needed to make it work.
It it's not supported, it should also be specified, and hopefully somebody can step-up to add support for it.

docs/Guide.md
smathson: @andreipfeiffer How does that look? iOS only needs the permission added for download.

Titozzz:
@allcontributors[bot] please add @smathson for code, doc

PR #204, opened 2018-12-10
Add example file upload app

vitorreis:
I am implementing this webview in an existing project, and I found helpful to have an example application to test some stuff like `onMessage` callback and the android file upload.

In case you think that could be helpful to other developers, consider merging this pr


jamonholmgren:
Hey @vitorreis , thanks for submitting this PR! At this time, we are not going to be merging in any full example apps to this repository (ref #57). I would recommend updating [the Guide in docs](https://github.com/react-native-community/react-native-webview/blob/master/docs/Guide.md) instead.

PR #219, opened 2018-12-28
fix: remove webview style flex 1, make height work properly

### LINKED ISSUE #216: WebView style height is not working
When I only use WebView, the style height is not working in both iOS and Android.
However, it can work correctly when I use maxHeight or wrap Webview inside View and set View height.
Is this a bug?

For example:
render() {
    return (
        <WebView
            source={{ uri: "https://github.com/react-native-community/react-native-webview" }}
            style={{ width:300,height:300 }}
            onLoadProgress={e => console.log(e.nativeEvent.progress)}
        />
    )
}
image

Chi-AnTai:
fix #216
The height didn't work because WebView has style flex 1 , remove it and height will work properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "Thibault Malbranche <malbranche.thibault@gmail.com>"
   ],
   "license": "MIT",
-  "version": "2.14.2",
+  "version": "2.14.3",
   "homepage": "https://github.com/react-native-community/react-native-webview#readme",
   "scripts": {
     "test:ios:flow": "flow check",


jamonholmgren:
Thanks for the PR, @Chi-AnTai ! This will break backwards compatibility due to lacking any sort of height. Consider the following setup:

```jsx
  render() {
    return (
      <View style={{ flex: 1 }}>
        <WebView source={{ uri: "https://infinite.red" }} />
      </View>
    );
  }
}
```

With your branch:



This is with the current release, 2.14.3:



I would recommend that we not merge this, but rather just document that people should add `flex: null` to their styles if they want to use a `height` instead.

```jsx
  render() {
    return (
      <View style={{ flex: 1 }}>
        <WebView
          source={{ uri: "https://infinite.red" }}
          style={{ height: 200, flex: null }}
        />
      </View>
    );
  }
}
```

PR #225, opened 2019-01-02
add style doc in reference.md

Chi-AnTai:
Add style doc, as discuss in #216, mention that if you need to add `flex: null` to make `height` work.


Titozzz:
I think you might want to say instead that default styles are applied, list them, and say that if you want fixed height you can use flex: 0.

Thanks :)

PR #243, opened 2019-01-12
[WIP] Simplified postMessage implementation.

jordansexton:
This PR:

- incorporates @KoenLav's commits from #104
- merges the latest from master @ cf039a550b7206082b7dc2e17ad5d5834d3eccb2
- removes all the overwriting of `window.postMessage`
- implements and documents `window.ReactNativeWebView_postMessage` as described @ https://github.com/react-native-community/react-native-webview/pull/104#issuecomment-452132323
- on iOS WKWebView, proxies `window.ReactNativeWebView_postMessage` to `window.webkit.messageHandlers.ReactNativeWebview.postMessage`
- on iOS UIWebView, uses the namespaced private API `window.ReactNativeWebview_messageReceived` instead of using `dispatchEvent(new MessageEvent('message:received'))` globally (I imagine the related navigation mechanism exists because of some missing capability for handling communication with UIWebView?)

I haven't had a chance to test these changes yet, but I wanted to get it up for review.


KoenLav:
@jordansexton thanks for taking this up!

On first glance this looks really good, I'll review in more detail in the next few days.

jordansexton:
While testing I found that [`addJavascriptInterface`](https://developer.android.com/reference/android/webkit/WebView#addJavascriptInterface(java.lang.Object,%20java.lang.String)) expects an object rather than a function.

The latest commit implements `window.ReactNativeWebview` as a global object instead, with `postMessage` as a method.

ios/RNCUIWebView.m
```diff
@@ -246,7 +246,7 @@ - (BOOL)webView:(__unused UIWebView *)webView shouldStartLoadWithRequest:(NSURLR
       @"data": data,
     }];
 
-    NSString *source = @"document.dispatchEvent(new MessageEvent('message:received'));";
+    NSString *source = [NSString stringWithFormat:@"window.@.messageReceived();", MessageHandlerName];
```
KoenLav: This makes a lot more sense than the previous implementation :)

Here, within native code, window.ReactNativeWebview.messageReceived is called inside the WebView to let it know we have finished processing a message.

ps I do wonder how much time we should invest in this particular WebView; it's supposed to be replaced with WKWebView, I think.

@Titozzz?

ios/RNCWKWebView.m
```diff
@@ -88,7 +88,20 @@ - (void)didMoveToWindow
...
+      [wkWebViewConfig.userContentController addScriptMessageHandler: self name: MessageHandlerName];
+
+      NSString *source = [NSString stringWithFormat:
+        @"window.@ = {"
+         "  postMessage: function (data) {"
+         "    window.webkit.messageHandlers.@.postMessage(String(data));"
+         "  }"
+         "};", MessageHandlerName, MessageHandlerName
+      ];
+      [self injectJavaScript: source];
```
KoenLav: This will make sure window.ReactNativeWebview.postMessage is defined, the same way it is defined in Android, which is great!

In here we don't need messageReceived because we can pass messages without changing the location.

So, this all looks fine!

ios/RNCWKWebView.m
```diff
@@ -122,6 +135,8 @@ - (void)didMoveToWindow
     [self addSubview:_webView];
     [self setHideKeyboardAccessoryView: _savedHideKeyboardAccessoryView];
     [self visitSource];
+  } else if (_messagingEnabled) {
```
KoenLav: When is this code called?

Should this be NOT _messagingEnabled?

ios/RNCWKWebView.m
```diff
@@ -171,7 +186,7 @@ - (void)setBackgroundColor:(UIColor *)backgroundColor
 
 /**
  * This method is called whenever JavaScript running within the web view calls:
- *   - window.webkit.messageHandlers.[MessageHanderName].postMessage
+ *   - window.webkit.messageHandlers[MessageHandlerName].postMessage
```
KoenLav: Nice catch 👍

ios/RNCWKWebView.m
jordansexton: It looks like that line was added @ https://github.com/react-native-community/react-native-webview/commit/e0fe48b716e8b9278c6c78e8b0902f7b1ec91eaa

I don't know exactly what that whole conditional block is supposed to do so I left it alone for now. My reading of it (which may be totally off) was something like

```objective-c
  // stuff that happens to setup the webview
} else {
 // stuff that happens to teardown the webview
}
```

If this is accurate, the teardown check `else if (_messagingEnabled)` makes sense, because it's removing the script handler that would have been added in setup if messaging was enabled.

jordansexton:
I can confirm things work on Android in testing. I'll be continuing testing on iOS tomorrow.

ios/RNCUIWebView.m
Titozzz: Yeah we will probably drop UIWebview soon. I've wanted to do it for so long anyway 😢

android/build.gradle
```diff
@@ -48,6 +48,7 @@ android {
 repositories {
   mavenCentral()
   jcenter()
+  google()
```
Titozzz: @Salakar Do we need this ?

js/WebView.android.js
```diff
@@ -152,8 +152,8 @@ class WebView extends React.Component<WebViewSharedProps, State> {
         javaScriptEnabled={this.props.javaScriptEnabled}
         thirdPartyCookiesEnabled={this.props.thirdPartyCookiesEnabled}
         domStorageEnabled={this.props.domStorageEnabled}
-        messagingEnabled={typeof this.props.onMessage === 'function'}
         onMessage={this.onMessage}
+        messagingEnabled={typeof this.props.onMessage === 'function'}
```
Titozzz: avoid theses kind of changes as they will mostly just create merge conflicts

android/build.gradle
Salakar: Nope, there's only 1 dependency, and it's `org.jetbrains.kotlin` which comes from Maven central: https://mvnrepository.com/artifact/org.jetbrains.kotlin

jordansexton: I added this line because it was suggested by Android Studio to resolve a Gradle sync error when first syncing the project from master. I can try to replicate and provide steps.

js/WebView.android.js
jordansexton: I kept these from @KoenLav's changes and resolved the conflicts when merging from master. Let me know if you want me to change it back.

jordansexton:
This is confirmed working on iOS now. 7c1fd0a20212870097e5fd5e59c845aec33bcb54 was necessary because `[self.webview evaluateJavaScript:...]` doesn't do anything, and never calls its callback, if called before the page loads.

android/build.gradle
jordansexton: This is what I get when cloning master and opening the `android` directory in Android Studio:

screen shot 2019-01-17 at 12 13 34 am

screen shot 2019-01-17 at 12 12 35 am

jordansexton:
I haven't tested the UIWebView changes yet, but I will in the next day or
two, and then squash things here, unless you prefer to do that (or leave as
is).

Titozzz:
Nono take your time, just ping me when all good :)

jordansexton:
Tested and working on UIWebView, WKWebView, and Android. Good to go!

docs/Reference.md
```diff
@@ -194,9 +194,9 @@ Function that is invoked when the `WebView` is loading.
 
 ### `onMessage`
 
-A function that is invoked when the webview calls `window.postMessage`. Setting this property will inject a `postMessage` global into your webview, but will still call pre-existing values of `postMessage`.
+Function that is invoked when the webview calls `window.ReactNativeWebview_postMessage`. Setting this property will inject this global into your webview.
```
KoenLav: window.ReactNativeWebview.postMessage?

docs/Reference.md
```diff
@@ -194,9 +194,9 @@ Function that is invoked when the `WebView` is loading.
...
 
-`window.postMessage` accepts one argument, `data`, which will be available on the event object, `event.nativeEvent.data`. `data` must be a string.
+`window.ReactNativeWebview_postMessage` accepts one argument, `data`, which will be available on the event object, `event.nativeEvent.data`. `data` must be a string.
```
KoenLav: Same?

ios/RNCWKWebView.m
```diff
@@ -88,7 +88,22 @@ - (void)didMoveToWindow
         wkWebViewConfig.processPool = [[RNCWKProcessPoolManager sharedManager] sharedProcessPool];
     }
     wkWebViewConfig.userContentController = [WKUserContentController new];
-    [wkWebViewConfig.userContentController addScriptMessageHandler: self name: MessageHanderName];
+
+    if (_messagingEnabled) {
+      [wkWebViewConfig.userContentController addScriptMessageHandler:self name:MessageHandlerName];
+
+      NSString *source = [NSString stringWithFormat:
+        @"window.%@ = {"
```
KoenLav: Would it not be better/more consistent to do:

window.%@ = window.webkit.messageHandlers.%@

KoenLav: This is definitely not the same as it is now; if window.webkit.messageHandlers.%@ is not defined when this code runs it won't work, but if we know it always will be then the code I suggested will be more in line with the Android implementation.

KoenLav: Thinking of this some more; what if people decide to extend the window.ReactNativeWebview object; is there any chance their code might run before the code here is run (same question for Android)?

Should we check whether this object exists and if so only add the postMessage method? 

I don't think that this requires any changes at this point, but I'm just considering scenarios where this might cause issues. 

If this is a problem, then a simple fix would obviously be for end users not to extend the object and create their own.

Titozzz:
@jordansexton can you review @KoenLav comments when you have time :)?

tapir:
Thanks for this work!! Can we finally get this merged and finally fix this 2.5 years old bleeding wound of a bug?

KoenLav:
@jamonholmgren @Titozzz do note that both of my comments are nitpicking; as far as I am concerned this can be merged as is (possibly improved upon later).

docs/Reference.md
jordansexton: Good catch! Resolved by 373cf36717461ba2c6461b925529e675732bc576

docs/Reference.md
jordansexton: Ditto, 373cf36717461ba2c6461b925529e675732bc576

jordansexton:
Anecdotal but FWIW, we've been testing the forked code for a week in our iOS and Android apps, and there have been no issues.

abendi:
Does this finally fix the issue where postMessage bridge would not be ready on document load? If so, can't wait for this to be merged.

Titozzz:
Alright, I consider this good to go, I'll make a few tests and get ready for merge 😄

Titozzz:
Quick update: I've had issues when I tested on android, so I need to dig deeper before I merge

Titozzz:
Alright, there was no bug, only stupid fail on my side. Proceeding to merge!

mkashlev:
When will this fix make it to react native build?

Titozzz:
I not sure I understand what you mean @mkashle. React-native will not pull back this repository. This repository is part of the lean-core effort to remove non-essential components from the core. You can read more here: https://github.com/react-native-community/discussions-and-proposals/issues/6

PR #244, opened 2019-01-12
ad android ci

gengjiawen:
build log: https://circleci.com/gh/gengjiawen/react-native-webview/30


jamonholmgren:
Thanks for your hard work on this, @gengjiawen!

I've been resistant to adding any example apps up to this point. (#57) Rather than checking the whole app into the repo, could we instead have the CI script spin up a new React Native app each time and then copy over the relevant files? For example, we could do something like this:

```shell
cd ./.circleci/builds
# this folder would be gitignored
react-native init WebTest
cp ../templates/App.js ./App.js
# do any other config
# install and link RN WebView
# now kick off build
```

This would make the build time slower, but would have the big advantage of always using the latest React Native on a fresh app.

Thoughts?

gengjiawen:
If someone changes the native code, the test still test against the main repo.

gengjiawen:
@jamonholmgren Thought on merge this ?

PR #250, opened 2019-01-17
Show/hide WebView scroll indicators

jordansexton:
When using a WebView in a hybrid app, it's often desirable to make it look like a native view that doesn't show scrollbars.

This implementation currently uses the iOS nomenclature but supports both platforms. I haven't tested these changes yet.


jordansexton:
Tested on WKWebView, UIWebView, and Android. On iOS it works as expected. On Android, it seems like there may or may not be an effect depending on manifest settings and the content of the WebView, but at worst it does nothing.

Titozzz:
Can someone investigate the real behavior on android ?

oliverdolgener:
I can confirm that it makes no difference on Android. Would be cool if it would work for Android as well but at least it does not break anything!

Titozzz:
Thanks for the feedback. WOuld be great to have it work on both before merging

oliverdolgener:
I didn't try the manifest thing. I will later and report back.

oliverdolgener:
I couldn't find anything within the manifest file that would enable the behaviour BUT when calling `webView.setVerticalScrollBarEnabled(false);` on line 356 in `RNCWebViewManager.java` it works. Maybe we could enable it from there somehow?

oliverdolgener:
Oh wait, I'm stupid! I forgot to set the prop to false. It DOES work on Android without putting anything in the manifest!

Titozzz:
so everything is fine?

oliverdolgener:
Yes, it works fine for me on Android! Haven't tested on iOS but @jordansexton did.

oliverdolgener:
I had an error on Kotlin Version below 1.3.0 but I think this was due to an older master branch version.

Titozzz:
should release as 5.2.0

jordansexton:
@Titozzz @oliverdolgener I was traveling, so thanks for following up on this! Glad it worked out.

Issue #251, opened 2019-01-17
The Android Gradle plugin supports only Kotlin Gradle plugin version 1.3.0 and higher


sheva007:
I am getting this error when building with gradle:

`* What went wrong:
The Android Gradle plugin supports only Kotlin Gradle plugin version 1.3.0 and higher.`

I had to solve it temporarily in my android build.gradle by adding this to the end to over write the original webview gradle file:
`subprojects {project ->
    if (project.name.contains('react-native-webview')) {
        buildscript {
            dependencies {
                classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.11"
            }
        }
    }
}`

Titozzz:
#267 Will solve that.

sheva007:
thanks, I saw a new version up, but they did not include #267  yet, waiting for it.

guhungry:
@Titozzz this issue can be closed.

lixiaopeng168:
你需要更改为 "org.jetbrains.kotlin:kotlin-stdlib-jdk7$kotlin_version"  注意是 jdk7 不是jre7

PR #260, opened 2019-01-19
Update kotlin version to 1.3.11

jordansexton:
I experienced the same problem described in #251 after updating Android Studio, Kotlin, and the Gradle plugin to 3.3 (necessary when using Android Studio's "Build Bundle/APK" feature).

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-  ext.kotlin_version = '1.2.71'
+  ext.kotlin_version = '1.3.11'
   repositories {
     google()
     jcenter()


android/build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-  ext.kotlin_version = '1.2.71'
+  ext.kotlin_version = '1.3.11'
```
hssrrw:
```suggestion
  ext.kotlin_version = rootProject.hasProperty('kotlin_version') ? rootProject.kotlin_version : '1.3.11'
```

How about allowing users to specify kotlin version the way we already do with SDK and build tools versions?
So that build.gradle could have something like this
```
buildscript {
    ext {
        buildToolsVersion = "28.0.3"
        minSdkVersion = 21
        compileSdkVersion = 28
        ...
        kotlin_version = "1.3.11"
    }
     ...
}
```

jerolimov:
Sorry, but why couldn't we replace the kotlin files with pure java files?

The main part of this project is written in java and only these 6 event files are written in kotlin. Sorry but I didn't see the benefit in this case?

[/src/main/java/com/reactnativecommunity/webview](https://github.com/react-native-community/react-native-webview/tree/master/android/src/main/java/com/reactnativecommunity/webview)
[/src/main/java/com/reactnativecommunity/webview/events](https://github.com/react-native-community/react-native-webview/tree/master/android/src/main/java/com/reactnativecommunity/webview/events)

Most react-native projects works with pure Objective-C (without Swift) and Java (without Kotlin) because this helps to keep the build system and app small and simple.

thinkmobilede:
for people who want to adapt their project before the pull request is merged: 

if you manually change the kotlin_version in your node_modules folder, you might have to re-import your project in Android Studio to update dependencies. On the Androd Studio start screen instead of opening an existing project, select "Import project" and select the project in your /android folder.

Titozzz:
@jerolimov the goal here with this project aside from being a webview is to set a new standard for react-native-community packages, plans are to change all the code to kotlin, typescript and eventually swift if the tooling allows it.

--- 

I'll test and merge this today.

thinkmobilede:
@Titozzz how are you going to handle the issues with different Kotlin and Swift versions in different dependencies? Not all repositories are maintained regularly. These issues nearly never happen with Obj-C and Java.

Titozzz:
Closing in favor of #267 

@thinkmobilede right now I will allow custom kotlinVersion variable, same as how you handle all the different versions of support libraries. Do that seems ok for you ?

nol13:
Well, I'll just post here instead of creating a new issue, in case is useful to someone.

1.3.11 appear not compatible with gradle 4.10.1, had to set kotlinVersion back to 1.2.71, even with no other kotlin dependencies. Java 1.8.

With gradle 4.10.3 can build with 1.3.11+ though.

during './gradlew assembleRelease' 4.10.1 gets..
```
Could not resolve project :react-native-webview.
     Required by:
         project :app
      > java.lang.NullPointerException (no error message)
```

PR #265, opened 2019-01-21
Add `androidHardwareAccelerationDisabled` prop

YangXiaomei:
Hardware acceleration is enabled by default if your Target API level is >=14. However, because hardware acceleration is not supported for all of the 2D drawing operations, turning it on might affect some of your custom views or drawing calls. To disable hardware acceleration in application level will slow down the app, so we'd better to add a prop to enable user to disable hardware acceleration for android platform in view level
 Please view the details in [Hardware acceleration](https://developer.android.com/guide/topics/graphics/hardware-accel#controlling)

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -452,6 +452,8 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   @ReactProp(name = "javaScriptEnabled")
   public void setJavaScriptEnabled(WebView view, boolean enabled) {
     view.getSettings().setJavaScriptEnabled(enabled);
+    view.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
+    FLog.w(ReactConstants.TAG, "Hello World from Margaret~~~~~~~~~~~~~~~~~~~~~");
   }
 
   @ReactProp(name = "overScrollMode")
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -529,6 +529,13 @@ public void setJavaScriptEnabled(WebView view, boolean enabled) {
     view.getSettings().setJavaScriptEnabled(enabled);
   }
 
+  @ReactProp(name = "androidHardwareAccelerationDisabled")
+  public void setHardwareAccelerationDisabled(WebView view, boolean disabled) {
+    if (disabled) {
+      view.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
+    }
```
Titozzz: maybe re-enable it in an else case ?

YangXiaomei: Thanks for your suggestion, from the guide [
Hardware acceleration](https://developer.android.com/guide/topics/graphics/hardware-accel#controlling)  currently user cannot enable hardware acceleration at the view level.

Titozzz: you are 100% right

Titozzz:
@allcontributors[bot] please add @YangXiaomei for code, doc

PR #270, opened 2019-01-24
Allow _target links to be opened up in the same way as UIWebView.

benjosantony:
Fixes https://github.com/react-native-community/react-native-webview/issues/139 .


benjosantony:
@Titozzz Anyway I could help assist in the review ?

Titozzz:
Code looks good, but I did not have time to test yet. Hopefuly soon

Issue #266, opened 2019-01-21
Problem with file download on Android 9


m-ochyra:
Hi,

I have problem with downloading files on physical Android 9 devices (eg. Motorola Moto One, on emulator everything works fine, and older Android devices also).

When I click on link to download file from web page, toast says that downloading is started, but after complete, this file is not visible in downloaded files on phone.

File downloading works well on pure Android implementation with WebView (the same page, link, device).

Is this problem with library or some configuration is required?

PR #273, opened 2019-01-24
Fix for issue #266

m-ochyra:
Fixed issue #266. Now downloading works on Motorola Moto One with Android 9.


Titozzz:
@smathson What do you think ?

smathson:
Looks good to me. Not sure why the original code was trying to parse the contentDisposition on its own and only falling back to guessFileName. Based on https://developer.android.com/reference/android/webkit/URLUtil#guessFileName(java.lang.String,%20java.lang.String,%20java.lang.String) that seems like exactly what guessFileName already does.

Issue #22, opened 2018-09-10
MIGRATED: WebView not scrollable when under ScrollView


jamonholmgren:
Migrated from https://github.com/facebook/react-native/issues/20347.

@mehulmpt says:

---

Environment
React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.3
      CPU: x64 Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz
      Memory: 29.76 MB / 8.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 8.1.4 - /usr/local/bin/node
      npm: 4.6.1 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 11.4, macOS 10.13, tvOS 11.4, watchOS 4.3
    IDEs:
      Android Studio: 3.1 AI-173.4720617
      Xcode: 9.4.1/9F2000 - /usr/bin/xcodebuild
    npmPackages:
      react: ^16.4.1 => 16.4.1
      react-native: ^0.56.0 => 0.56.0
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7
      react-native-rename: 2.2.2
Description
WebView component when placed under ScrollView loses its ability to scroll content. However, if scrollEnabled={false} is passed to ScrollView, WebView is scrollable again.

Reproducible Demo
This render function would suffice:

```js

export default class App extends Component {
  render() {
    return (
      <ScrollView contentContainerStyle={{ flexGrow: 1 }}>
        <WebView source={{ uri: 'http://localhost/somepage-with-scrollable-content' }}
                 style={{ height: 300 }}/> {/* this webview is not scrollable now */}
      </ScrollView>)
  }
}
```
This has also been mentioned in issue #10723 but it was closed without any solution

---

jamonholmgren:
@vvusts The way the core team handles issues in this repo is we will review them, tag them, get more info, and then ask the community for help fixing them. Then we will review, test, suggest changes, and eventually merge and deploy PRs that fix issues. So it's possible someone else might fix it, but if you want to try, it's much appreciated.

Salakar:
Hey @jamonholmgren @vvusts @mehulmpt @masako-yamada 👋, 

I've just tried to reproduce with the provided example and also slightly tweaked it but cannot reproduce on the latest webview:

[...]

PR #281, opened 2019-01-28
fix(android): Allow scrolling inside ScrollView

masako-yamada:
Try fixing #22 
This should work in general, however, there would be some edge cases where it doesn't work.
Hope others improve this.


MarkIgnaczakBB:
This seems to indeed fix this issue: https://github.com/react-native-community/react-native-webview/issues/22.

WebView now correctly scrolls and follows user's move gesture (e.g. such as if an image zooms in by user hovering their finger around).

vvusts:
I just tried this PR fixed webview scrolling when webview is inside scroll view.

N-Milinkovic:
@masako-yamada This has conflicts, but also doesn't have anyone set as reviewers, maybe need to add them to get their attention?

masako-yamada:
I fixed the conflict, @N-Milinkovic or somebody, would you check if this works?
I couldn't test it by myself since the latest master requires androidx (NN >= 60) and my project has not migrated to it. :-(

N-Milinkovic:
> I fixed the conflict, @N-Milinkovic or somebody, would you check if this works?
> I couldn't test it by myself since the latest master requires androidx (RN >= 60) and my project has not migrated to it. :-(

I have tested in a RN >= 60 project and it works just need approval from one of the reviewers

gasolin:
@jamonholmgren / @Titozzz the conflict is fixed, would you mind to review this PR?

brentvatne:
can you explain your PR in more detail?

i'm not sure this is desirable, my understanding is that it always makes the webview scrollable when inside of a scrollview. you could use react-native-gesture-handler to coordinate this instead - wrap the WebView in `NativeViewGestureHandler` and set a ref on it, then use the `ScrollView` imported from rngh and use the `waitFor` prop

jamonholmgren:
@brentvatne Can you provide an example code snippet that shows that approach in action?

brentvatne:
@jamonholmgren - can't repro the original issue from the code snippet posted in #22 op, any tips? https://snack.expo.io/@notbrent/authentic-watermelon

gasolin:
I use the example https://snack.expo.io/r1E6viX0N to test with snack version, the horizontal scroll is still sluggish on Android

brentvatne:
gotcha, here's a repro: https://snack.expo.io/@notbrent/insane-almond

couldn't get it working with RNGH so a solution may indeed be needed on the webview itself

Titozzz:
let's ship it, but I'll be monitoring issues to make sure no one is complaining about that change

brentvatne:
@Titozzz - maybe it makes sense to put the behavior behind a prop so it's easy for people to enable, then make it default in a major version bump?

wsliaw:
@Titozzz, agree with @brentvatne, make it a prop seems more reasonable

Titozzz:
It was causing too many issues for people (see #878) so I reverted it. Feel free to submit a new PR 😄

Titozzz:
The thing with a prop is that it's encouraging keep 2 half working versions instead of finding something that'll fix for everyone 😢

kefyru:
COOL!!!!

jordangrant:
Applied this PR manually, exactly what I needed, thanks

PR #285, opened 2019-01-29
Add canGoBack and canGoForward functions

smathson:
Needed this to build a native navigation bar that can update on `window.location.pushState/popState`. We send a message over the bridge when the location changes and need to call the native method to update the navigation bar state.

I'd like to add this for Android too but I can't seem to figure out how to write a `ReactMethod` with a return value that also has access to the native WebView instance in the same way that I was able to on the iOS side. Anyone have any ideas?

I skipped adding documentation until we figure out if we need an Android implementation or not.


SebiVPS:
Maybe this PR: #551 also fixes your problem.

Titozzz:
should we merge this iOS only?

PR #293, opened 2019-01-31
Support all valid URI schemes and add testing

MalcolmScruggs:
Currently URI schemes that contain +,-, or . are not supported. We are working on adding a custom scheme for WKWebView that contains a + character. It would never allow our scheme to pass the whitelist.

An additional bug was preventing unwanted characters before a valid scheme. For example fakehttp://example.com would pass the whitelist because the regex was not matching from the start of the string.

Also added support for Jest tests and added tests for the changes.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,12 @@
+module.exports = function (api) {
+    api && api.cache(false);
+    return {
+      env: {
+        test: {
+          presets: [
+            "module:metro-react-native-babel-preset"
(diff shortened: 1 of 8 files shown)


Titozzz:
Wow that's a great contribution. Thank you so much ! I'll review this ASAP ❤️

Titozzz:
@allcontributors[bot] please add @MalcolmScruggs for code, tool, test

PR #296, opened 2019-01-31
Reapplied #134 with additional checks to tackle unintentional scrolls

stammbach:
I've also had problems with the #134 fix just like #210, on a very long form page the view would scroll to the top every time a user closes the keyboard (also happens when the user clicks a checkbox) which is not very user friendly.

Im currently calculating the maximum content offset / viewport, and only scroll to **that** maximum if the current offset is overlapping it. On my current project it seems to work perfectly.


jamonholmgren:
Looks really good to me. Need someone to test & verify that it works, but otherwise 👍 .

stammbach:
Hey @jamonholmgren, I've updated the keyboardDisplacementFix method to catch some edge cases where the contentSize of the scrollView might be smaller than the frame, just to be safe.

stammbach:
Maybe one of you guys can help us with testing this 🙏@iamtommcc @malonguwa

Titozzz:
@ryanlntn You were the one requesting #210, can you test this one to make sure it's not breaking anymore?

Titozzz:
Tested and approved on my app ✅ , going to merge this now

ryanlntn:
Apologies for the belated response, been in the middle of a big launch prep. I can confirm that this is working for me as well! 😄

PR #303, opened 2019-02-01
feat(Android/iOS postMessage): refactoring the old postMessage implementation

Titozzz:
Clone of #302 😭


koofka:
Thanks greatly for this update.  Working with it currently and not catching any onMessage events in a code base that was previously catching in v3 & v4 using the call window.ReactNativeWebView.postMessage(JSON.stringify({ action, payload }));.  

Saw the messagingEnabled change but the onMessage function is definitely a function.

Any potential advice on what else to change / check?  Appreciated.

PR #317, opened 2019-02-05
Deleted onShouldStartLoadWithRequest platform reference

Momazo7u7:
according to https://github.com/react-native-community/react-native-webview/commit/b1b662628e3ceb5c89f1d105814f01c4aba0b701 it's supported on android and Ios so  no platform  reference is needed

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -260,9 +260,9 @@ On iOS, when [`useWebKit=true`](Reference.md#usewebkit), this prop will not work
 
 Function that allows custom handling of any web view requests. Return `true` from the function to continue loading the request and `false` to stop loading.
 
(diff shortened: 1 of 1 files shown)


Titozzz:
Please, can you specify that on Android it is not called on the first load?

Momazo7u7:
It is not being called on the first load on android 6.0.1 and 7.0.1, tested on https://www.google.com

jamonholmgren:
Yeah, add a note that it is not called on first load, and then it should be good to go.

Momazo7u7:
Done

Titozzz:
@allcontributors[bot] please add @Momazo7u7 for doc

PR #325, opened 2019-02-07
Support fullscreen video on Android

dvicory:
Fixes https://github.com/react-native-community/react-native-webview/issues/294.

Let me know if we want to add anything as part of this.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -5,12 +5,16 @@ import android.annotation.TargetApi;
 import android.app.DownloadManager;
 import android.content.Context;
 import android.content.Intent;
+import android.content.pm.ActivityInfo;
 import android.graphics.Bitmap;
+import android.graphics.Color;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Environment;
 import android.text.TextUtils;
+import android.view.Gravity;
 import android.view.View;
+import android.view.ViewGroup;
 import android.view.ViewGroup.LayoutParams;
 import android.webkit.ConsoleMessage;
 import android.webkit.CookieManager;
@@ -24,6 +28,7 @@ import android.webkit.WebResourceRequest;
 import android.webkit.WebSettings;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
+import android.widget.FrameLayout;
 
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.LifecycleEventListener;
@@ -137,7 +142,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   @TargetApi(Build.VERSION_CODES.LOLLIPOP)
   protected WebView createViewInstance(ThemedReactContext reactContext) {
     RNCWebView webView = createRNCWebViewInstance(reactContext);
-    webView.setWebChromeClient(new RNCWebChromeClient(reactContext));
+    webView.setWebChromeClient(new RNCWebChromeClient(reactContext, webView));
     reactContext.addLifecycleEventListener(webView);
     mWebViewConfig.configWebView(webView);
     WebSettings settings = webView.getSettings();
@@ -604,10 +609,52 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   }
 
   protected static class RNCWebChromeClient extends WebChromeClient {
-    protected ReactContext mReactContext;
+    protected static final FrameLayout.LayoutParams FULLSCREEN_LAYOUT_PARAMS = new FrameLayout.LayoutParams(
+            LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT, Gravity.CENTER);
 
-    public RNCWebChromeClient(ReactContext reactContext) {
+    protected ReactContext mReactContext;
+    protected View mWebView;
+
+    protected View mVideoView;
+    protected WebChromeClient.CustomViewCallback mCustomViewCallback;
+
+    public RNCWebChromeClient(ReactContext reactContext, WebView webView) {
       this.mReactContext = reactContext;
+      this.mWebView = webView;
+    }
+
+    @Override
+    public void onShowCustomView(View view, CustomViewCallback callback) {
+      if (mVideoView != null) {
+        callback.onCustomViewHidden();
+        return;
+      }
+
+      mVideoView = view;
+      mCustomViewCallback = callback;
+
+      mReactContext.getCurrentActivity().setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE);
+      mVideoView.setBackgroundColor(Color.BLACK);
+      getRootView().addView(mVideoView, FULLSCREEN_LAYOUT_PARAMS);
+
+      mWebView.setVisibility(View.GONE);
+    }
+
+    @Override
+    public void onHideCustomView() {
+      if (mVideoView == null) {
+        return;
+      }
+
+      mReactContext.getCurrentActivity().setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_PORTRAIT);
+      mVideoView.setVisibility(View.GONE);
+      getRootView().removeView(mVideoView);
+      mCustomViewCallback.onCustomViewHidden();
+
+      mVideoView = null;
+      mCustomViewCallback = null;
+
+      mWebView.setVisibility(View.VISIBLE);
     }
 
     @Override
@@ -659,6 +706,10 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
       boolean allowMultiple = fileChooserParams.getMode() == WebChromeClient.FileChooserParams.MODE_OPEN_MULTIPLE;
       Intent intent = fileChooserParams.createIntent();
       return getModule(mReactContext).startPhotoPickerIntent(filePathCallback, intent, acceptTypes, allowMultiple);
+    }
+
+    protected ViewGroup getRootView() {
+      return (ViewGroup) mReactContext.getCurrentActivity().findViewById(android.R.id.content);
     }
   }
 
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -137,59 +137,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   @TargetApi(Build.VERSION_CODES.LOLLIPOP)
   protected WebView createViewInstance(ThemedReactContext reactContext) {
     RNCWebView webView = createRNCWebViewInstance(reactContext);
-    webView.setWebChromeClient(new WebChromeClient() {
-      @Override
-      public boolean onConsoleMessage(ConsoleMessage message) {
-        if (ReactBuildConfig.DEBUG) {
-          return super.onConsoleMessage(message);
-        }
-        // Ignore console logs in non debug builds.
-        return true;
-      }
-
-
-      @Override
-      public void onProgressChanged(WebView webView, int newProgress) {
-        super.onProgressChanged(webView, newProgress);
-        WritableMap event = Arguments.createMap();
-        event.putDouble("target", webView.getId());
-        event.putString("title", webView.getTitle());
-        event.putBoolean("canGoBack", webView.canGoBack());
-        event.putBoolean("canGoForward", webView.canGoForward());
-        event.putDouble("progress", (float) newProgress / 100);
-        dispatchEvent(
-          webView,
-          new TopLoadingProgressEvent(
-            webView.getId(),
-            event));
-      }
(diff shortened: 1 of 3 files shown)


smathson:
@dvicory This works great! Wish I had seen it earlier this week before I did almost identical work on my own fork. Your implementation is much cleaner though, so going to use yours and hope it gets merged soon. Nice work!

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -219,6 +210,113 @@ public void setUrlPrefixesForDefaultIntent(ReadableArray specialUrls) {
...
+    public void onShowCustomView(View view, CustomViewCallback callback) {
+      if (mVideoView != null) {
+        callback.onCustomViewHidden();
+        return;
+      }
+
+      mVideoView = view;
+      mCustomViewCallback = callback;
+
+      mReactContext.getCurrentActivity().setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE);
```
smathson: Trying to force the activity to landscape orientation seems to cause an issue with some video players (i.e. YouTube) if the video was recorded in portrait orientation. In my testing the video starts to go fullscreen and then immediately closes as (I assume) the video player tries to change the orientation to portrait. Removing this line starts all videos at the orientation of the parent view and then allows the user to change the orientation of the device if they wish which seems to fix the issue.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -219,6 +210,113 @@ public void setUrlPrefixesForDefaultIntent(ReadableArray specialUrls) {
...
+      mWebView.setVisibility(View.GONE);
+    }
+
+    @Override
+    public void onHideCustomView() {
+      if (mVideoView == null) {
+        return;
+      }
+
+      mReactContext.getCurrentActivity().setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_PORTRAIT);
```
smathson: Also removed this line in our fork for the same reason stated above.

Titozzz:
I'll wait for updates from @dvicory to merge, but many thanks for testing @smathson

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
dvicory: Interesting. I didn't test with portrait video - I'll give that a go. What I do know is that when I tested on an older device with a custom ROM (so take with a grain of salt), is that landscape videos on YouTube had graphical glitching when minimizing the video, unless a rotation happened at some point after. I figured most videos would be landscape anyway, and forcing the orientation seemed to fix it.

I will try to test tomorrow at work on more of our QA phones to see if I can get more information on what's going on. Maybe it would be reasonable to choose landscape if it's a landscape video and portrait if not? Or would you prefer that, if possible, no orientation is forced?

smathson: I think it's reasonable to change the orientation based on the content of the video if we can do that. I'll bet the vast majority of users will re-orient the device for landscape videos anyway. It also looks like if you open a portrait video and then turn the phone to landscape the video closes on it's own which seems really strange.

dvicory: After some more research, I've found the cause of the video glitching after minimizing. I don't know why it happens, but if you don't hide the system UI/status bar, and minimize from fullscreen, it glitches hard (the video acts as though it is larger than the `<video>` element, making the controls impossible to reach unless you get through them by rotating). Was able to reproduce reliably across two different devices. Sticky immersive mode fixes this (and seems to be what Chrome uses), but we have to handle the user possibly switching apps or locking the screen and coming back to re-hide the system UI. That's only available in KitKat or newer, so I have to figure out how it works for pre-KitKat. In some cases in multi-window, it still glitches. I don't yet know how to fix that.

I didn't have much luck at being able to determine whether a video is portrait or landscape. Someone with more Android chops might need to do that. So no more forcing orientation.

I'll be making these changes within the next day or two hopefully.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -381,57 +479,7 @@ protected RNCWebView createRNCWebViewInstance(ThemedReactContext reactContext) {
...
-      @TargetApi(Build.VERSION_CODES.LOLLIPOP)
-      @Override
-      public boolean onShowFileChooser(WebView webView, ValueCallback<Uri[]> filePathCallback, FileChooserParams fileChooserParams) {
-        String[] acceptTypes = fileChooserParams.getAcceptTypes();
-        boolean allowMultiple = fileChooserParams.getMode() == WebChromeClient.FileChooserParams.MODE_OPEN_MULTIPLE;
-        Intent intent = fileChooserParams.createIntent();
-        return getModule().startPhotoPickerIntent(filePathCallback, intent, acceptTypes, allowMultiple);
-      }
-    });
+    webView.setWebChromeClient(new RNCWebChromeClient(reactContext, getModule(), webView));
```
johman10: I'm definitely no Android expert but I was able to get this code into our project by taking this PR and the [Custom Android Guide](https://github.com/react-native-community/react-native-webview/blob/master/docs/Custom-Android.md) which worked really well!

The only thing that I had to change was change this `getModule` with `null` instead. I am really not sure why and how this works, but I figured to comment on it in case this might be relevant for others, or a potential bug.

The problem was basically that the `aPackage` was undefined in the `getModule` function which caused it to throw an error. As far as I can see that could have consequences for file upload.

dvicory: Hmm, I haven't had a problem with this in my example app. The file upload implementation is definitely different than what we were rolling with on our fork before we started to use `react-native-webview`. I would be glad to look into this more if I had a reproducible demo, though unless others have an issue with this too I don't believe it should be a blocker for merging this PR.

dvicory:
@Titozzz and @smathson sorry for the very late followup, but I've addressed the original issue of forcing orientation. Currently, I've made the best fullscreen experience only for KitKat or greater. I hope that's acceptable as getting it working well pre-KitKat turned out to be quite a challenge. Fullscreen video will still work there, just with the navigation and status bars.

Right now it uses sticky immersive mode (for KitKat and above), doesn't force orientation, and works well if a user locks the screen or backgrounds the app. Let me know if I can address anything further.

smathson:
@dvicory Looks great to me. I won't have a chance to test this in our production app anytime soon but I'll try to run it through a test app tomorrow or early next week.

Titozzz:
@dvicory Any change you can get a look at the conflicts? I merged quite a lot of stuff recently

dvicory:
@Titozzz Rebased and resolved conflicts. I've tested on the following devices we have:

Nexus 7 with Android 4.3
Nexus 5 with Android 6.0.1
Samsung S7 with Android 7.0
Samsung S9 with Android 8.0
Pixel 2 with Android 9.0

Couldn't find our phone with KitKat unfortunately.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
dvicory: This changed after some changes on master, and I've rebased and made this better. Now I no longer pass in the module here, and instead `getModule(reactContext)` from within the `RNCWebChromeClient` instance like is done elsewhere now. Shouldn't be a problem any longer.

Titozzz:
Great, thanks

Strate:
did you checked how embedded youtube works in fullscreen? I've tried your PR, but there was an issues:
1. Sometimes "fullscreen" button is not tapable on real device (Xiaomi Redmi 6A)
2. When button is tapable, video missing on fullscreen, only audio available.

dvicory:
@Strate I experienced the same problem on my Nexus 6P with a custom ROM (Pixel Experience), but attributed it to the custom ROM... Other Nexus, Samsung, and Pixel phones have no problems. I would appreciate any help in investigating what's going wrong here. I don't know that this should be a blocker to merging?

Titozzz:
@Strate @dvicory please keep me updated if this is a blocker or not :)

Strate:
@Titozzz actually, this happens with other webview components (I've tried https://github.com/teamairship/react-native-android-fullscreen-webview-video). Looks like this is a youtube-related problem, because Vimeo works great.

Titozzz:
So you are saying we can merge this as is?

Strate:
@Titozzz I can't say this, because actually I am not an android developer, I just tried to find out youtube fullscreen problem, found this PR and tested it )

adrianchinghc:
Manually applied the changes to test this out. This is for youtube and vimeo embed in a website. (Huawei Nova 3)

For youtube embed, the fullscreen icon is not greyed out anymore. But when I tap on the icon, nothing happens. Seeing this `mSecurityInputMethodService is null` in logcat.

As for vimeo embed, when I tap on the fullscreen icon, the screen becomes white colour. But the video is actually already full screen and playing below the white screen.

miltoneiji:
Is there anything that I can do to help this PR to get merged?

I was thinking of creating a repo with this branch and try to reproduce the bug. Would it help?

karvulf:
I tested it on my Galaxy Tab S2 with Android 7.0 and in fullscreen, my screen gets black, but I can hear the sound and clicking on the video also works

Titozzz:
If anyone can find the reasons for the last few bugs it would be great as it already looks like it's working on many devices

Titozzz:
If anyone can find the reasons for the last few bugs it would be great as it already looks like it's working on many devices

sibelius:
can we ship this as experimental? would this break other features?

Titozzz:
I could. I can always revert if something gets broken but I guess it should only affect people using full screen and since it's broken right now..

dvicory:
I could put this behind a JSX feature flag if people desire that, e.g. `fullscreenVideoEnabled`. Thoughts?

I haven't had time to jump into this more to see why some devices have issues with fullscreen video. 😞 It's probably something really nonsense...

ambrosiora:
> I could put this behind a JSX feature flag if people desire that, e.g. `fullscreenVideoEnabled`. Thoughts?
> 
> I haven't had time to jump into this more to see why some devices have issues with fullscreen video.  It's probably something really nonsense...

The flag `fullscreenVideoEnabled` sounds good to me.

I'm a Systems Analyst/Developer at a Brazilian University. Our app will run through many devices, at least 200 hundred thousand students will use it. We will put a log feature into entire app and if any user report errors with fullscreen or something like that, we will be able to retrieve some informations about OS, Brand, Android version, etc... I think is a good way to test on a large variety of devices.

What do you think about it?

sibelius:
let's ship behind a flag

dvicory:
I just added the flag in the latest commit: `allowsFullscreenVideo`. By default it's false. I've tested some corner cases on my Pixel 3 with switching this value at runtime, so far no bad effects. Welcome extra testing on this.

karvulf:
When I add this branch with "yarn add https://github.com/appfolio/react-native-webview.git#b41de3b", I get an Error because my lib folder doesn't exist in node_modules/react-native-webview. Is there another way to add that branch to my package.json?

dvicory:
@karvulf Since moving to TypeScript, the package has to be built. The way I locally developed with this branch is to have my branch locally checked out, `yarn install`, `yarn build`, then install `react-native-webview` by file reference in my example project.

I don't know of a good way to add my branch to your project right now that doesn't involve either publishing our fork or you adding `postinstall` scripts to handle the dependency specially (I don't recommend this). If you'd like I could create another branch for you that is installable.

karvulf:
That would be great and easier for me to test it! @dvicory

dvicory:
@karvulf Okay, I made a branch with the TypeScript buit.

You should be able to add it like so: `yarn add "github:appfolio/react-native-webview#built-android-fullscreen-video"`. Please let me know how it works out for you.

karvulf:
Thank you very much! It works @dvicory

dvicory:
@sibelius @smathson Anything more I can do to help push this through?

Titozzz:
Nope, you already rock 😍

MalcolmScruggs:
@allcontributors[bot] please add @dvicory for code doc

ambrosiora:
Nice job fellows!
Like I said above, if any user reports some unexpected behaviour , I will stat you guys.
Thanks!

ouabing:
Any ideas about get notified when enter/exit fullscreen? I'm seeking for a way on Android to unlock the orientations when the webview enters fullscreen mode to enable auto-rotate feature, and lock it again when exits. 

WKWebView handles this case well on iOS side, so it's a bit inconsistent on this. Temporarily my workaround is changing the Java code to emit exit and enter events. I don't know if it's a good idea to add another callback for Android : /.

dvicory:
@ouabing perhaps continue this discussion in #586.

Josephdias92:
Shows white screen when we disconnect internet in full-screen mode. Backbutton stops working.

Issue #213, opened 2018-12-17
[iOS WKWebView] crash on iOS 9.x due to WKWebView bug where it tries to use scrollView.delegate while being deallocated. 


malonguwa:
@Titozzz @NeilChan 
This is a old issue for WKWebView happens in iOS 9, due to WKWebView bug where it tries to use scrollView.delegate while being deallocated. 

Crash log from iOS 9.x
- [WKScrollView _updateDelegate] + 229  <-- (try to use scrollView.delegate after WKWebView dealloc)
- | WebKit | -[WKWebView dealloc] + 267

I already see other people's solution for this issue by `scrollView.delegate = nil `

But in our RN webView's current WKWebView.m,

Should I do something like this to avoid the WKWebView bug in iOS 9.x ?

```
- (void)removeFromSuperview
{
    if (_webView) {
        _webView.scrollView.delegate = nil;
        _webView.UIDelegate = nil;
        _webView.navigationDelegate = nil;
        [_webView.configuration.userContentController removeScriptMessageHandlerForName:MessageHanderName];
        [_webView removeObserver:self forKeyPath:@"estimatedProgress"];
        [_webView removeFromSuperview];
        _webView = nil;
    }

    [super removeFromSuperview];
}

```

Titozzz:
I know this issue is painful for you but please let's keep the focus in one thread #150 !

@NeilChan Can you have a look at this issue ?

NeilChan:
@Titozzz OK. It's a webkit bug in iOS9. I will try to fix it and submit a PR later.

Titozzz:
Thank you so much

jerolimov:
@malonguwa @NeilChan I created PR #342 to fix this and another iOS 9 issue.

If would be also easy and great if you can test and verify my PR. You find some instructions to test my branch in your project in the PR #342. Thank you.

PR #342, opened 2019-02-13
Fix crash with WebKit on iOS 9.x.

### LINKED ISSUE #150: [WKWebview] iOS 9 - Crash on unmounting

jerolimov:
This PR fixes two crashes when enableWebKit is true (which was the default) on iOS 9.x. The first one when the WebView component was mounted (fixes issue #150) and the second one, when the component was unmounted (fixes issue #213).

The first problem happen when the RNCWKWebView.m was loaded:

The programming guide for WebKit (version 2006) ([pdf](http://mirror.informatimago.com/next/developer.apple.com/documentation/Cocoa/Conceptual/DisplayWebContent/WebKit_DisplayWebContent.pdf)) from Apple said, that it is (was) required to check if the 'new' WebKit Framework was available. This was required when the WebKit framework was only available on Mac OS X (10.2) when the Safari was installed. 😆

Because WebKit is (currently..) a fix part of iOS we didn't need this check this anymore to determinate if WebKit is available. I also see no other reference that this is required in iOS so I just remove this code. Without this code the WebView works also on iOS 9.x.

The second issue happen when the WKWebView was removed from the native view hierarchy. Its required (only on iOS 9) to remove the UIScrollView delegate before cleaning up the webview.

Its possible to try this PR with:

```
npm install --save "react-native-webview@jerolimov/react-native-webview#fix-ios9-crash"
# or
yarn add "react-native-webview@jerolimov/react-native-webview#fix-ios9-crash"
```

If you use CocoaPods, you should also update your Pods with

```
cd ios && pod update && cd ..
```

### DIFF
--- a/ios/RNCWKWebView.m
+++ b/ios/RNCWKWebView.m
@@ -43,8 +43,6 @@ static NSURLCredential* clientAuthenticationCredential;
   BOOL _savedHideKeyboardAccessoryView;
 }
 
-- (void)dealloc{}
-
 - (instancetype)initWithFrame:(CGRect)frame
 {
   if ((self = [super initWithFrame:frame])) {
@@ -71,6 +69,11 @@ static NSURLCredential* clientAuthenticationCredential;
   }
 
   return self;
+}
+
+- (void)dealloc
+{
+  [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
 /**
@@ -164,6 +167,7 @@ static NSURLCredential* clientAuthenticationCredential;
         [_webView.configuration.userContentController removeScriptMessageHandlerForName:MessageHandlerName];
         [_webView removeObserver:self forKeyPath:@"estimatedProgress"];
         [_webView removeFromSuperview];
+        _webView.scrollView.delegate = nil;
         _webView = nil;
     }
 
--- a/ios/RNCWKWebView.m
+++ b/ios/RNCWKWebView.m
@@ -45,24 +45,6 @@ static NSURLCredential* clientAuthenticationCredential;
 
 - (void)dealloc{}
 
-/**
- * See https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/DisplayWebContent/Tasks/WebKitAvail.html.
- */
-+ (BOOL)dynamicallyLoadWebKitIfAvailable
-{
-  static BOOL _webkitAvailable=NO;
-  static dispatch_once_t onceToken;
-
-  dispatch_once(&onceToken, ^{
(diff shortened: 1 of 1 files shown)


jerolimov:
Its possible to try this PR with:

```
npm install --save "react-native-webview@jerolimov/react-native-webview#fix-ios-crash"
# or
yarn add "react-native-webview@jerolimov/react-native-webview#fix-ios-crash"
```

If you use CocoaPods, you should also update your Pods with

```
cd ios && pod update && cd ..
```

Titozzz:
Cc @adrienharnay

Titozzz:
Great work. Will merge tomorrow after testing 💯

ios/RNCWKWebView.m
```diff
@@ -91,6 +71,11 @@ - (instancetype)initWithFrame:(CGRect)frame
   return self;
 }
 
+- (void)dealloc
+{
+  [[NSNotificationCenter defaultCenter] removeObserver:self];
```
Titozzz: What does this change do?

Titozzz: Considering this: https://useyourloaf.com/blog/unregistering-nsnotificationcenter-observers-in-ios-9/

AndrewJack:
Just testing this now, thanks. One thing, there's a typo in your install instructions:
```diff
-yarn add "react-native-webview@jerolimov/react-native-webview#fix-ios-crash"
+yarn add "react-native-webview@jerolimov/react-native-webview#fix-ios9-crash"
```

AndrewJack:
Seems to work. 
Tested before with crash, and after no crash. On an iOS 9.3 Simulator.
👍

Titozzz:
Yeah, I'm just not sure the `[[NSNotificationCenter defaultCenter] removeObserver:self]` is mandatory, but everything else is good. I'll merge when @jerolimov replies to the comment above.

AndrewJack:
Not sure about that line either. Doesn't look like it's needed 🤷‍♂️.

jerolimov:
@AndrewJack Could catch 😄  Thanks for verifying this.

ios/RNCWKWebView.m
jerolimov: Oh maybe this isn't required, but the init method ~10 lines above registers the RNCWKWebView class as observer. So I think its "good style" remove the observer (or any other listener, subscriptions) when the class was deallocated. I could remove the dealloc method if you think this could confuse someone. I would prefer to keep it.

PR #357, opened 2019-02-22
Android: Support file extensions in file input accept attributes

ryanlntn:
The current implementation doesn't support a list of file extensions in the accept attribute (e.g. `<input type="file" accept=".jpg, .png">`) while that is a valid value per the HTML spec. I've updated the implementation to convert any file extensions to mime types before we set the type(s) on the intent to rectify the issue. In addition I've updated the `acceptsImages` and `acceptsVideo` methods to handle file extensions as well.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
@@ -250,7 +250,7 @@ public class RNCWebViewModule extends ReactContextBaseJavaModule implements Acti
       _acceptTypes = DEFAULT_MIME_TYPES;
     }
     if (acceptTypes.matches("\\.\\w+")) {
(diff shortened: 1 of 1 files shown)


Titozzz:
I remember asking for this on the initial PR. Glad someone Finally did it ❤️❤️❤️.

Need testing and good to go

jamonholmgren:
@ryanlntn  Is there any documentation that should be updated to include info on this? A quick look didn't reveal anything, and maybe it's not necessary, but just thought I'd ask.

FlorisWarmenhoven:
Tested and confirmed to add the functionality described. After applying the changes proposed by @ryanlntn I was able to successfully see and select images.

Tested on Windows 10 with Android Emulator and this jsfiddle: https://jsfiddle.net/bfp15m4v/

Before fix: 
Before Image

After fix:
After Image

ryanlntn:
@jamonholmgren I don't think there is any documentation update needed. It now works as you would it expect it to.

PR #382, opened 2019-03-04
Add multiple file upload selection in the docs

andreipfeiffer:


andreipfeiffer:
@Titozzz can you please restart the CI build? I've only changed the docs, which shouldn't make the tests fail, since previously it passed.

jamonholmgren:
Ignore the test failure -- they'll go away when #395 is merged.

PR #393, opened 2019-03-06
Add typescript definition for isFileUploadSupported

ryanlntn:
This static async function was missing a type definition for typescript so I added one 😄 

Before:
screen shot 2019-03-06 at 10 56 40 am

After:
screen shot 2019-03-06 at 10 56 15 am


jamonholmgren:
Looks like we have some flowcheck errors?

jamonholmgren:
The flow errors are unrelated -- I tested and we are having the same issues on `master`.

jamonholmgren:
Flow errors are fixed in #395.

PR #394, opened 2019-03-06
Rewrite Kotlin sources to Java

mizinin:
In order to maintain the consistency of codebase and remove some extra dependencies for Android build, I rewrote Kotlin code to Java.

This already was tested on production.


jamonholmgren:
Interesting. I suppose this is a good idea, given we don't have much written fully in Kotlin anyway. I'd say 👍 if it works as advertised.

Titozzz:
Actually, sorry about that but the plan is to do the opposite, I only converted events to kotlin as a POC but everything will need to change 😊

PR #397, opened 2019-03-07
Add injectedJavaScriptBeforeDocumentLoad

### LINKED ISSUE #229: Injecting javascript prior to document load
Is there any way to inject javascript prior to the document loading? I think this would be related to #227

Thanks!

paulxuca:
Fixes #229 
Demo app here: https://github.com/paulxuca/react-native-webview-demo

cc @jamonholmgren


paulxuca:
@jamonholmgren Could you take a look again? I think it's ready! :)

jamonholmgren:
Thanks for the ping @paulxuca and for the PR! Reviewing...

jamonholmgren:
@paulxuca It looks good -- what is all the code regarding `OkHttpClient`, though? That doesn't look relevant to this PR unless I'm mistaken?

paulxuca:
I think the 2 files in question are:
`RNCWebViewCookieJar.java`: Used here - https://github.com/react-native-community/react-native-webview/pull/397/files#diff-186a7fa0ea29d9ede1ed98c48d63340fR231
As we intercept requests and request the intercepted webpage seperately for js script injection, it is important to preserve response cookies in order to match the same behaviour as a normal webpage load.
`HeaderParser.java`: Used here - https://github.com/react-native-community/react-native-webview/pull/397/files#diff-186a7fa0ea29d9ede1ed98c48d63340fR302

jamonholmgren:
This seems pretty good. You'll want to resolve conflicts (since @Titozzz merged in his TypeScript rewrite).

@Titozzz , what do you think about this? It seems fine to me overall, but I'm just not familiar enough with the cookie handling code here to say for sure.

alexanderby:
image

paulxuca:
@jamonholmgren I think it's ready for another go! I fixed the typescript stuff

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -648,6 +670,128 @@ protected WritableMap createWebViewEvent(WebView webView, String url) {
...
+    private WebResourceResponse interceptRequest(WebView webView, WebResourceRequest webRequest) {
+      Request request = buildRequest(webRequest);
+      String address = request.url().toString();
+
+      if (request == null || address.startsWith("data:")) {
+        return null;
+      }
+
+      try {
+        if (temporaryResponse != null) {
```
Titozzz: temporaryResponse is never given a value, how could it be different from null?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -540,7 +554,6 @@ public void receiveCommand(WebView root, int commandId, @Nullable ReadableArray
         if (args == null) {
           throw new RuntimeException("Arguments for loading an url are null!");
         }
-        root.loadUrl(args.getString(0));
```
Titozzz: How does LOAD_URL command works now?

paulxuca: This is my bad - accidently removed while I was pulling from upstream.

bashen1:
Continuous attention, this is a very good PR

paulxuca:
@Titozzz Can you take a look at the PR again? Thanks!

Titozzz:
I'm very worried about rewriting the whole way network request works on android.. Have you considered other ways to do all that?

paulxuca:
@Titozzz Thanks for chiming in - We've been using this method without fail in a production app for most of this year so far. If you have any other ideas about how to implement the logic, please let me know! I'd love to adapt the PR.

Titozzz:
Well I'm trying to understand the needs behind the feature, can you detail these?

paulxuca:
Sure - I'm sure others from #229 can chime in too, but my need is that I need to inject `window.web3` (https://web3js.readthedocs.io/en/1.0/web3.html) prior to the document loading, as we are hosting applications with javascript that depends on `window.web3` being defined prior to script evaluation. Injecting it after the document is loaded is too too late in the document load lifecycle, so I have to inject it prior to the document loading.

bashen1:
Continuous attention, this is a very good PR

bashen1:
@Titozzz

Titozzz:
Ok, I knew about that use case from @jeluard. I'm trying to think if this is the only reason we would want to do that.

Chengchcc:
In Android:
   remote uri such as ```www.google.com``` -> it works
   local uri such as ```file:///xxxx/index.html``` -> app crash

jinzunyue:
@Titozzz @paulxuca I wonder when can this feature be released?I really need this,but there is some other problems.I know you guys are busy,i also hope you can merge and release this as soon as possible.

petekp:
@Titozzz Another possible use case is hiding DOM elements before the document fully loads. For example, in our app we are providing native navigation controls for the WebView and need to remove the existing web navigation within the WebView DOM. Unfortunately, this can only be done after the document has loaded, which means the web navigation is visible for a brief moment until it's removed with a script delivered via `injectedJavascript`.

This could be solved by adding some code on the website being loaded that using device or media detection to omit elements, but I think there is an argument to be made for enabling direct manipulation without needing to add detection logic to the web-end.

sapjax:
Any update about this? It's an important feature.

sapjax:
Hi @paulxuca 
In your repo [demo-upstream-injectjavascriptbeforedocumentload](https://github.com/paulxuca/react-native-webview/blob/demo-upstream-injectjavascriptbeforedocumentload/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L534) branch, there is some broken code block.

sapjax:
@paulxuca The API exposed in Android (`injectedJavaScriptBeforeDocumentLoad`) is not the same as it in iOS (`injectedJavaScriptBeforeLoad`)

kognise:
Any update on this? I would love to use this.

Fausto95:
Is this going anywhere?

Issue #384, opened 2019-03-04
testID not propagated on iOS?


ziomanlio:
Unlike the Android's _WebView.android.js_ class, the iOS's _WebView.ios.js_ class does not propagate `testID` to `NativeWebView`. So I can't select WebView when running my tests in iOS.
Any solution? 
Thank you

-----------
RN v0.58.3
react-native-webview v4.0.0

Titozzz:
#206 related maybe

jamonholmgren:
We would love a PR that fixes this, @ziomanlio !

CubeSugar:
- about original view props

In my opinion, the way to set original view props by using 

```testID={this.props.testID}```(like pr #402 ) 

or 

```someProp={this.props.someProp}```

is not the best practice.  

@ziomanlio needs `testID` this time, someone may need more other props somewhen.

#206 may be one of the ways to solve problems.  

- workaround

U can set `testID` by using prop [NativeConfig](https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#nativeconfig)

demo:

```jsx
<WebView
    {...otherProps}
    nativeConfig={{
        props: {
            testID: "some string",
        },
  }}
/>
```

ziomanlio:
> * about original view props
> 
> In my opinion, the way to set original view props by using
> 
> `testID={this.props.testID}`(like pr #402 )
> 
> or
> 
> `someProp={this.props.someProp}`
> 
> is not the best practice.
> 
> @ziomanlio needs `testID` this time, someone may need more other props somewhen.
> 
> #206 may be one of the ways to solve problems.
> 
> * workaround
> 
> U can set `testID` by using prop [NativeConfig](https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#nativeconfig)
> 
> demo:
> 
> ```js-jsx
> <WebView
>     {...otherProps}
>     nativeConfig={{
>         props: {
>             testID: "some string",
>         },
>   }}
> />
> ```

Thank you @CubeSugar ! I will temporarily fix my issue using this workaround.

Titozzz:
This will be fixed very soon actually, doing lasts checks before release

PR #206, opened 2018-12-13
fix: support original view props, such as onLayout

CubeSugar:
`WebView` should have the `View`'s generic props, such as `onLayout`, according to the [doc](https://facebook.github.io/react-native/docs/webview#props).

PR #402, opened 2019-03-07
Let iOS's WebView component propagate testID to NativeWebView component

ziomanlio:
#384 related.
It's my first pull request... advice welcome :)


harrisrobin:
I believe this might be solved when #206 is resolved and merged so perhaps it's better to wait for this one: 
https://github.com/react-native-community/react-native-webview/pull/206

Titozzz:
This will effectively be solved when #206 is merged. Thank you for you PR tho, and keep contributing ❤️

PR #403, opened 2019-03-07
Failed to load local html files in real devices

jimerino:

PR #406, opened 2019-03-08
Fixes unable to load from local file on android on real device

Arjan-Zuidema:
Reload once on a net::ERR_ACCESS_DENIED error when loading from a local file.

This fixes a net::ERR_ACCESS_DENIED error when trying to load a local file on a real device. Initial load triggers this error, but when a reload is triggered the file loads properly


Titozzz:
Even if this works, are we sure we cannot find a way to fix the first crash instead of using a workaround 🤔 ?

Arjan-Zuidema:
@Titozzz I am not that familiar with Android and their WebView. This can be a temporary workaround if someone else finds a better solution, or maybe even the cause of the problem.

For now I am unable to use any version higher than v4.1.0

jamonholmgren:
@Arjan-Zuidema Can you test with #403 to see if that works for you? If so, maybe that will be a better solution than this.

Arjan-Zuidema:
@jamonholmgren That pull request only fixes iOS. My problem occurs on Android. I don't think that #403 will fix it

jamonholmgren:
@Arjan-Zuidema Can you fix the conflicts?

Arjan-Zuidema:
Updating to the latest version (5.4.1) seems to load the local file now on Android. But I still get the yellow box warning with the "net::ERR_ACCESS_DENIED" warning. I was unable to resolve the conflicts as too much has changed.

So I will close this for now.

PR #414, opened 2019-03-13
Android scaling issue

sadvit:
When setting a custom font size in the Android system, an undesirable scale of the site interface in WebView occurs.

I researched that when setting the standard textZoom (100) parameter size, this undesirable effect disappears.

This can be very useful if you need to avoid the scale of content in WebView when changing the size of system fonts, or change textZoom property directly.

Example:

`
<WebView
  textZoom={100}
/>
`


Titozzz:
Can you fix the merge conflicts or do you need me to do it? I'd like to get this merged

sadvit:
@Titozzz 
Conflicts resolved, could you review please?

Titozzz:
I wonder if setting a default value is a good idea. Cause this means it would be "breaking". I think I'd prefer this to be opt-in 🤔?

What do you think?

src/WebView.android.tsx
```diff
@@ -283,6 +284,7 @@ class WebView extends React.Component<AndroidWebViewProps, State> {
         onLoadingProgress={this.onLoadingProgress}
         onLoadingStart={this.onLoadingStart}
         onMessage={this.onMessage}
+        textZoom={textZoom}
```
Titozzz:
```suggestion
```

src/WebView.android.tsx
```diff
@@ -226,6 +226,7 @@ class WebView extends React.Component<AndroidWebViewProps, State> {
       source,
       style,
       nativeConfig = {},
+      textZoom,
```
Titozzz:
```suggestion
```

sadvit:
@Titozzz 
Yes, you are right, it can break - I deleted default value

src/WebView.android.tsx
Titozzz: Same

PR #418, opened 2019-03-14
Add flow libdef generation script

alpha0010:
Automated creation of flow library definitions from typescript types file.  Results are imperfect, as the only way to get types from `react-native` is to copy them over, which seems unwise.

Usage: `bin/gen-flow.js typings/index.d.ts > flow/react-native-webview.js`

Since this has been extracted from core, flow types no longer work for apps using `WebView`.  The consuming app can add this projects `flow` directory to the app's `.flowconfig`.

Opinions?  (I have not yet thoroughly tested.)


Titozzz:
I'm fairly surprised that you say flow types don't work cause they are inside the code 🤔 .

However, I've been doing a full typescript rewrite since yesterday so this is going to come handy I guess. Do you think you would be able to add this part to my PR if everything else is good?

alpha0010:
There are two problems with using flow types from this library.  `index.js` is not marked as `// @flow`.  More significantly, it appears `flow` cannot resolve imports between modules in `node_modules`.  So, statements like `import type { SyntheticEvent } from 'CoreEventTypes';` fail (flow's internal library does allow modules to `import type { XXX } from 'react';` and a few other core types).

I will take a look at your rewrite.

Titozzz:
Oh okay, well this means right now no one uses this lib with flow, so I could potentially release TS first and then add flow typing.

Titozzz:
Do you confirm @alpha0010?

alpha0010:
Correct.

Titozzz:
@alpha0010 TS pr landed if you want to get this done

alpha0010:
Thanks, will work on this soon.

Titozzz:
Great news 🚀

alpha0010:
@Titozzz Ready for review.

PR #423, opened 2019-03-15
OnOriginWhitelistFailed overrides default behaviour Open Link in Safari

manuelblum:
default behaviour of opening not whitelisted links in safari can be overwritten with this prop. 

Usefull if you want to keep the user in the app and present an for example an "In App Browser" or asking him if he really wants to open url in browser, instead of sending him to standard browser


Titozzz:
Interesting feature!

I'm doing a complete TS modification on the repository right now but i'll make sure to merge this asap

Titozzz:
After thinking about this, I wrote https://github.com/react-native-community/react-native-webview/issues/438. Please let me know what you think.

PR #425, opened 2019-03-15
feat(typescript): Source code rewrite using typescript

Titozzz:


Titozzz:
@linusU Can you have a look, I've tried to take into account everything that was said in the previous PR #171

src/WebViewShared.ts
```diff
@@ -42,16 +40,16 @@ const createOnShouldStartLoadWithRequest = (
     url: string,
     lockIdentifier: number,
   ) => void,
-  originWhitelist: ?$ReadOnlyArray<string>,
```
LinusU: TypeScript also has a read-only array: `ReadonlyArray<T>` ☺️

Titozzz: Well we do have eslint to not modify received parameters, but that's nice to know

LinusU: It's nice to use because if users are using readonly variables to enforce mutability, TypeScript won't let them pass those variables to us since it thinks we are going to mutate them.

Basically, it's a way to document the interface 📝

Titozzz: ok nice I'll make those changes before merging :)

LinusU:
Awesome work @Titozzz 👏 🚀

PR #434, opened 2019-03-19
fix(android): Do not hold a strong reference to native module in package

sjchmiela:
View managers can access native modules belonging to the current `ReactContext` via `.getNativeModule(Class nativeModuleClass)` method, relying on _last created instance of native module_ may be a too optimistic approach to always work. 🙂


Titozzz:
Niiice 👍

PR #435, opened 2019-03-19
add raw compability with react-native-windows

mroswald:
Libraries like [react-native-canvas](https://github.com/iddan/react-native-canvas) rely on `react-native-webview` to export the `WebView`. Unfortunalty for `react-native-windows` this won't work as an internal import can't be resolved.

This PR resolves the WebView to the original WebView provided with `react-native-windows` in order to give _some_ compability which seem better to me than _no_ compability :)

I saw this comment in the contribution guide: `If you want to add another React Native platform to this repository, you will need to create another .flowconfig for it.` Although I value testing things, I don't think it makes much sense to cover a pure export/import.


sjchmiela:
🤔 Removing `.windows` part would also fix basic compatibility (compile-time) with `react-native-web` 🙃

Titozzz:
@sjchmiela that's smart. Might do it 😁

mroswald:
Yes, on the other hand I wanted to avoid problems with platforms that are not yet covered and do not have `WebView` exported. This in mind, would you also suggest to remove the file suffix?

Titozzz:
Well importing something that does not exist should not crash, it would just be null right?

mroswald:
You're right. Will adjust this PR.

Titozzz:
Because of the TS rewrite, I directly included this in my PR so you don't fight against merge conflicts

Issue #438, opened 2019-03-20
[WIP] Clean code // Remove unnecessary props -- PART 1


Titozzz:
Over time, WebView has evolved a lot and has gone through many updates, resulting today in a hard and messy code to maintain.

In order to keep the webview API as clean as possible, we are going to remove some props and do some code refactoring. But keep in mind that this will not affect what you can achieve with WebView.

### Affected props:
* **originWhitelist**: this prop is used to limit the URLs that webview can actually load. 

*Motivations*
1) This prop can be completely emulated with onShouldStartLoadWithRequest().

Current code is 
```js
({ nativeEvent }: WebViewNavigationEvent) => {
    let shouldStart = true;
    const { url, lockIdentifier } = nativeEvent;

    if (!passesWhitelist(compileWhitelist(originWhitelist), url)) {
      Linking.openURL(url);
      shouldStart = false;
    }

    if (onShouldStartLoadWithRequest) {
      shouldStart = onShouldStartLoadWithRequest(nativeEvent);
    }

    loadRequest(shouldStart, url, lockIdentifier);
  };
```
The future code will be something like
```js
({ nativeEvent }: WebViewNavigationEvent) => {
	const { url, lockIdentifier } = nativeEvent;

    if (
      onShouldStartLoadWithRequest
      && !onShouldStartLoadWithRequest(nativeEvent)
    ) {
      loadRequest(false, url, lockIdentifier);
    }
    loadRequest(true, url, lockIdentifier);
  };
```

This means that if you want to limit URLs, you'll need to do it directly inside your onShouldStartLoadWithRequest prop.

2) Removing it will even improve the current API by giving the user control over what should happen when the whitelist does not match, (So instead of doing #423 and adding a prop we can actually remove one) right now the webview forces Linking.openURL()

jamonholmgren:
I agree with this removal. If people want to emulate it, they could provide a subclass of the WebView that implements `onShouldStartLoadWithRequest` and emulate the `originWhitelist` behavior.

👍

Salakar:
Discussed this yesterday with you but for public visibility, I also agree with this removal. There's no point in having 2 features where one can also provide what the other does anyway.

jamonholmgren:
Could we provide a warning if that prop is used? And say something like "`originWhitelist` has been removed; use `onShouldStartLoadWithRequest` instead".

Titozzz:
this is planned actually 👍

PR #439, opened 2019-03-20
Remove originWhitelist (see #438)

Titozzz:


jamonholmgren:
We'll want to remove it from the reference, too:

https://github.com/react-native-community/react-native-webview/blob/453b7dd3a0b821de6dd0578c5df7d4004d74f46c/docs/Reference.md#L19

Salakar:
> We'll want to remove it from the reference, too:

Done

PR #442, opened 2019-03-21
Add missing type definitions of the methods

svbutko:
Also made the exported component a class (to add methods) and made him extend Component instead of ComponentType.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,8 +1,35 @@
-import { ComponentType } from 'react';
+import { Component } from 'react';
 // eslint-disable-next-line
 import { IOSWebViewProps, AndroidWebViewProps } from './lib/WebViewTypes';
 
-declare const WebView: ComponentType<IOSWebViewProps & AndroidWebViewProps>;
+export class WebView extends Component<IOSWebViewProps & AndroidWebViewProps> {
+    /**
+     * Go back one page in the webview's history.
+     */
+    goBack: () => void;
 
-export { WebView };
-export default WebView;
+    /**
+     * Go forward one page in the webview's history.
+     */
+    goForward: () => void;
+
+    /**
+     * Reloads the current page.
+     */
+    reload: () => void;
+
+    /**
+     * Stop loading the current page.
+     */
+    stopLoading(): void;
+
+    /**
+     * Extra Native Component Config.
+     */
+    extraNativeComponentConfig: () => any;
+
+    /**
+     * Executes the JavaScript string.
+     */
+    injectJavaScript: (script: string) => void;
+};


Titozzz:
You removed the named export 😢

svbutko:
> You removed the named export 😢

If you mean

```typescript
export {WebView};
export default WebView;
```

I can surely put them back but does it makes any sense?

jvandenaardweg:
Thanks @svbutko !

This should also fix the problem of `ref` not available, because you are now extending the React `Component` 👍

Screenshot 2019-03-21 15 11 45

svbutko:
> Thanks @svbutko !
> This should also fix the problem of ref not available, because you are now extending the React Component 👍

That was the actual reason behind this PR, only after getting inside I noticed that it also lacks methods definitions 😄

Titozzz:
We need export {WebView}; before the TS rewrite everyone was using:
import {WebView} from 'react-native-webview';
now you can do both named or default.

jvandenaardweg:
> I can surely put them back but does it makes any sense?

It is more in line with other React Native Community packages, they also have an export default 👍

Examples:
https://github.com/react-native-community/react-native-netinfo/blob/master/typings/index.d.ts
https://github.com/sunnylqm/react-native-storage/blob/master/src/storage.d.ts

Titozzz:
Yeah I really wanted to have the default but I tried not to be breaking 😄

Titozzz:
Great thanks

Titozzz:
@allcontributors[bot] please add @svbutko for doc

jvandenaardweg:
> I think this PR also fixes the problem i'm having of `ref` not being available

Awesome, it worked. Thanks for the quick merge!

Titozzz:
Keep the nice PRs coming guys 💯

vovkasm:
@jvandenaardweg How does it work for you?
I just try latest version of `react-native-webview` (5.4.2) and got this:
```
node_modules/react-native-webview/index.d.ts:5:1 - error TS1046: A 'declare' modifier is required for a top level declaration in a .d.ts file.

5 class WebView extends Component<IOSWebViewProps & AndroidWebViewProps> {
  ~~~~~

node_modules/react-native-webview/index.d.ts:35:2 - error TS1036: Statements are not allowed in ambient contexts.

35 };
    ~
```

vovkasm:
Also, package.json contains incorrect typings field: https://github.com/react-native-community/react-native-webview/blob/master/package.json#L5

jvandenaardweg:
@vovkasm probably because i have `"skipLibCheck": false` in my `tsconfig.json`. I'm getting too much errors related to external modules when I set that one to `true`.

vovkasm:
Hmm... Couple of months ago I remove skipLibCheck and fixed all errors... )))
I can provide additional info what was fixed if needed.

Do you plan to fix typings or I should create PR? ;-)

jvandenaardweg:
> Hmm... Couple of months ago I remove skipLibCheck and fixed all errors... )))

Yeah, `"skipLibCheck": false` isn't really a permanent fix. 

I think you should create a PR for this! 👍

vovkasm:
Will do today.

vovkasm:
done: #453

PR #455, opened 2019-03-26
Expose the WebView.postMessage function

sbeca:
The `postMessage` function is exposed in `src/WebView.android.tsx` and `src/WebView.ios.tsx`, but not in `index.d.ts` or the documentation so this PR adds it to those locations.

Also, while testing I noticed that Android and iOS where dispatching the message event to different elements, Android using `document.dispatchEvent` and iOS using `window.dispatchEvent`. This meant that listening for the event wasn't reliable between platforms so I changed iOS to use `document` as well. iOS was changed in commit f3bdab5a228aaf8ee5bc36d8f66442658b778d6a which looks like it might have been changed by accident while making the Web -> React Native `postMessage` change.

### DIFF
--- a/docs/Guide.md
+++ b/docs/Guide.md
@@ -214,11 +214,12 @@ Add permission in AndroidManifest.xml:
 
 You will often find yourself wanting to send messages to the web pages loaded by your webviews and also receiving messages back from those web pages.
 
-To accomplish this, React Native WebView exposes three different options:
+To accomplish this, React Native WebView exposes four different options:
 
 1. React Native -> Web: The `injectedJavaScript` prop
 2. React Native -> Web: The `injectJavaScript` method
-3. Web -> React Native: The `postMessage` method and `onMessage` prop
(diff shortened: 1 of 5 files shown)


Titozzz:
@sbeca I'm sorry, as the code contained inside your PR is great, and documentation was very clean and nice to read but as already said here https://github.com/react-native-community/react-native-webview/issues/411 and inside the readme, postMessage function will be removed. As you can do everything with injectJavascript.

If you have any more question, I'll be happy to answer it!

sbeca:
No problem. From my reading I thought that you were only removing the Web -> React Native `postMessage` and didn't realise that you were also planning on removing the React Native -> Web `postMessage`. I'll just replicate the `postMessage` functionality within my library using `injectedJavascript`.

Titozzz:
The react native -> web is the only one going away. The other one has been moved into it's own scope in the window object and should not move anymore 🙂 . I use injectJavascript and not injectedJavascript to communicate once the Webview is started

PR #457, opened 2019-03-26
fix(types): export WebViewProps for external use

radiosilence:


radiosilence:
@Titozzz thanks you :)

PR #459, opened 2019-03-27
fix(android): Annotate RNCWebViewModule with @ReactModule to comply with latest changes

sjchmiela:
The app was crashing with `"Could not find @ReactModule annotation in class RNCWebViewModule"` exception. Searching for this message in RN's code I found [this commit](https://github.com/facebook/react-native/commit/0cd3994f1a4ec1172af840cf5de3dddb1ae9e9c4), introduced in React Native 0.58, which requires native modules to be annotated with @ReactModule annotation. 🤔 

After adding the annotation to the module, tapping on a file input field no longer crashes the app (in fact it shows the file browser). 🎉 

I haven't had caught it previously, testing only against React Native 0.57, sorry! 😞 

For reference see similar fix in `react-native-gesture-handler` — https://github.com/kmagiera/react-native-gesture-handler/pull/295.

Fixes https://github.com/react-native-community/react-native-webview/issues/458.


Titozzz:
❤️

PR #478, opened 2019-04-04
Removes hard-coded container white backgroundColor

### LINKED ISSUE #475: [iOS] Unable to set transparent background
After upgrading to `5.6.1` from `5.4.1` I'm no longer able to set a transparent background on a `WebView`.

I think it broke in 319a86e2361ac0161cbe2897a907d43b085029fc

The background is always white.

maxschmeling:
Fixes #475

#472 added a `backgroundColor` of `white` to the `container` style which broke the ability to set a `transparent` background color on the `WebView` component. This simply removes that unnecessary style.

### DIFF
--- a/src/WebView.styles.ts
+++ b/src/WebView.styles.ts
@@ -15,7 +15,6 @@ const styles = StyleSheet.create<Styles>({
   container: {
     flex: 1,
     overflow: 'hidden',
-    backgroundColor: 'white',
   },
   errorContainer: {
     flex: 1,


Titozzz:
Well, can't you just override it by providing style={{ backgroundColor: 'transparent' }}? 
The idea behind the change was to provide better defaults.
What do you think?

svbutko:
> Well, can't you just override it by providing style={{ backgroundColor: 'transparent' }}?
> The idea behind the change was to provide better defaults.
> What do you think?

Setting the background in the style does nothing, it still stays white

Titozzz:
Damn you are right, this is my mistake. Will merge that, while I think about this.

PR #479, opened 2019-04-04
Remove duplicate example of onShouldStartLoadWithRequest

MalcolmScruggs:
There is a duplicate example for onShouldStartLoadWithRequest. This removes the duplicate example.


Titozzz:
thanks

PR #496, opened 2019-04-11
Adds version badge to README

jamonholmgren:
Per @Rinkuk1993's [comment](https://github.com/react-native-community/react-native-webview/issues/437#issuecomment-480158586), the changelog might confuse some people as to what version is latest, so I've added a "latest version" badge to the README in two places. I've also reversed the order of the changelog so the most recent is at the top.


Titozzz:
chore(docs): added version badge 💯

PR #506, opened 2019-04-16
Add applicationNameForUserAgent support for iOS WKWebView

pmusaraj:
Allows appending to the user agent. It's an alternative to setting `userAgent`. iOS WKWebView only.


Titozzz:
Damn this was also missing in the typescript declaration can you submit a new PR to fix that @pmusaraj 🙂?

PR #509, opened 2019-04-17
In Devices moto e5 play and Samsung Galaxy Note 9 Video Capture via camera return data as null

mahavirvataliya:
In Devices moto e5 play and Samsung Galaxy Note 9 Video Capture via camera return data as null.


Titozzz:
@andreipfeiffer what do you think?

andreipfeiffer:
@Titozzz that might work, I know I had a reason why I did it the way I did it, because there's a comment there ;)

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
```diff
@@ -244,8 +244,8 @@ private Intent getVideoIntent() {
     Intent intent = new Intent(MediaStore.ACTION_VIDEO_CAPTURE);
     // @todo from experience, for Videos we get the data onActivityResult
```
andreipfeiffer: I think this comment is not needed anymore

PR #516, opened 2019-04-19
Add `onScroll` callback for iOS & Android

jaredly:
This code was mostly extracted from https://github.com/react-native-community/react-native-webview/pull/202

I tried and tried to make it work with `Animated.event`'s `useNativeDriver`, but I was unsuccessful 😢 that'll have to be done later once I understand better how Animated's native stuff is hooked up.


Titozzz:
I'm a bit worried that we will overload the bridge with such behavior. This should only trigger event if onScroll is defined right? We might also want to implement https://facebook.github.io/react-native/docs/scrollview#scrolleventthrottle

jaredly:
Yeah it should only happen if onScroll is passed in. I agree supporting `scrollEventThrottle` would be nice too

jamonholmgren:
`scrollEventThrottle` can be a separate PR if there's a noticeable performance bottleneck. I'd be fine with merging this. 👍

guptayash:
When is this going to merge?

ouabing:
Failed to build with RN 0.59.8, seems like `ScrollEventType.SCROLL.getJSEventName()` should be `ScrollEventType.getJSEventName(ScrollEventType.SCROLL)` ?

Titozzz:
Fixed in 6.0.1 @ouabing

Gregoirevda:
@Titozzz  @jaredly this is 5 years ago, but has there been progress on how to update the implementation to support `useNativeDriver: true`?

Gregoirevda:
Got iOS working with
```jsx
<AnimatedWebView
 onScroll={Animated.event([{ nativeEvent: { contentOffset: { y: animatedScrollYRef.current } } }], {
   useNativeDriver: true, // only false worked
})} />
  ```
  
  In `RNCWebViewManager.m` add 
 ```obj-c
RCT_REMAP_VIEW_PROPERTY(contentOffset, scrollView.contentOffset, CGPoint)
```
I struggled with `Animated.createAnimatedComponent(WebView)`, because the exported component is a fn component, but a simple class wrapper seem to loose the ref... (`getNode` isn't working). 

I updated `WebView.ios.tsx` 
```js
const RNCWebView = Animated.createAnimatedComponent(_RNCWebView);
```

`useNativeDriver: true` then started working

PR #520, opened 2019-04-22
[android] Migrate to Android X

Salakar:
React Native 0.60 will use AndroidX by default; because of this the dependencies are no longer available and causes webview to fail to build:

image

To fix this I've replaced all usages of the support libs to now use AndroidX and have enabled AndroidX usage in the `gradle.properties` file.

## Notes

 - The `android.support.FILE_PROVIDER_PATHS` usage in the `AndroidManfiest.xml` remains the same on AndroidX - no changes needed
 - AndroidX requires a minimum `compileSdkVersion` version of 28 - which this library is already set to - however, if anyone overrides this (`getExtOrDefault`) to a lower version this will cause build issues. Should this be a major version?
 - The alternative to not using AndroidX would be to explicitly add the v4 support libs to the library dependencies, but I took the migrating to AndroidX approach.

## Test Plan

App with these changes builds and displays a WebView on Android:

image

image


Salakar:
> verify it doesn't build with RN 0.60 on master

@jamonholmgren First build screenshot on description was with me building from RN.60 and latest webview

Second build screenshot was building on RN.60 with this change applied.

----

This failure can be viewed quickly on the RN.60 version of the init template using the following:

```bash
npx @react-native-community/cli@next init --template=@invertase/template <myProjectName>
cd <myProjectName>
yarn add react-native-webview
react-native run-android
```

(uses auto-linking - so no changes required to link webview)

PR #521, opened 2019-04-23
fix: iOS crash when WebView's title is nil

TMomemt:
This crash is found in our fabric.

I can't reproduce it.

But this protection is needed because in Apple's doc the title property is nullable.


Titozzz:
@allcontributors[bot] please add @TMomemt for code

PR #524, opened 2019-04-23
Add incognito prop to Android RNCWebViewManager

oNaiPs:
Allows the webview to be opened with an ephemeral data storage.

Follow-up to https://github.com/react-native-community/react-native-webview/pull/148.

Closes https://github.com/react-native-community/react-native-webview/issues/737

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -328,6 +328,23 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   public void setMessagingEnabled(WebView view, boolean enabled) {
     ((RNCWebView) view).setMessagingEnabled(enabled);
   }
+   
+  @ReactProp(name = "incognito")
+  public void setIncognito(WebView view, boolean enabled) {
+    // Remove all previous cookies
+    CookieManager.getInstance().removeAllCookies(null);
+
+    // Disable caching
+    view.getSettings().setCacheMode(WebSettings.LOAD_NO_CACHE);
+    view.getSettings().setAppCacheEnabled(!enabled);
+    view.clearHistory();
+    view.clearCache(enabled);
+
+    // No form data or autofill enabled
+    view.clearFormData();
+    view.getSettings().setSavePassword(!enabled);
+    view.getSettings().setSaveFormData(!enabled);
+  }
 
   @ReactProp(name = "source")
   public void setSource(WebView view, @Nullable ReadableMap source) {
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -831,9 +831,9 @@ If true, this will be able horizontal swipe gestures when using the WKWebView. T
 
 Does not store any data within the lifetime of the WebView.
 
-| Type    | Required | Platform      |
-| ------- | -------- | ------------- |
-| boolean | No       | iOS WKWebView |
+| Type    | Required | Platform               |
+| ------- | -------- | ---------------------- |
+| boolean | No       | Android, iOS WKWebView |
 
 ---
 
--- a/src/WebViewTypes.ts
+++ b/src/WebViewTypes.ts
@@ -212,6 +212,7 @@ export type OnShouldStartLoadWithRequest = (
 
 export interface CommonNativeWebViewProps extends ViewProps {
   cacheEnabled?: boolean;
+  incognito?: boolean;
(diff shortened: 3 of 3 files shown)


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -367,6 +367,23 @@ public void setInjectedJavaScript(WebView view, @Nullable String injectedJavaScr
   public void setMessagingEnabled(WebView view, boolean enabled) {
     ((RNCWebView) view).setMessagingEnabled(enabled);
   }
+   
+  @ReactProp(name = "incognito")
+  public void setIncognito(WebView view, boolean enabled) {
+    // Disable cookies
+    CookieManager.getInstance().setAcceptCookie(!enabled);
```
Collin3: Instead of not accepting cookies would it make more sense to use `removeAllCookies`? That way when you launch the WebView you get a completely new session as expected, but while the WebView is open, the website can still set cookies as needed. 

The reason I ask is because I pulled these changes that you have into my app, but the site that is used in my WebView needs to use cookies, so it won't work the way it is currently set up. However, I changed this line to be `CookieManager.getInstance().removeAllCookies(null);` and things started to work as expected.

Collin3: Instead of not accepting cookies would it make more sense to use removeAllCookies? That way when you launch the WebView you get a completely new session as expected, but while the WebView is open, the website can still set cookies as needed.

The reason I ask is because I pulled these changes that you have into my app, but the site that is used in my WebView needs to use cookies, so it won't work the way it is currently set up. However, I changed this line to be CookieManager.getInstance().removeAllCookies(null); and things started to work as expected.

oNaiPs: @Collin3 apologies for the late reply. Thanks for your input!

I did some experiments and indeed you're right, the `setAcceptCookie` disables cookie support which makes some web pages not work correctly.

Changed the code to your suggestion. 👍

Collin3: 👍 thanks for getting around to it! I'll pull these changes into my app and see how they work.

Titozzz: let me know how it goes @Collin3

Collin3: @Titozzz I just tested it out and everything is working as expected for me!

bradylcarey:
@titozzz I just pulled it down as well and it looks good to me! 👍:shipit:

Collin3:
@oNaiPs would you be able to update your PR so it is no longer out of date with master? Then maybe @Titozzz or @jamonholmgren could give it a review? I've been using a forked version of `react-native-webview` with these changes in there so we could get Android incognito and it has been working as expected for the past month.

oNaiPs:
@Collin3 @Titozzz @jamonholmgren I rebased the branch to the latest `master`

AlexK777:
Now that there are no conflicts could this please get a review @Titozzz @jamonholmgren? I am also waiting for this PR to land in a release.

DominicSherman:
@Titozzz @jamonholmgren Can this get approved?

MakhouT:
Any idea when this will be merged in? Kind of need it :)

Collin3:
@oNaiPs it looks like the PR is out of date with master again.. Would you mind updating it when you have a chance? Then maybe it can get merged 🤞

docs/Reference.md
```diff
@@ -831,9 +831,9 @@ If true, this will be able horizontal swipe gestures when using the WKWebView. T
 
 Does not store any data within the lifetime of the WebView.
 
-| Type    | Required | Platform      |
-| ------- | -------- | ------------- |
-| boolean | No       | iOS WKWebView |
```
chrisheninger: Maybe keep the Platform section and update it to `iOS WKWebView, Android` so users know that this boolean only works with a WKWebView and not a normal one on iOS.

MakhouT:
Nice, we've got an approval! Thanks for the hard work @oNaiPs 
Time to bring the branch back up to date and merge this baby in!

oNaiPs:
@Titozzz @jamonholmgren can you re-review? I added the suggestion by https://github.com/react-native-community/react-native-webview/pull/524#pullrequestreview-269739954

PR #551, opened 2019-05-02
Fix: OnNavigationStateChanged not always called

manuelblum:
The `OnNavigationStateChanged` Prop of react-native-community-webview is not called when the history state is manipulated via javascript (e.g.: `window.history.pushState({test: "test"}, "Test Title", "index.html")`). 

With this pull request each history change will be noticed by the native webviews (android and ios) and the navigation change will be forwarded to `onNavigationStateChanged` with `navigationType: "other"`.


jamonholmgren:
So I've tried to get this to work, but it doesn't appear to.

```jsx
import React, { Component } from "react";
import { WebView } from "react-native-webview";

export default class App extends Component<Props> {
  render() {
    const html = `
      <html>
      <head>
        <script>
          function changeurl() {
            document.body.innerHTML += '<p>clicked</p>';
            history.pushState({test: "test"}, "Test Title", "#hello");
            document.body.innerHTML += '<p>done clicking</p>';
          }
        </script>
        <style>body, button { font-size: 30px; padding: 30px; }</style>
      </head>
      <body>
        <h1>Testing</h1>
        <h1>Testing</h1>
        <h1>Testing</h1>
        <p>
          <button onclick="javascript:changeurl()">Hello</button>
        </p>
      </body>
      </html>
    `;
    return (
      <WebView
        source={{ html }}
        onNavigationStateChange={nav => alert(JSON.stringify(nav))}
      />
    );
  }
}
```

When I click the button, I see the appendChild fire, but the `pushHistory` doesn't fire `onNavigationStateChange` at all.

image

jamonholmgren:
Interesting -- it _does_ work on Android.

PR #552, opened 2019-05-02
Fix javascript injection on Android <= 4.3

### LINKED ISSUE #108: WebView on Android 4.3 fails to execute any injected Javascript
### Environment

React Native Environment Info:
System:
OS: macOS High Sierra 10.13.6
CPU: x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
Memory: 26.41 MB / 16.00 GB
Shell: 3.2.57 - /bin/bash
Binaries:
Node: 10.7.0 - /usr/local/bin/node
Yarn: 1.7.0 - /usr/local/bin/yarn
npm: 6.2.0 - /usr/local/bin/npm
Watchman: 4.9.0 - /usr/local/bin/watchman
SDKs:
iOS SDK:
Platforms: iOS 12.0, macOS 10.14, tvOS 12.0, watchOS 5.0
Android SDK:
Build Tools: 23.0.1, 26.0.1, 27.0.3
API Levels: 23, 26, 27
IDEs:
Android Studio: 3.1 AI-173.4819257
Xcode: 10.0/10A255 - /usr/bin/xcodebuild
npmPackages:
@lingui/react: 2.7.0 => 2.7.0
react: 16.5.0 => 16.5.0
react-native: 0.57.0 => 0.57.0
npmGlobalPackages:
react-native-cli: 2.0.1

### Description

Recent [changes](https://github.com/facebook/react-native/commit/e6b305b722a8f52fad8ac8bf9f95204fa2192944) that were carried over into this new repo broke execution of any injected Javascript (either via injectedJavaScript prop or manually via injectJavaScript method of the WebView on devices running on Android prior to 4.4.

Looking through the linked issues of the above mentioned commit, say [this one](https://github.com/facebook/react-native/issues/19611), javascript scheme RFC and Android implementation prior to 4.4 (see [issue](https://issuetracker.google.com/issues/36995865)) we can see that the following action is the culprit:

`loadUrl("javascript:" + URLEncoder.encode(script, "UTF-8"));`

This line will encode passed Javascript code for Android <= 4.3 and due to the latter's implementation of loadUrl that doesn't decode the scheme's text we end up with a problem.

In our application we need to grab loaded page's title and set it into our header in RN. Prior to 0.57.0 and the commit it worked without any issues.

As any injected Javascript is wrapped into self-executing anonymous function here:

`evaluateJavascriptWithFallback("(function() {\n" + injectedJS + ";\n})();");`

This is what goes to loadUrl:

`%28function%28%29+%7B%0Awindow.postMessage%28document.title%29%3B%0A%7D%29%28%29%3B
`
### Reproducible Demo

Setup your WebView with any correct value for injectedJavascript prop that would provide a feedback (console.log, data passed via onPostMessage, visible DOM manipulations, etc) on <=4.3 Android device / emulator.

So I suggest we remove the encoding bit from that line as it clearly messes up any passed Javascript.

Taym95:
Fix #108 [react-native-open-source#37](https://github.com/callstack/react-native-open-source-board/issues/37)

I tested it on android emulator 4.3 API 19


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -749,13 +749,7 @@ protected void evaluateJavascriptWithFallback(String script) {
         evaluateJavascript(script, null);
         return;
       }
-
-      try {
-        loadUrl("javascript:" + URLEncoder.encode(script, "UTF-8"));
-      } catch (UnsupportedEncodingException e) {
```
michalchudziak: Is it safe to remove this `try-catch`?

Taym95: Thanks for the review, I tried this:
```
try {
	loadUrl("javascript:(function() {\n" + script + ";\n})();");
} catch (UnsupportedEncodingException e) {
	// UTF-8 should always be supported
	throw new RuntimeException(e);
}
```
But it was always throwing an exception!

`try-catch` was added in these [changes ](https://github.com/facebook/react-native/commit/e6b305b722a8f52fad8ac8bf9f95204fa2192944), I don't think we need it anymore??

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -749,13 +749,7 @@ protected void evaluateJavascriptWithFallback(String script) {
...
-        // UTF-8 should always be supported
-        throw new RuntimeException(e);
-      }
+      loadUrl("javascript:(function() {\n" + script + ";\n})();");
```
jamonholmgren: Couldn't they simply provide this:

```
script = `; return true})() && (callSomethingElse()) && function() {`
```

... to bypass this protection?

Palid: Yep, it's pretty trivial.

fchenchen:
Is this PR merged to the latest release? The bug https://github.com/callstack/react-native-open-source-board/issues/37 seems to still exist.

Issue #24, opened 2018-09-10
MIGRATED: iOS WebView doesn't support hash url change. The onNavigationStateChange listener will not be called when url hash changes. (#20447)


jamonholmgren:
Migrated from https://github.com/facebook/react-native/issues/20447.

@ll929 says:

---
Environment
React Native Environment Info:
System:
OS: macOS High Sierra 10.13.3
CPU: x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
Memory: 109.62 MB / 8.00 GB
Shell: 5.3 - /bin/zsh
Binaries:
Node: 9.11.1 - ~/.nvm/versions/node/v9.11.1/bin/node
Yarn: 1.7.0 - /usr/local/bin/yarn
npm: 5.6.0 - ~/.nvm/versions/node/v9.11.1/bin/npm
Watchman: 4.9.0 - /usr/local/bin/watchman
SDKs:
iOS SDK:
Platforms: iOS 11.2, macOS 10.13, tvOS 11.2, watchOS 4.2
Android SDK:
Build Tools: 23.0.1, 23.0.2, 23.0.3, 24.0.1, 25.0.1, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 26.0.3, 27.0.0, 27.0.3
API Levels: 23, 24, 25, 26, 27
IDEs:
Android Studio: 3.1 AI-173.4819257
Xcode: 9.2/9C40b - /usr/bin/xcodebuild
npmPackages:
react: 16.4.1 => 16.4.1
react-native: 0.56.0 => 0.56.0

Description
Describe your issue in detail. Include screenshots if needed. If this is a regression, let us know.

Reproducible Demo
Let us know how to reproduce the issue. Include a code sample, share a project, or share an app that reproduces the issue using https://snack.expo.io/. Please follow the guidelines for providing a MCVE: https://stackoverflow.com/help/mcve

Expected Behavior
When url hash change, the webivew onNavigationStateChange listener will be invoked, and the canGoBack properties is true

Actual Behavior
When url hash changes, the webivew onNavigationStateChange hasn't been invoked
---

KoenLav:
Should it change?

It should definitely be consistent between Android and iOS, but should it change when the URL hash changes?

The URL hash typically refers to some portion of the document; clicking an anchor link will fast forward through the document to that portion. 

But is the behavior really any different from scrolling? Scrolling also moves the viewport to a different portion of the document. 

[...]

PR #555, opened 2019-05-05
Adds guide for intercepting hash URL changes

jamonholmgren:
Per @janicduplessis's [comment](https://github.com/react-native-community/react-native-webview/issues/24#issuecomment-483956651) on issue #24, there's a decent workaround to capturing hash URL changes. I've added this to the Guide.md file.

There are a bunch of Prettier updates as well. We should probably run Prettier on the whole project and add `lint-staged` and `husky` to avoid this issue in the future. I'll work on that if @Titozzz agrees.


janicduplessis:
Nice! Thanks @jamonholmgren

Issue #189, opened 2018-12-03
I hope listening to the loading fail （500 or 404 or 401 ）


cuibeihong:
My current plan ：（iOS）
add WKNavigationDelegate in RNCWKWebView.m

- (void)webView:(WKWebView *)webView decidePolicyForNavigationResponse:(WKNavigationResponse *)navigationResponse decisionHandler:(void (^)(WKNavigationResponsePolicy))decisionHandler{
  if ([navigationResponse.response isKindOfClass:[NSHTTPURLResponse class]]) {
    
    NSHTTPURLResponse * response = (NSHTTPURLResponse *)navigationResponse.response;
    if (response.statusCode == 401 || response.statusCode == 500 || response.statusCode == 404) {
      if (_onLoadingError) {
        NSMutableDictionary<NSString *, id> *event = [self baseEvent];
        [event addEntriesFromDictionary:@{
                                          @"didFailProvisionalNavigation": @YES,
                                          @"code": @(response.statusCode),
                                          }];
        _onLoadingError(event);
      }
    }
  }
  decisionHandler(WKNavigationResponsePolicyAllow);
}

Is there a better way to do it？

TheAlmightyBob:
Have you tried this code already? I _think_ that the call to `decisionHandler(WKNavigationResponsePolicyAllow)` will allow the page (e.g. the 404 error page) to successfully load and thus reset the `react-native-webview` error state. So if your app is using something like `renderError` then you might see your own error screen quickly appear and disappear.

(but if your app is just handling `onError` as a one-time notification to go spin off some other logic, not caring when the error state gets cleared, then this might be fine)

Also note that this won't just fire for the top-level page. If the main page you're loading succeeds but then has an embedded resource that gets a 404, you'll get the same error. (again depending on your scenario that may not be a problem)

PR #558, opened 2019-05-07
Http error callback (Catch http error statuscode)

manuelblum:
Catching and handling of HTTP Errors (Client/Server Error Codes)

We often have the requirement in our apps to handle certain http errors/status codes. 

For example, if a specific url returns a "Statuscode 401 - Not Authorised". This is the indicator that the users access token has been expired or has been revoked. So if we receive a 401 status code, we try to refresh the access token via api call or logout the user on refresh failure.

Another use case could be to present a native error view if a server error (Statuscode: 500) happens.

Eventually `onError` and `onHttpError` could be combined to `onError` but we explicitly decided that we do not want to do this, because:
- Handling http errors and network errors should be two different things.
- Native implementation (ios and android) has also separated catching status code into separate functions.
    - `-(void)webView:(WKWebView *)webView decidePolicyForNavigationResponse:(WKNavigationResponse *)navigationResponse decisionHandler:(void (^)(WKNavigationResponsePolicy))decisionHandler` (iOS)
    - `void onReceivedHttpError(WebView webView, WebResourceRequest request, WebResourceResponse errorResponse)` (Android)

We think that this feature is not that uncommon and we are looking forward to see this feature being shipped. What do you think about it?

Feature works on iOS and Android >= 21 (Lolipop). As far as i see, there is no way to support Android < 21.

Closes Issue #189


ios/RNCWKWebView.m
```diff
@@ -718,6 +719,25 @@ - (void)                  webView:(WKWebView *)webView
   decisionHandler(WKNavigationResponsePolicyAllow);
 }
 
+-(void)webView:(WKWebView *)webView decidePolicyForNavigationResponse:(WKNavigationResponse *)navigationResponse decisionHandler:(void (^)(WKNavigationResponsePolicy))decisionHandler
```
TheAlmightyBob: `decidePolicyForNavigationResponse` will also be invoked for nested frames. Could this handler avoid reporting those or provide the info with the event?

I've been using `isTopFrame = [response.URL isEqual:webView.URL];` in a local patch to detect this, based on the idea that by the time this is invoked `webView.URL` should have already been updated to be the main document URL.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -627,6 +630,24 @@ public void onReceivedError(
         new TopLoadingErrorEvent(webView.getId(), eventData));
     }
 
+    @Override
+    public void onReceivedHttpError(WebView webView, WebResourceRequest request, WebResourceResponse errorResponse) {
```
TheAlmightyBob: Could this use `request.isForMainFrame()` to filter out errors from nested frames or include with the JS event whether or not it's for the main frame?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -627,6 +630,24 @@ public void onReceivedError(
...
+      super.onReceivedHttpError(webView, request, errorResponse);
+      int statusCode = 0;
+      if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
+        statusCode = errorResponse.getStatusCode();
+        if(statusCode >= 400 && statusCode < 600)
+        {
+          WritableMap eventData = createWebViewEvent(webView, request.getUrl().toString());
+          eventData.putDouble("statusCode", statusCode);
```
TheAlmightyBob: Might be helpful to include `errorResponse.getReasonPhrase()` as well?

TheAlmightyBob:
+1 to keeping `onError` and `onHttpError` separate. For example, in some cases it may be desirable to just go ahead and show the HTTP error page from the server rather than hiding it.

jamonholmgren:
Looks like there's a small conflict, @manuelblum, can you resolve?

Titozzz:
Plus answering all the good point @TheAlmightyBob made 😀

avertin:
This is something that would be super helpful for my project as well. +1 on this PR

calmyournerves:
Would still be great to have this in, any chance to make it happen @manuelblum?

PR #567, opened 2019-05-10
Add focus functionality for devices without touch screen

alimyhamed:
(faced problem while developing for android TV, cause there only remote controller for device)


Titozzz:
Are we sure there is no side effect, if you load a webview next to a form, or anything else? I'm not confident with the approach and would rather have a focus() method you could call on the ref

alimyhamed:
> This looks good to me. Have you tested it on the device in question, @alimyhamed?

Yes, I've tested it on tv device, this is the only option that helped me, at least trigger focus on input.

alimyhamed:
> Are we sure there is no side effect, if you load a webview next to a form, or anything else? I'm not confident with the approach and would rather have a focus() method you could call on the ref

It would be great, but its the only option that I found to fix the issue. 
For more, I don't have enough knowledge.
In our app, we need to open a bank payment form and focus on the input to open the keyboard and start interacting with a form. And it's pretty hard if you don't have a possibility to focus or click.

alimyhamed:
Thanks for the help!

Titozzz:
I sadly don't have much time right now, I'll tag this as needing help!

jamonholmgren:
@Titozzz I think this is worth implementing as a small change. If we get reported problems, we can roll it back easily (1-line) and implement as a ref method like you suggest.

jamonholmgren:
After discussion with @Titozzz , we think this will cause problems when released.

As an alternative, @alimyhamed, can you implement as a ref method? You can see examples of how to do that by searching the code base for `goBack`.

https://github.com/react-native-community/react-native-webview/search?q=goBack

Try looking over how that's implemented and see if you can take a crack at it. Make sure to updated documentation too.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -470,6 +472,8 @@ public Map getExportedCustomDirectEventTypeConstants() {
       "injectJavaScript", COMMAND_INJECT_JAVASCRIPT,
       "loadUrl", COMMAND_LOAD_URL
     );
+    map.put("requestFocus", COMMAND_FOCUS);
```
Titozzz: You don't need to store mapbuilder to a variable, you can just add it above

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -379,6 +380,7 @@ public void setSource(WebView view, @Nullable ReadableMap source) {
           }
         }
         view.loadUrl(url, headerMap);
+        view.requestFocus();
```
Titozzz: Remove this so focus only happens when user calls the function

Titozzz:
Thanks for taking the effort. Just a few fixes needed

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
alimyhamed: Forget about this line.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
alimyhamed: Sorry, I'm not sure how.
Cause MapBuiler is 7 items max.
image

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
alimyhamed: https://github.com/react-native-community/react-native-webview/pull/567/commits/5c31cf3b12ba94569b17bd2fdabb2d54518d8b86

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
alimyhamed: Can someone help with mapbuilder fix?

WarrenMoore:
Could I ask what's left fix-wise on this before you'd be able to consider merging it in? I see your concern @Titozzz is with how `MapBuilder.of` isn't being used but as @alimyhamed pointed out overrides only exist for a maximum of 7 pairings (https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/common/MapBuilder.java#L84).

PR #577, opened 2019-05-17
fix Java syntax error

452MJ:
`getJSEventName()` needs a ScrollEventType parameter.

```java
public static String getJSEventName(ScrollEventType type) {
    switch (type) {
      case BEGIN_DRAG:
        return "topScrollBeginDrag";
      case END_DRAG:
        return "topScrollEndDrag";
      case SCROLL:
        return "topScroll";
      case MOMENTUM_BEGIN:
        return "topMomentumScrollBegin";
      case MOMENTUM_END:
        return "topMomentumScrollEnd";
      default:
        throw new IllegalArgumentException("Unsupported ScrollEventType: " + type);
    }
  }
```


452MJ:
#578 the same.

dvicory:
@Titozzz can we also apply this PR to a branch with 5.9.0 and release 5.9.1? Currently it's broken and 6.0.0 is a pretty major breaking change.

Titozzz:
What do you mean it's a pretty major breaking change ? Unless if you have a very custom build that requires custom buidtools it shouldn't even be breaking

PR #578, opened 2019-05-17
Fixed SCROLL  getJSEventName compatibility

htala:
Fixed SCROLL  getJSEventName compatibility with newer react-native versions to fix this error:
method getJSEventName in enum ScrollEventType cannot be applied to given types;


Titozzz:
Thanks, I merged the first PR since the code was identical

PR #587, opened 2019-05-21
Remove hardcoded white backgroundColor

atlewee:
I need my webview to have transparent backgroundColor.
(My webview displays augmented reality content on top of a video)
I have tried:
<WebView style = {{backgroundColor: "transparent"}}>
and 
<WebView backgroundColor = {"transparent"}> 
but none of them work, it is always white because of this hardcoded white backgroundColor.

When I remove the hardcoded white line from the styles file, I can select whatever color I need with the backgroundColor property, also transparent.
( Tested on version 5.9.1 )

## Compatibility
| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/src/WebView.styles.ts
+++ b/src/WebView.styles.ts
@@ -46,7 +46,7 @@ const styles = StyleSheet.create<Styles>({
     marginBottom: 10,
   },
   webView: {
-    backgroundColor: '#ffffff',
+  
   },
 });
 


KoenLav:
Might it be a good idea to leave the backgroundColor which is currently set as a default?

So if no backgroundColor provided then white, otherwise whatever is provided.

atlewee:
I dont know, but i belive the default
Content /wepage background is also white.
But as long as its possible to set transparent background somehow, Im happy :)

Titozzz:
I agree it should be possible to override it, but maybe we should not change the default as @KoenLav said, I'll try to look at this later if you don't have time.

Sexual:
Any update regarding this?

brandons:
Without being able to change the background color it results in a white flash before our content appears. @Titozzz can we please get this merged?

Titozzz:
I just checked, it's possible to override the style to affect the default behavior. Eg:  style={{backgroundColor: 'transparent'}} so I don't understand the need for the PR

brandons:
@Titozzz That is not what I'm seeing. Setting the background color has no effect at all on my end, it is still white. I just started migrating our implementation from the built in web view in React Native 0.59 and this issue was not present there.

brandons:
@Titozzz It is clearly an issue for more than just me. https://github.com/react-native-community/react-native-webview/issues/318

Titozzz:
Does removing the background fixes it, cause I don't see this on my end..? Maybe this is instead due to the way WKWebview is integrated. Were you using UI or WK?

joelennon:
We've had an issue where the WebView (using WK) flickers white before changing to the desired background colour, and are currently working around it by changing the `opacity` after the content has loaded, which is obviously a little hacky.

robertying:
This feature is really needed when I tried to apply dark mode to the webview. The way to set backgroundColor to transparent does not work so well. As @joelennon pointed out, there would be flickers from the default white to the color we use.

Titozzz:
fixed by #868

PR #592, opened 2019-05-23
add-javaScriptEnabled-in-IOS

aksmfosef11:
# Summary
When opening Google Docs in the IOS WebView
If `JavaScriptEnabled` is true, Google Docs will have a header.
There is also a page called Clear the cache and the cookies.

I fixed this problem by adding `javaScriptEnabled` to IOS.

One `javaScriptEnabled` variable and five lines of code have been added.

I used javaScriptEnabled in `WKPreferences`.

In addition to the issues I mentioned, you can solve problems that occur when iOS can not use `javaScriptEnabled` as false.

## Test Plan

Just add 5 lines to RNCWKWebView.m.

```
WKPreferences * prefs = [[WKPreferences alloc] init];
     if (! _ javascriptEnabled) {
       prefs.javaScriptEnabled = NO;
       wkWebViewConfig.preferences = prefs;
     }
```

The simplest test is to open Google Docs with WebView in IOS.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RNCWKWebView.h
+++ b/ios/RNCWKWebView.h
@@ -45,6 +45,7 @@
 @property (nonatomic, copy) NSString *userAgent;
 @property (nonatomic, copy) NSString *applicationNameForUserAgent;
 @property (nonatomic, assign) BOOL cacheEnabled;
+@property (nonatomic, assign) BOOL javaScriptEnabled;
 @property (nonatomic, assign) BOOL allowsLinkPreview;
 @property (nonatomic, assign) BOOL showsHorizontalScrollIndicator;
 @property (nonatomic, assign) BOOL showsVerticalScrollIndicator;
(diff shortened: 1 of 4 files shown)


Titozzz:
Please update types also :)

aksmfosef11:
@Titozzz 
How do I update types?
Is it correct to edit the` WebView.ios.tsx` file?

Titozzz:
Moved types from android to shared

mars-lan:
For whatever reason this broke loading Instagram mobile site in WebView. It's stuck in the initial loading screen.

Screen Shot 2019-08-29 at 7 37 37 PM

PR #598, opened 2019-05-24
Fix cookies not added to Android download request

atlatosBast:
# Summary

We never got any cookies on our server backend from a download request with the approach of retrieving the cookies with the base URL. If we just use the URL provided by onDownloadStart to retrieve the cookies everything works as expected.

It would be great if someone could confirm this.

Attempt to download a file which needs cookies (e.g. for authentication).

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist


atlatosBast:
Ping @jamonholmgren @Titozzz

PR #603, opened 2019-05-28
fix(iOS WKWebView): contentInset not properly applied

ptrocki:
Resolve
https://github.com/react-native-community/react-native-webview/issues/559
https://github.com/react-native-community/react-native-webview/issues/513

## Checklist


chrise86:
This is great, but `scrollIndicatorInsets` is missing :(

PR #614, opened 2019-05-31
fixes the problem of tag <input type='file'> without attribute accept…

timothyawest:
…=' not allowing user to choose to take a picture or video with the camera. Before it only allowed file upload without explicitly declaring the mime types in accept attribute now it will allow the option of taking picture or video as most browsers do.  Can be tested at: https://andreipfeiffer.github.io/react-native-webview-android-file-upload/index.html

# Summary

file upload

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |   /not necessary i think
| Android |    ✅     |

## Checklist


Titozzz:
Closing as this code in contained in your other PR #615

PR #615, opened 2019-05-31
Bugfix input file

timothyawest:
# Summary
this fixes a number of issues with the file upload when using <input type="file" without the attribute accepts. It fixes the inability for the user to use camera or video in addition to file upload. (the default behavior on most browsers). I also fixes is so that when the accepts attribute is not used the user can upload any type of file (the default behavior on most browsers).   It still can't choose multiple files... Not sure why... Maybe will fix that later. Check it out using: https://andreipfeiffer.github.io/react-native-webview-android-file-upload/index.html

RNCWebViewModule.java
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
@@ -282,7 +282,7 @@ public class RNCWebViewModule extends ReactContextBaseJavaModule implements Acti
 
   private Boolean acceptsImages(String[] types) {
     String[] mimeTypes = getAcceptedMimeType(types);
-    return isArrayEmpty(mimeTypes) || arrayContainsString(mimeTypes, "image") || arrayContainsString(mimeTypes,"*/*"); //mimeTypes is never really empty it has */* in it if webpage input="file" does not have tag 'accept="image/*"'
+    return isArrayEmpty(mimeTypes) || arrayContainsString(mimeTypes, "image");
   }
 
   private Boolean acceptsVideo(String types) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
```diff
@@ -250,39 +250,46 @@ private Intent getVideoIntent() {
...
 
   private Intent getFileChooserIntent(String[] acceptTypes, boolean allowMultiple) {
+    String[] _acceptTypes = acceptTypes.clone();
     Intent intent = new Intent(Intent.ACTION_GET_CONTENT);
     intent.addCategory(Intent.CATEGORY_OPENABLE);
     intent.setType("*/*");
-    intent.putExtra(Intent.EXTRA_MIME_TYPES, getAcceptedMimeType(acceptTypes));
+    if(isArrayEmpty(_acceptTypes) )
+      _acceptTypes[0] = "*/*";
+    for(int i=0;i<_acceptTypes.length;i++){
```
Titozzz: Leftover code?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
```diff
@@ -295,7 +302,7 @@ private Boolean acceptsVideo(String types) {
 
   private Boolean acceptsVideo(String[] types) {
     String[] mimeTypes = getAcceptedMimeType(types);
-    return isArrayEmpty(mimeTypes) || arrayContainsString(mimeTypes, "video");
+    return isArrayEmpty(mimeTypes) || arrayContainsString(mimeTypes, "image") || arrayContainsString(mimeTypes, "*/*");
```
Titozzz: this should still be video, right?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
jamonholmgren: Looks like it.

timothyawest:
Hi all,
There is still a sticking point that was there before and still there in
how the files are named... Working the issue now.   It is a bit of a
challenge in that I have to find a way to resolve the intent.
Tim

PR #619, opened 2019-06-03
Fix onShouldStartLoadWithResquest.

ptrocki:
Resolve 

https://github.com/react-native-community/react-native-webview/issues/185

# Summary

onShouldStartLoadWithRequest does not work on Android, when the first url is loaded.  

## Test Plan

Just test, if onShouldStartLoadWithRequest is called after first url load. 

## Checklist


Titozzz:
OnShouldStartLoadWithRequest isn't just an event you listen to, you need to be able to stop the current request from starting. This is why it was trickier to implement on android.

ptrocki:
Close, because it does not solve the problem.

kulbhushan-ucreate:
Hey folks,
I am also stuck in this is an issue from webview and finally am fix this issue by getting a fork from react-native-community/react-native-webview by doing some android native code.
You guys check out my git package.

https://github.com/kulbhushan-ucreate/react-native-web-view.

Issue #640, opened 2019-06-14
`window.history` API does not generate events on iOS


swansontec:
# Problem

Web pages can use the `window.history` API to manipulate the back button & browser location without reloading the page. This is a popular technique in modern apps, and should work correctly in the `WebView` component.

On iOS however, navigating this way doesn't generate any events. This means there is no way for the host application to update the document URL or back button behavior, leading to a broken user experience.

## To reproduce

Visit the [demo page](https://www.swansontec.com/temp/history/a.html). The "push" buttons create virtual history entries, and the "back" and "forward" buttons navigate using the `window.history` API.

* In a browser, you can watch the address bar and browser history change.
* In an Android WebView, these navigation actions generate `onLoad`, `onLoadEnd`, & `onNavigationStateChange` events.
* In an iOS WKWebView, these navigation actions don't trigger any events (the bug).

## Expected behavior

API's like `window.history.pushState` should trigger the same events on iOS as they do on Android (like `onNavigationStateChange`).

## Environment:
 - OS: iOS 11
 - react-native version: 0.59
 - react-native-webview version: 5.8.2

swansontec:
## Possible solution 1: Match Android behavior

If iOS were to generate the appropriate `onNavigationStateChange` events for non-navigation history changes, this would allow existing apps to keep working.

The current way to implement a smart back button is to capture a `canGoBack` flag using the `onNavigationStateChange` callback:

```javascript
<WebView
  onNavigationStateChange={navState => {
      this.canGoBack = navState.canGoBack
  }}
  ref={element => {
      this.webView = element
  }}
/>
```

Now, the back button handler can use `this.canGoBack` to decide whether to go back within the page, or leave the scene:

```javascript
onBackButtonPress () {
  if (this.webView && this.canGoBack) {
    this.webView.goBack()
  } else {
    this.props.navigation.pop() // Return to the main app
  }
}
```

This solution isn't great, though, both the end-user has to do extra book-keeping to track the `canGoBack` flag.

I also tried implementing this solution in #641, but it was harder than expected, and my solution stil has issues. Getting this work 100% will be hard.

## Possible solution 2: Provide a `canGoBack` flag

Perhaps the `WebView` could track the `canGoBack` flag on its own, exposing it as a property. This would prevent the user from having to implement an `onNavigationStateChange` callback:

```javascript
onBackButtonPress () {
  if (this.webView && this.webView.canGoBack) {
    this.webView.goBack()
  } else {
[...]

PR #641, opened 2019-06-14
Hook the `window.history` API on iOS to generate events

swansontec:
# Summary

This pull request solves #640, where navigating the WebView using the `window.history` API doesn't generate any events on iOS. This seems to be a low-level problem with the `WKWebView`, since none of the `WKNavigationDelegate` methods fire in these cases.

Since the native code doesn't tell us when the location changes this way, the semi-hacky solution is to inject Javascript into the user page. This Javascript intercepts calls to the `window.history` API and sends an event to the native code when this happens. The native code then simulates a `onLoadingFinish` event with `{ navigationType: 'history' }` in the payload.

## Test Plan

I created a fresh `react-native init` project, and replaced `render` with the following code:

```javascript
  render() {
    return (
      <WebView
        style={{ flex: 1, alignSelf: "stretch" }}
        source={{ uri: "https://www.swansontec.com/temp/history/a.html" }}
        onNavigationStateChange={event =>
          console.log(event.url, event.canGoBack, event.navigationType)
        }
      />
    );
  }
}
```

Then, I ran the debugger while performing the same sequence of events on various platforms:

* push a
* push a
* back
* back

Using the unmodified react-native-webview on iOS generated no output. After the changes, iOS generates the following logs:

```
https://www.swansontec.com/temp/history/a.html#1 true history
https://www.swansontec.com/temp/history/a.html#2 true history
https://www.swansontec.com/temp/history/a.html#1 true history
https://www.swansontec.com/temp/history/a.html false history
```

And Android generates:

```
https://www.swansontec.com/temp/history/a.html#1 true undefined
https://www.swansontec.com/temp/history/a.html#2 true undefined
https://www.swansontec.com/temp/history/a.html#1 true undefined
https://www.swansontec.com/temp/history/a.html false undefined
```

So now the behavior matches!

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅ (no change)     |

## Checklist

### DIFF
--- a/ios/RNCWKWebView.m
+++ b/ios/RNCWKWebView.m
@@ -14,6 +14,7 @@
 #import "objc/runtime.h"
 
 static NSTimer *keyboardTimer;
+static NSString *const HistoryShimName = @"ReactNativeHistoryShim";
 static NSString *const MessageHandlerName = @"ReactNativeWebView";
 static NSURLCredential* clientAuthenticationCredential;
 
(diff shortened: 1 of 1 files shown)


swansontec:
I should mention that although the basic case works great, there are still some glitches. For instance, using the `window.history.back` API generates a spurious `onLoad` event in cases where the page *actually* needs to navigate to a new location, as opposed to manipulating the local browser history.

Fixing the glitch is hard, since the injected Javascript doesn't have enough information to distinguish the two cases. There would need to be some type of state machine in the native code to truly filter these out.

On the other hand, the main use of the `window.history.back()` API is to achieve in-page history manipulation for single-page web apps, so cases that would generate these glitches should be super-rare in the wild.

swansontec:
Found another problem: Calling `webViewRef.goBack()` from the React Native side does not generate events either on iOS. The page does navigate backwards through its virtual history, but the WebView never generates the `onNativationStateChange` events. This will need to be fixed as well for feature-parity with Android.

swansontec:
This approach doesn't seem viable. I will submit another pull request with a different solution.

PR #688, opened 2019-06-29
Added postMessage to Typescript index.d.ts declaration

hsource:
postMessage is defined with typings in the following files, but not exposed via index.d.ts. This fixes that

Android:
https://github.com/react-native-community/react-native-webview/blob/d61d335f80a9d9aa613f00e4e2592dba3f289bb4/src/WebView.android.tsx#L108

iOS:
https://github.com/react-native-community/react-native-webview/blob/d61d335f80a9d9aa613f00e4e2592dba3f289bb4/src/WebView.ios.tsx#L177

# Summary

postMessage wasn't type-checking correctly in Typescript

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -34,6 +34,13 @@ declare class WebView extends Component<WebViewProps> {
      * Executes the JavaScript string.
      */
     injectJavaScript: (script: string) => void;
+    
+    
+    /**
+     * Posts a message that can be listened to by the page via
+     * `window.addEventListener`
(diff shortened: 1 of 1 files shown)


Titozzz:
PostMessage will be removed, it's already deprecated in favor or injectJavascript, see #438

PR #694, opened 2019-07-04
Support loading local files on WebKitWebView iOS

adbs1:
https://github.com/react-native-community/react-native-webview/issues/137#issuecomment-497304273

# Summary

Allows read access in iOS to the directory where the WebView html source file is loaded from. Previously it only allowed access to the html source file itself.
Motivation is the many scenarios where files are downloaded by the app and must be readable by WebView (think map tiles.) In Android this works out of the box, iOS however that's not the case.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |

## Checklist
- [x] I  >and others<  have tested this on a device and a simulator
- [N/A] I added the documentation in `README.md`
- [N/A] I mentioned this change in `CHANGELOG.md`
- [N/A] I updated the typed files (TS and Flow)
- [N/A] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/ios/RNCWKWebView.m
+++ b/ios/RNCWKWebView.m
@@ -398,7 +398,7 @@ static NSURLCredential* clientAuthenticationCredential;
         [_webView loadRequest:request];
     }
     else {
-        [_webView loadFileURL:request.URL allowingReadAccessToURL:request.URL];
+        [_webView loadFileURL:request.URL allowingReadAccessToURL:[request.URL URLByDeletingLastPathComponent]];
     }
 }
 


Liestambeur:
This will allow loading local files in the same folder of the file loaded. But not to files somewhere else is the filesystem. 
For example: I open a html file in the local folder 'data/subfolder/page.html' that references a css ans js file: 'data/js.js' and 'data/css.css'
This is solved if I delete the last 2 parts of the url:
`        [_webView loadFileURL:request.URL allowingReadAccessToURL:[[request.URL URLByDeletingLastPathComponent] URLByDeletingLastPathComponent]];
`
Can we maybe make it a variable that we can pass so we can cover all use cases?

adbs1:
@Liestambeur Agreed and that was my suggestion also in the bug comments. It's easy to work around that by moving files where you need them outside of WebView.

Then as now I don't have time to do the variable work and test it thoroughly, it's sat here for weeks and people have asked for a PR as it is. I don't see any harm pushing it as is and perhaps adding a feature request for the variable option. It's a better scenario than this sitting here another 6 month ??

Liestambeur:
@adbs1 agree :). I just commented since I'm working on a project where I cannot change the folder structure. If I have some time this week I'll try to make a clean fix for this.

animaonline:
Any updates on this?

hsource:
#771 is a more general solution for this problem and reflects @Liestambeur 's feedback

Titozzz:
Closed in favor of #771

PR #697, opened 2019-07-04
Fix `originWhitelist` escape vulnerablity

swansontec:
# Summary

If the user specifies a specific domain for the origin whitelist, such as `https://example.co`, any domain that starts with this pattern will be valid, including `https://example.com` or `https://example.co.malicious-site.com`.

Fix this by adding a `$` to the end of the regular expression, requiring the full origin to match the provided pattern.

Note that this may be a breaking change if people are using patterns like `"https://"`, since these will no longer match. The fix is to use a wildcard, as shown in all the documentation and examples (`"https://*"`).

## Test Plan

The unit tests have been updated to exercise this vulnerability. Without the fix in place, the newly-added tests would fail.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] I added the documentation in `README.md` (Actually `Reference.md`)
- [ ] I mentioned this change in `CHANGELOG.md` (File does not exist!!!)

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -394,7 +394,9 @@ Note that this method will not be invoked on hash URL changes (e.g. from `https:
 
 ### `originWhitelist`
 
-List of origin strings to allow being navigated to. The strings allow wildcards and get matched against _just_ the origin (not the full URL). If the user taps to navigate to a new page but the new page is not in this whitelist, the URL will be handled by the OS. The default whitelisted origins are "http://*" and "https://*".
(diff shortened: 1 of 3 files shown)


Titozzz:
This is a very good PR, but these are the reason I want to remove extra props such as OriginWhitelist #439. I'll definitly merge that PR before tho, as it's kind of very dangerous.

jamonholmgren:
Love it. Thanks for the contribution @swansontec !

swansontec:
Seriously? This security vulnerability is still present in all released versions. Why has this not been merged?

cnotin:
IMO this is a minor security issue and should be merged!

chrisbobbe:
@github-actions [said](https://github.com/react-native-community/react-native-webview/pull/697#issuecomment-528652775):
> If you think this is a mistake please comment and ping a maintainer to get this merged ASAP!

Sure! 🙂 @Titozzz

PR #712, opened 2019-07-10
Linking Error handling

sandeepsh3090:
When we have open url that redirected to installed app. than react-native-webview not handle error to redirectting on that app.

when we are try open url like(https://apps.apple.com/in/app/facebook/id284882215) which is open in app store than Linking.openURL(url) can open url but if there is any error occure than we don't know any status that why any lodder visible always. so  to Know status of Linking.openURL(url) we have catch error.

# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist


Titozzz:
#439 will address theses kind of issue

PR #771, opened 2019-08-08
Add prop for allowingReadAccessToURL on iOS WKWebView

hsource:
# Summary

By default, the WKWebView sets `allowingReadAccessToURL` to just the file itself when provided with `source={{ uri: 'file://myFile' }}`. This means that it's not possible to access any other CSS or JS on the filesystem from within the WebView.

This change adds a prop that lets users customize `allowingReadAccessToURL`. The default is still to only allow access to just the file itself, but users can now override it.

**Important: allowingReadAccessToURL is ignored on the emulator. This must be tested on your device**

## Test Plan

See https://github.com/hsource/test-allowing-read-access/tree/master

To test that it works:

1. Clone the test repository `git clone https://github.com/hsource/test-allowing-read-access.git`
2. Go into the directory and run `yarn install` and `cd ios; pod install; cd ..`
3. Go into Xcode and make sure the project is signed with a development team so it can be loaded onto device
4. Run on a real iOS device by running `react-native run-ios --device`

You should see that in the first WebView, the JS doesn't replace the first line and the second line isn't red, while in the second WebView, the first line is replaced by the JS and the second line is red.

You can check that both WebViews don't execute JS or apply styles in previous versions of React Native WebView by doing:
1. `yarn add react-native-webview` - this installs the last published version of WebView
2. `cd ios; pod install; cd ..`
3. `react-native run-ios --device`

Now both WebViews should have black text that's not replaced by JS.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌ (not needed)     |

## Checklist

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -41,6 +41,7 @@ This document lays out the current public properties and methods for the React N
 - [`directionalLockEnabled`](Reference.md#directionalLockEnabled)
 - [`geolocationEnabled`](Reference.md#geolocationenabled)
 - [`allowUniversalAccessFromFileURLs`](Reference.md#allowUniversalAccessFromFileURLs)
+- [`allowingReadAccessToURL`](Reference.md#allowingReadAccessToURL)
 - [`useWebKit`](Reference.md#usewebkit)
 - [`url`](Reference.md#url)
 - [`html`](Reference.md#html)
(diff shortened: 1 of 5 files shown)


hsource:
For the output of the test project, refer to these screenshots:

With patch:
New

Without patch:
Old

Issue #503, opened 2019-04-15
Add injectedJavaScriptBeforeDocumentLoad


bashen1:
This is very important to me. When will release be released? [#397](https://github.com/react-native-community/react-native-webview/pull/397)

Titozzz:
When it's ready and reviewed... Please comment on the PR instead of creating issues

PR #784, opened 2019-08-15
Scheme Handlers

ccorcos:
This PR is a rebased version of https://github.com/react-native-community/react-native-webview/pull/209. 

I'm working on upgrading my app for iOS 13 so I thought I'd revisit this. I'd really like to get this thing merged so I don't have to keep rebasing on the mainline.

## To Do
- [x] Test on iOS
- [x] Test the Android code

## Test Plan

I've been running this code [in production](https://apps.apple.com/us/app/notion-notes-tasks-wikis/id1232780281) with over 100k users for almost a year now and it seems to work although I will need to test my new changes! 😉

### What's required for testing?

More information in the old PR. I can write a better description when this PR is actually ready.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅❌     |

## Checklist


android/build.gradle
```diff
@@ -53,6 +53,9 @@ repositories {
   jcenter()
   google()
 
+  // @rob do we need this?
+  maven { url "https://jitpack.io" }
```
ccorcos: @rbscott do we need this?

RangerMauve:
Woot! 🎉🎉🎉

android/build.gradle
rbscott: @ccorcos  Yeah, that was how I found the dependency for Persistent Cookie.

ccorcos:
I see. Cool. Will be testing Android soon. Almost done upgrading iOS to RN60.

Titozzz:
Glad you picked it up, please let me know when you are done so I can review

ccorcos:
@Titozzz @RangerMauve this is ready for review. I've done some basic testing that it works on both iOS and Android. Going to do some more intensive testing after upgrading to iOS 13.

One breaking change to note: at some point iOS moves from `document.addListener("message")` to `window.addListener("message")` but Android was not updated to be consistent, so I updated Android to reflect the same API.

ccorcos:
@Titozzz @RangerMauve what's going on here? Every time I have rebase, I have to QA the whole thing all over again 😭

RangerMauve:
Sadly I'm not a collaborator so my approval won't mean anything. 😅😅😅

ccorcos:
I see. @Titozzz ?

Titozzz:
There was an emergency fix needed (UIWebView removal) and I didn't want people to have 2 breaking changes in a row so I also renamed the files. Not much more since...

Titozzz:
I was waiting on the:
> I've done some basic testing that it works on both iOS and Android. Going to do some more intensive testing after upgrading to iOS 13.

src/WebViewTypes.ts
```diff
@@ -432,6 +503,19 @@ export interface IOSWebViewProps extends WebViewSharedProps {
...
+	 * Specify a url scheme to intercept.
+	 */
+	urlScheme?: string;
+
+  /**
+   * Intercept a url scheme request. If you return a status, that means this is a response. If there
+   * is a method, that means that we should make the request from native code and pipe it. This is
+   * useful for requests such as images that won't encode well.
+   */
+  onUrlSchemeRequest?: (event: WebViewUrlSchemeRequest) =>
```
Titozzz: Maybe share this prop with android? Same name / input and outputs

src/WebViewTypes.ts
```diff
@@ -542,6 +626,33 @@ export interface AndroidWebViewProps extends WebViewSharedProps {
    * Sets ability to open fullscreen videos on Android devices.
   */
   allowsFullscreenVideo?: boolean;
+
+  /**
+   * A Base URL that is intercepted and sent through onUrlSchemeRequest. Android has a few issues
+   * with custom schemes, using a baseInterceptURL instead.
+   */
+  baseInterceptUrl?: string;
```
Titozzz: How different would the value passed here be from iOS urlScheme props (trying to see if we can unite thoses)

ios/RNCWKWebView.m
```diff
@@ -540,6 +549,13 @@ - (void)setScrollEnabled:(BOOL)scrollEnabled
 {
   _scrollEnabled = scrollEnabled;
   _webView.scrollView.scrollEnabled = scrollEnabled;
+
```
Titozzz: Does this really comes from this PR?

android/build.gradle
```diff
@@ -53,6 +53,9 @@ repositories {
   jcenter()
   google()
 
+  // @rob do we need this?
```
Titozzz: Please remvoe this comment then

ios/RNCWKWebView.m
ccorcos: Looks like a merge issue.

https://github.com/react-native-community/react-native-webview/commit/2cef5cbb3ff5967a8f207b0c54340003ab1abc80

android/build.gradle
ccorcos: done

ccorcos:
Regarding the different APIs between iOS and Android, there is some more explanation here: https://github.com/ccorcos/react-native-webview/pull/1

It all comes down to issues with CORS.
- On Android, making a request to `https://www.example.com` from `example://www.example.com` is considered a cross-origin request. Even though we are proxying the request using `WebViewClient.shouldInterceptRequest`, the request header includes the origin `example://www.example.com` and therefore blocks the cross origin request. However, Android's `baseInterceptUrl` lets you intercept `http` and `https` traffic. So we simply serve a website as `https://www.example.com` and proxy those requests and then we no longer have CORS issues.
- On iOS, `WKURLSchemeHandler` doesn't let you intercept `http` or `https` traffic -- you have to use a custom scheme handlers. Conveniently, proxying a request from `example://www.example.com` does not cause and CORS issues.

src/WebViewTypes.ts
ccorcos: See comment below about the differences between Android and iOS.

src/WebViewTypes.ts
ccorcos: See comment below about the differences between Android and iOS.

ccorcos:
@Titozzz Sorry -- everything works on iOS 13! I forgot to mention that.

package.json
```diff
@@ -11,12 +11,12 @@
   "version": "7.0.3",
   "homepage": "https://github.com/react-native-community/react-native-webview#readme",
   "scripts": {
```
Titozzz: Please do not change this

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -550,25 +653,27 @@ public void receiveCommand(WebView root, int commandId, @Nullable ReadableArray
           RNCWebView reactWebView = (RNCWebView) root;
           JSONObject eventInitDict = new JSONObject();
           eventInitDict.put("data", args.getString(0));
+          // Make sure this is the same as iOS!
```
Titozzz: I'm not sure this is related to this PR + this is deprecated and will be removed soon so I don't really see a reason to change it

ccorcos: This is going to be removed? Are you just expecting this to be implemented by individuals themselves?

package.json
ccorcos: Sure.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
ccorcos: I can remove if you want, but I am relying on this for my app. It was changes on iOS but not Android so now they're inconsistent which is awkward.

Titozzz: For some context #438

ccorcos: Well that work can always be done on top of this one...

ccorcos:
@Titozzz if I merge master and fix the conflicts again, are you going to merge this PR? Otherwise, I won't waste my time.

jamonholmgren:
@ccorcos thanks as always for your really nice work. The core team is mainly concerned that it is a lot of code to maintain, even if high quality which yours always is. I’m curious if there’s a way to make it an extension or plugin that would work with the WebView? So it remains separate but consumes the latest version of WebView. We might still merge this into master but just wondering about that other option.

ccorcos:
@jamonholmgren a plugin system would be nice -- so long as I don't have to keep handling merge conflicts against this package...

Titozzz:
What would be the needed interface for this @ccorcos :)?

Titozzz:
This is the same dilema we are having with #503, huge quantity of code that is needed for some uses cases but not needed for everyone and potentially harder to maintain

jamonholmgren:
Bump to keep open.

ccorcos:
I think @Titozzz determined that is is not going to get merged :/

Titozzz:
@ccorcos I did not get an answer to the 

> What would be the needed interface for this @ccorcos :)?

So it kinda stayed hanging there

PR #791, opened 2019-08-19
Fix onShouldStartLoadWithRequest does not work on Android.

### LINKED ISSUE #185: onShouldStartLoadWithRequest don't work on Android
onShouldStartLoadWithRequest don't work on Android when the first url load -- v2.14.0

kulbhushan-ucreate:
Resolve

#185

Summary
onShouldStartLoadWithRequest does not work on Android, when the first url is loaded.

Test Plan
Just test, if onShouldStartLoadWithRequest is called after first url load.

Checklist
 I have tested this on a device and a simulator


Titozzz:
I'm not sure what you are trying to do here, you need to fork webview and then submit webview from your fork over here. Tell me if you need more help

Issue #610, opened 2019-05-30
Preventing videos playing when app is background (Especially Android, Youtube video)


cereme:
**Is your feature request related to a problem? If so, Please describe.**
My Android app displays youtube videos with react-native-webview, and is rejected from Google Play Store. The reason was "Device and Network abuse" and it appears when Youtube video can be played when the app is background.

**Describe the solutions you came up with**
I wrote the code in ```RNCWebViewManager.java```:
```java
protected static class RNCWebView extends WebView implements LifecycleEventListener {
//...
    @Override
    public void onHostResume() {
      super.onResume();
    }

    @Override
    public void onHostPause() {
      super.onPause();
    }
```
and I solved my problem.
I think adding "disableBackgroundPlaying" props and enabling my code above will work.

thebergamo:
There is any PR to add this feature?

cereme:
I couldn't find test case, "background media playing" in my devices(Android 8, 9), except Google Play Store's approval. Cause I'm not Android developer, if someone has good idea on that, please let me know or PR!

thebergamo:
In your case @cereme  did you forked this module and wrote that code inside? I'm dealing with same problem with Play Store.

I'm not sure if I should write a entire custom WebView just to add those method or if I should just fork it and use it like that for now.

cereme:
@thebergamo

> In your case @cereme did you forked this module and wrote that code inside? I'm dealing with same problem with Play Store.

Yes, I forked and wrote that code insdie and resolved Play Store issue

freesongs:
Hello, Can I have more information about this app that plays video in background which is something I am looking for? I don;t care about Google app, as this is needed for private use. Thanks!

PR #801, opened 2019-08-23
feat: Add prop useNativeResumeAndPauseLifecycleEvents

thebergamo:
# Summary
Add prop `useNativeResumeAndPauseLifecycleEvents` to take advance of `onResume` and `onPause` to avoid issues with Play Store.

This implementation is based on feature request by @cereme on issue #610 

We experienced problems while submitting new version of our apps to Play Store because of some content that play youtube videos inside a web view. 

Looks like, because of we're not using `onHostPause` it would cause some background activity in Android devices that is not permitted by [Play Store policies](https://play.google.com/about/privacy-security-deception/device-network-abuse/).  

**By the way, I'm not Android developer, so would appreciate some help to improve this PR even with just some guidance.**

Create a WebView that point to a page with a youtube video (could be directly to youtube website).

Minimize app and check logcat for background activity

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -490,6 +490,15 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   public void setOnScroll(WebView view, boolean hasScrollEvent) {
     ((RNCWebView) view).setHasScrollEvent(hasScrollEvent);
   }
+  
+  @ReactProp(name = "useNativeResumeAndPauseLifecycleEvents")
+  public void setUseNativeResumeAndPauseLifecycleEvents(WebView view, boolean useHostResumeAndPause) {
+    // This prop will ensure no background activity will occur while playing any media in a webview
+    // In cases where this prop is not set and a media streaming is consumed inside the WebView,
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -490,6 +490,15 @@ public void setGeolocationEnabled(
   public void setOnScroll(WebView view, boolean hasScrollEvent) {
     ((RNCWebView) view).setHasScrollEvent(hasScrollEvent);
   }
+  
+  @ReactProp(name = "useNativeResumeAndPauseLifecycleEvents")
+  public void setUseNativeResumeAndPauseLifecycleEvents(WebView view, boolean useHostResumeAndPause) {
+    // This prop will ensure no background activity will occur while playing any media in a webview
+    // In cases where this prop is not set and a media streaming is consumed inside the WebView,
+    // You might experience  "Device and Network abuse" policy problems in Play Store.
+    ((RNCWebView) view).setHasScrollEvent(useHostResumeAndPause);
```
cereme: Just typo? I think ```setHasScrollEvent``` should be ```setUseHostResumeAndPause```

PR #813, opened 2019-08-28
[WIP] Load webview even if not added to a window

philippeauriach:
Fix a problem where a webView could be rendered but not loaded, for example using `createBottomTabNavigator` with `lazy: false` (all tabs are rendered before been displayed, but webviews will wait to appear to load).

A code example can be found in the SO question here:  https://stackoverflow.com/questions/56923273/react-native-webview-only-starts-loading-when-on-active-tab-in-react-navigation

The problem do not appear on Android, therefore modifications are only in the iOS side.

### DIFF
--- a/ios/RNCWKWebView.m
+++ b/ios/RNCWKWebView.m
@@ -92,6 +92,8 @@ static NSURLCredential* clientAuthenticationCredential;
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(toggleFullScreenVideoStatusBars) name:@"_MRMediaRemotePlayerSupportedCommandsDidChangeNotification" object:nil];
   }
   
+  [self initWebView];
+  
   return self;
 }
 
@@ -111,9 +113,9 @@ static NSURLCredential* clientAuthenticationCredential;
   return nil;
 }
 
-- (void)didMoveToWindow
+- (void)initWebView
 {
-  if (self.window != nil && _webView == nil) {
+  if (_webView == nil) {
     WKWebViewConfiguration *wkWebViewConfig = [WKWebViewConfiguration new];
     if (_incognito) {
       wkWebViewConfig.websiteDataStore = [WKWebsiteDataStore nonPersistentDataStore];


Titozzz:
There was a reason for the code being inside didMoveToWindow, I ll try to remember why

philippeauriach:
> There was a reason for the code being inside didMoveToWindow, I ll try to remember why

Before my commit, there was a check for `if(window != nil)`, maybe something needs the window ? I looked into it but nothing seems to need the window.

philippeauriach:
> There was a reason for the code being inside didMoveToWindow, I ll try to remember why

@Titozzz , it seems that with my PR, some features do not work anymore (like `window.ReactNativeWebView` being undefined even if I have an `onMessage` prop). Using the debbuger, it seems that the property `_messagingEnabled` is set to NO when initializing the webview. Maybe it has something to do with when the properties are initiated ?

Titozzz:
It doesn't but it might be why it was implemented like this

jamonholmgren:
Bump to keep open.

shirakaba:
> There was a reason for the code being inside didMoveToWindow, I ll try to remember why

I'm digging up what I can find, but it sounds like it's been there since the Facebook implementation... Of note, [react-native-wkwebview](https://github.com/CRAlpha/react-native-wkwebview) doesn't implement anything equivalent and furthermore has no issues filed related to off-screen rendering (unlike this implementation).

https://github.com/react-native-community/react-native-webview/issues/696 said that: 1) Off-screen WebViews are not "loaded/rendered"; `didMoveToWindow` is called only when the WebView starts to be shown on-screen, and is triggered by the "addSubview" method in "react_updateClippedSubviewsWithClipRect".

https://github.com/react-native-community/react-native-webview/issues/81 said that, without the `_webView != nil` condition within `didMoveToWindow`:

> I have a root page include three webViews. And this code would result to some situations like, I push a new native controller, and when I back to origin page, it would recreate three webViews and add above the origin webView. So memory cost more and more high, the interaction more and more slow.

This was addressed in PR https://github.com/react-native-community/react-native-webview/pull/84

So I think it's something like an optimisation for off-screen rendering, but I would really appreciate if – provided anyone knows – it could be documented so that we can know which properties of the WebView are initialised at the moment of setting props on an off-screen WebView.

adimshev:
+1

goleary:
I'd love to see this capability, is anyone opposed to reviving this PR?

focux:
A heads up: I was having the same issue with the tab bars and patched my version locally with these changes and it worked perfectly. Thank you @philippeauriach.

samuelg0rd0n:
> **EDIT:** I'm having some issues with this solution. It seems that when the webview is created, the config is not ready so it's not using the config that you pass from the JS side.

@focux I'm sorry but what config do you mean? You mean like all props that you pass to WebView component?

PR #815, opened 2019-08-28
Allow user to rotate fullscreen video on Android (No Android X)

AkeemMcLennon:
# Summary

This change updates the fullscreen video support in Android to match the behavior in iOS so that the user can rotate the screen regardless of the default orientation defined in the app. For example, if `android:screenOrientation="portrait"` is defined within the manifest, then the video will always displayed in portrait mode regardless of the aspect ratio in the video. Related to https://github.com/react-native-community/react-native-webview/issues/586. Also see Android X version of this PR at https://github.com/react-native-community/react-native-webview/pull/816  

This change also avoids some of the pitfalls of forcing screen orientation discussed in https://github.com/react-native-community/react-native-webview/pull/325#pullrequestreview-203973316 
- Set `android:screenOrientation="portrait"` in AndroidManifest.xml
- Open a webview with an embedded landscape video containing a full screen button
- Rotate the screen between landscape / portrait mode

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅     |

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -5,6 +5,7 @@ import android.annotation.TargetApi;
 import android.app.DownloadManager;
 import android.content.Context;
 import android.content.Intent;
+import android.content.pm.ActivityInfo;
 import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.net.Uri;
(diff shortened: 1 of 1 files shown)


Josephdias92:
Hello @Titozzz,
Can you create a release out of this for non androidX users.
create a release version with 5.12.2

wulaizi:
This fix has no effect. My webview version is 5.12.1. It can be seen from the upgrade document that it has been fixed, but it is not. My Android still cannot work like ios, and the video can be rotated horizontally. full screen      android:screenOrientation="portrait"

PR #829, opened 2019-08-30
Update README to improve first encounter

schmijos:
I had problems when I tried `react-native-webview` for the first time in a freshly created react native 0.60 project. It just showed a blank page because width and height were zero. Scaling the page should at least show something and improve the initial experience with this library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ import { WebView } from 'react-native-webview';
 class MyWebComponent extends Component {
   render() {
     return (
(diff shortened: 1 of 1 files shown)


Titozzz:
I'm not sure why settings scalePagesToFit fixes your issue but please keep in mind that WebView is just a view so if you want it to fill your screen please use flex: 1

Titozzz:
I'll update the example to reflect this

PR #845, opened 2019-09-04
Fix(iOS):use RCTConvert to convert allowingReadAccessToURL.

hank121314:
# Summary
It's not safe to convert NSString to NSURL by URLWithString directly.
If NSString is not a valid URL (maybe contains some special characters), it will return nil.
This PR change URLWithString into RCTConvert.
Not only fix the above issue but prop allowingReadAccessToURL also don't need the file:// prefix.

## Test Plan

[Test repository](https://github.com/hank121314/iOS_WKWebview_Test)(reference to: https://github.com/hsource/test-allowing-read-access/tree/master)

1.Clone the test repository and
`yarn && cd ios && pod install`
2.Run on a real device.

## Result
image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |

## Checklist

### DIFF
--- a/ios/RNCWebView.m
+++ b/ios/RNCWebView.m
@@ -438,7 +438,11 @@ static NSURLCredential* clientAuthenticationCredential;
         [_webView loadRequest:request];
     }
     else {
-        NSURL* readAccessUrl = _allowingReadAccessToURL ? [NSURL URLWithString:_allowingReadAccessToURL] : request.URL;
+        NSURL* readAccessUrl = request.URL;
+        if(_allowingReadAccessToURL){
+            NSString *accessUrl = [_allowingReadAccessToURL stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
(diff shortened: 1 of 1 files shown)


Titozzz:
Did you make sure it still works for people using file:// ?

hank121314:
Yes, I have tested on a real device for both situations.

hank121314:
Here is an explanation of this PR.

When using RCTConvert to convert NSString to NSURL, it will enter into [RCTConvert.m](https://github.com/facebook/react-native/blob/v0.60.5/React/Base/RCTConvert.m) line 85.
And then it will convert the parameters to NSString name path(line 87~90).
Line 94 uses URLWithString to convert the path to NSURL name URL.
If the URL is nil means that it may contain some special character.
So the program will keep going to line 99 and do stringByAddingPercentEncodingWithAllowedCharacters to get the formatted URL.
I think RCTConvert is doing the same thing as URLString. But it has furthermore protection.

PR #847, opened 2019-09-04
fix(deps): updated dependancies and types

Titozzz:
# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist


cozmo:
For posterity - This PR fixes a bug I described in #850.

sunnylqm:
Unfortunately still get an error

node_modules/react-native-webview/lib/WebViewTypes.d.ts:13:18 - error TS2430: Interface 'RNCWebViewUIManager' incorrectly extends interface 'UIManagerStatic'.
  Types of property 'getViewManagerConfig' are incompatible.
    Type '(name: "RNCWebView") => { Commands: WebViewCommands; }' is not assignable to type '(name: string) => { Commands: { [key: string]: number; }; }'.
      Types of parameters 'name' and 'name' are incompatible.
        Type 'string' is not assignable to type '"RNCWebView"'.

13 export interface RNCWebViewUIManager extends UIManagerStatic {

cozmo:
I'm not getting that error, however trying to call methods changed in `src/WebView.ios.tsx` (maybe android too but haven't tested that) results in: "UIManager.dispatchViewManagerCommand was called with 2 arguments but expects 3 arguments. If you haven't changed this method yourself, this usually means that your versions of the native code and JavaScript code are out of sync. Updating both should make this error go away."




Which conceptually makes sense, I guess RN has strong requirements around argument number? 

I _think_ this means any of the class methods exposed on the component are broken as of this release.

stijnwesterhof:
@cozmo  im getting this error when I call `webview.goBack()`. Any idea how to fix this?

stijnwesterhof:
I made a question on Stackoverflow: https://stackoverflow.com/questions/58027898/webview-goback-throws-error-uimanager-dispatchviewmanagercommand-got-2-argume

cozmo:
@stijnwesterhof This was a bug introduced in this PR (released as version `7.0.7`). As such to avoid the issue use a version before this bug was introduced (aka any version less than `7.0.7`). I just opened #886 with a fix, but not sure the timeline to get it merged/deployed etc.

Issue #575, opened 2019-05-16
Android 9: App crashes if hardware acceleration is enabled


danielang:
**Bug description:**
Our app crashes on Android 9 if the hardware acceleration was used in the webview.
If I disable the hardware acceleration on the embedded webviews the app works fine. 
If I disable the acceleration in the AndroidManifest the app also works without crashes.
So I think it has something to do with the webview implementation and their usage of the hardware acceleration on Android 9.

**To Reproduce:**
The bug occurse in our currently productive app. The repository could be found here: [postillonmedia/app](https://github.com/postillonmedia/app) 

**Expected behavior:**
The app should not crash when the acceleration is enabled 😄 

**Screenshots/Videos/Traces:**
Current reports of the Google-Play console
![image](https://user-images.githubusercontent.com/6752808/57890937-d72c8980-7839-11e9-8d33-f48deb2edacf.png)

One of the traces:
```
05-14 23:07:01.963 10286  7866 22340 W com.postillon: 0xebadde09 skipped times: 0
05-14 23:07:01.964 10286  7866 22340 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x1c in tid 22340 (RenderThread), pid 7866 (com.postillon)
05-14 23:07:02.049 10286 22987 22987 I crash_dump32: obtaining output fd from tombstoned, type: kDebuggerdTombstone
05-14 23:07:02.051  1058  5437  5437 I /system/bin/tombstoned: received crash request for pid 22340
05-14 23:07:02.052 10286 22987 22987 I crash_dump32: performing dump of process 7866 (target tid = 22340)
05-14 23:07:02.065 10286 22987 22987 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
05-14 23:07:02.065 10286 22987 22987 F DEBUG   : Build fingerprint: 'samsung/beyond2lteeea/beyond2:9/PPR1.180610.011/G975FXXU1ASD5:user/release-keys'
05-14 23:07:02.065 10286 22987 22987 F DEBUG   : Revision: '26'
05-14 23:07:02.065 10286 22987 22987 F DEBUG   : ABI: 'arm'
05-14 23:07:02.065 10286 22987 22987 F DEBUG   : pid: 7866, tid: 22340, name: RenderThread  >>> com.postillon <<<
[...]

Issue #40, opened 2018-09-13
[Android][WebView] injectedJavaScript don't work


retyui:
When I [debug `WebView`](https://bamtech.gitbook.io/dev-standards/react-native/debugging/debug-webviews.mo) 
I understand that after serialize (on andoid) from the script, delete all the line breaks, because of what you have to be very careful.

## Example:
```js
// it's work fine
const validScript = `(function(){
alert(123);
})()`;
<WebView injectedJavaScript={validScript} />
```

```js
// It's not work on Android
const invalidScript = `
(function(){
        alert(123); // some comment
})();`;
<WebView injectedJavaScript={invalidScript} />
```
### As a result, the  webview will get this code: (SyntaxError: Unexpected end of input)
```js
(function(){        alert(123); // some comment})();
```

And another problem may arise because of a forgotten semicolon:
```
(function(){
alert(1)
window.postMessage('123')
})();
```
### SyntaxError: Unexpected identifier
```js
(function(){alert(1)window.postMessage('123')})();
```

jamonholmgren:
@retyui Thanks for the report! Do you have suggestions on how this might be fixed?

retyui:
No

Titozzz:
Should be fixed by #41

Issue #39, opened 2018-09-13
Manual Installation 


laukaichung:
`react-native link` doesn't always work on `MainActivity.java` after installing `react-native-navigation`.  

Can you provide documentation for manual installation? I want to make sure there's nothing missing after running `react-native link react-native-webview`.

By the way, I can confirm that [this issue](https://github.com/facebook/react-native/issues/10865) is resolved with this repository on IOS.

Titozzz:
Can you extend what you mean by doesn't always work ?
It would be interesting to know what is you react-native version too, as 0.57 shipped with some gradle changes

stonecold123:
`react-native link` doesn't add `RCTWebViewPackage` to `MainActivity.java`. I have to add it manually. The second version of `react-native-navigation` ([Document](https://wix.github.io/react-native-navigation/v2/#/docs/Installing)) changes a lot of stuff in  `MainActivity.java` so running `react-native link` may not work as expected.

````
import com.reactnativenavigation.NavigationApplication;
import com.reactnativecommunity.webview.RCTWebViewPackage;

public class MainApplication extends NavigationApplication {

      @Override
      public List<ReactPackage> createAdditionalReactPackages() {
          return Arrays.<ReactPackage>asList(
             new RCTWebViewPackage()
          );
      }

}
````

Titozzz:
Oh yeah, of course I remember this from when I was using react native navigation, but that is the same with all the packages, you need to edit MainActivity manually. Please submit an issue with react-native directly to see for a proposal to make the cli compatible with more setups such as react-native-navigation, even tho it seems hard to do.

We will document manual link steps in a near future. If anyone wants to do it before me, feel free :)

jamonholmgren:
@stonecold123 This would be a great opportunity for you to help out the community and document the steps you're taking to get this working! We would really appreciate the help. This is a community-driven repo and everybody pitching in is what will make it awesome. <3

Titozzz:
I've updated https://github.com/react-native-community/react-native-webview/blob/master/docs/Getting-Started.md#2-link-native-dependencies

Issue #38, opened 2018-09-13
WKWebView has no hideKeyboardAccessoryView prop


musicode:
react-native-wkwebview-reborn has the `hideKeyboardAccessoryView` prop, so that i can hide it, it's so useful.

Titozzz:
Feel free to submit a PR, we would  be glad to add more features into this repo, else we will add it to the TODO list when we create it :)

PR #854, opened 2019-09-08
fix(Android): hardware acceleration issue

thanakij:
…sues/575

# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -256,7 +256,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
   @ReactProp(name = "androidHardwareAccelerationDisabled")
   public void setHardwareAccelerationDisabled(WebView view, boolean disabled) {
-    if (disabled) {
+    if (disabled || !view.isHardwareAccelerated()) {
       view.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
     } else {
       view.setLayerType(View.LAYER_TYPE_HARDWARE, null);
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -259,7 +259,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
     if (disabled) {
       view.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
     } else {
-      view.setLayerType(View.LAYER_TYPE_NONE, null);
+      view.setLayerType(View.LAYER_TYPE_HARDWARE, null);
     }
   }
 
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -256,7 +256,10 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
   @ReactProp(name = "androidHardwareAccelerationDisabled")
   public void setHardwareAccelerationDisabled(WebView view, boolean disabled) {
-    if (disabled || !view.isHardwareAccelerated()) {
+    ReactContext reactContext = (ReactContext) view.getContext();
+    final boolean isHardwareAccelerated = (reactContext.getCurrentActivity().getWindow()
+        .getAttributes().flags & WindowManager.LayoutParams.FLAG_HARDWARE_ACCELERATED) != 0;
+    if (disabled || !isHardwareAccelerated) {
       view.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
     } else {
       view.setLayerType(View.LAYER_TYPE_HARDWARE, null);


thanakij:
@Titozzz, this is for fixing #575.

Titozzz:
I'm a bit worried as this changes everyone rendering to default to hardware acceleration.
https://developer.android.com/guide/topics/graphics/hardware-accel#layers

Could other people, including those affected by the issue weight in?

thanakij:
For those who may have been affected, they can always set "androidHardwareAccelerationDisabled" like what have been suggested in [#575](https://github.com/react-native-community/react-native-webview/issues/575#issuecomment-508845949). The problem is rather on the opposite site, which is when the hardware acceleration is enabled. And this is what this pull request is trying to solve.

In addition, [Android doc](https://developer.android.com/guide/topics/graphics/hardware-accel.html) says hardware acceleration is enabled by default. In addition, setting view's layer type to be _LAYER_TYPE_NONE_ sounds suspicious to me because it would imply neither hardware (GPU) nor software is being used.

It is like one would need _LAYER_TYPE_NONE_ only when one does not need any visual effect. And this may cause the issue when incorporating with other navigation libraries like people found (They tend to do some animation when moving screens or pressing "back").

j0x0j:
@thanakij wow, thanks for this fix ... been struggling with this issue for a couple of weeks, now my webview  with `androidHardwareAccelerationDisabled={false}` doesn't crash on android 9!

Titozzz:
@thanakij docs says:
LAYER_TYPE_NONE: The view is rendered normally and is not backed by an off-screen buffer. This is the default behavior.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -259,7 +259,7 @@ public void setHardwareAccelerationDisabled(WebView view, boolean disabled) {
     if (disabled) {
```
dulmandakh: It looks like fix is correct, but need to determine if hardware rendering is available, see https://developer.android.com/guide/topics/graphics/hardware-accel#determining. 

```java
if (disabled || !view.isHardwareAccelerated())
```

thanakij: This is interesting. :)

Let me summarize the possible scenarios:
- _LAYER_TYPE_SOFTWARE_ -> will be fine on most situations, but people find that iframe videos are not playing
- _LAYER_TYPE_NONE_ -> people find that app may crash when a specific set of conditions are met (ie, Android 9? + using some kind of navigation libraries (with visual effect) + using WebView on the screen)
- _LAYER_TYPE_HARDWARE_ -> this would be new (in context of this library) and I agree that we will need to test more

From the [doc](https://developer.android.com/guide/topics/graphics/hardware-accel#view-level), it is like the existing code is already correct for "disabled" case by simply using `myView.setLayerType(View.LAYER_TYPE_SOFTWARE, null)`

Maybe we can make use of _isHardwareAccelerated()_ for the "else" case...
How about below code then?

```
} else {
  if (view.isHardwareAccelerated()) {
    view.setLayerType(View.LAYER_TYPE_HARDWARE, null);
  } else {
    view.setLayerType(View.LAYER_TYPE_NONE, null);
  }
}
```

thanakij: The critical point to decide is on the else=>else case. What should we set when we want to use the hardware (GPU) but it is not available. If it is supposed to fallback to use software, then the code will be what you've suggested earlier.

```
if (disabled || !view.isHardwareAccelerated()) {
  view.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
} else {
  view.setLayerType(View.LAYER_TYPE_HARDWARE, null);
}
```

dulmandakh: how about

```java
if (disabled) {
  view.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
} else {
  view.setLayerType(view.isHardwareAccelerated() ? View.LAYER_TYPE_HARDWARE : View.LAYER_TYPE_NONE, null);
}
```

thanakij: I like it but we have to decide whether we want to get rid of _LAYER_TYPE_NONE_ totally. 

If yes, then

```
if (disabled || !view.isHardwareAccelerated()) {
  view.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
} else {
  view.setLayerType(View.LAYER_TYPE_HARDWARE, null);
}
```

If no, then

```
if (disabled) {
  view.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
} else {
  view.setLayerType(view.isHardwareAccelerated() ? View.LAYER_TYPE_HARDWARE : View.LAYER_TYPE_NONE, null);
}
```

Personally I am convinced on the yes version.

dulmandakh: IIRC, NONE is causing issues, so it might be better to get rid of it.

thanakij:
Hi, I have updated my code to include checking if a view is hardware accelerated.

sapjax:
@thanakij We tested this patch on real Android device, It causes an obvious performance issue, The Webview got stuck when page scrolling, Just same as the `androidHardwareAccelerationDisabled` is set to `true`.
But before this patch, if we don't set androidHardwareAccelerationDisabled, this issue does not exist.

thanakij:
> @thanakij We tested this patch on real Android device, It causes an obvious performance issue, The Webview got stuck when page scrolling, Just same as the `androidHardwareAccelerationDisabled` is set to `true`.
> But before this patch, if we don't set androidHardwareAccelerationDisabled, this issue does not exist.

Just a guess here. Because you said after the patch, now with or without `androidHardwareAccelerationDisabled`, it behaves the same. It hints that `view.isHardwareAccelerated()` is _false_ and both will simply use _LAYER_TYPE_SOFTWARE_ (the result is therefore the performance issue that you have found).

Can you help verify if this is true (that your view is not accelerated)?
- If yes, then can you try again after changing the view to be hardware enabled (you may have some customization to disable this in many places like AndroidManifest.xml)
- If no, then we really need some more work here. As it turns out that we cannot rely on `view.isHardwareAccelerated()`

sapjax:
> > @thanakij We tested this patch on real Android device, It causes an obvious performance issue, The Webview got stuck when page scrolling, Just same as the `androidHardwareAccelerationDisabled` is set to `true`.
> > But before this patch, if we don't set androidHardwareAccelerationDisabled, this issue does not exist.
> 
> Just a guess here. Because you said after the patch, now with or without `androidHardwareAccelerationDisabled`, it behaves the same. It hints that `view.isHardwareAccelerated()` is _false_ and both will simply use _LAYER_TYPE_SOFTWARE_ (the result is therefore the performance issue that you have found).
> 
> Can you help verify if this is true (that your view is not accelerated)?
> 
> * If yes, then can you try again after changing the view to be hardware enabled (you may have some customization to disable this in many places like AndroidManifest.xml)
> * If no, then we really need some more work here. As it turns out that we cannot rely on `view.isHardwareAccelerated()`

@thanakij 
We already set `android:hardwareAccelerated="true"` in AndroidManifest.xml, and didn't found any place we disabled it.

I found the `view.isHardwareAccelerated()` is always return `false`. 
But for the code in `android/view/View.java`:
```java
    public boolean isHardwareAccelerated() {
        return mAttachInfo != null && mAttachInfo.mHardwareAccelerated;
    }
```
the `mAttachInfo.mHardwareAccelerated;` is always `true`.
the problem is when `view.isHardwareAccelerated()` is called, it didn't run into `isHardwareAccelerated()` method in `android/view/View.java` actually.

thanakij:
> > > @thanakij We tested this patch on real Android device, It causes an obvious performance issue, The Webview got stuck when page scrolling, Just same as the `androidHardwareAccelerationDisabled` is set to `true`.
> > > But before this patch, if we don't set androidHardwareAccelerationDisabled, this issue does not exist.
> > 
> > 
> > Just a guess here. Because you said after the patch, now with or without `androidHardwareAccelerationDisabled`, it behaves the same. It hints that `view.isHardwareAccelerated()` is _false_ and both will simply use _LAYER_TYPE_SOFTWARE_ (the result is therefore the performance issue that you have found).
> > Can you help verify if this is true (that your view is not accelerated)?
> > 
> > * If yes, then can you try again after changing the view to be hardware enabled (you may have some customization to disable this in many places like AndroidManifest.xml)
> > * If no, then we really need some more work here. As it turns out that we cannot rely on `view.isHardwareAccelerated()`
> 
> @thanakij
> We already set `android:hardwareAccelerated="true"` in AndroidManifest.xml, and didn't found any place we disabled it.
> 
> I found the `view.isHardwareAccelerated()` is always return `false`.
> But for the code in `android/view/View.java`:
> 
> ```java
>     public boolean isHardwareAccelerated() {
>         return mAttachInfo != null && mAttachInfo.mHardwareAccelerated;
>     }
> ```
> 
> the `mAttachInfo.mHardwareAccelerated;` is always `true`.
> the problem is when `view.isHardwareAccelerated()` is called, it didn't run into `isHardwareAccelerated()` method in `android/view/View.java` actually.

Thank you, @sapjax. My thoughts on this are:

(1) As [some discussion](https://stackoverflow.com/questions/12119753/view-ishardwareaccelerated-is-always-false) shows that `view.isHardwareAccelerated()` will work only after the View is attached to Window. [This](https://stackoverflow.com/questions/6792115/detect-hardware-acceleration-at-runtime-android) suggests an alternative which is to check flag `FLAG_HARDWARE_ACCELERATED` in _ActivityInfo_. I will take a look into this and update the patch.

(2) Another thing is you can actually test this patch quickly by removing ` || !view.isHardwareAccelerated()` part for now while waiting for (1) to be fixed. It should not crash with  _LAYER_TYPE_HARDWARE_ mode being set.

thanakij:
Hi @dulmandakh,

I and @sapjax have found that View's `isHardwareAcelerated()` might not be working as expected. So I update my code to check the flag from current Activity's Window instead. Please have a look for review then.

Next, according to [doc](https://developer.android.com/reference/android/content/pm/ActivityInfo.html#flags), checking flags from _ActivityInfo_ is corresponding to manifest which is supposed for global setting (throughout the app). Hence, this might not be ideal. So I move to a more fine-grain Window level checking instead.

sapjax:
@thanakij 
I tested your new code, It works as expected now.
Thanks for your amazing work!

Titozzz:
👋 this looks good to me, just waiting on a final input from @dulmandakh

KestasVenslauskas:
Did this just came back in 7.4.1?

bcgilliom:
I'm also seeing a crash here from GooglePlay:

```
Caused by: java.lang.NullPointerException: 
  at com.reactnativecommunity.webview.RNCWebViewManager.setHardwareAccelerationDisabled (RNCWebViewManager.java:284)
```

It says the device is `Google Pixel XL (marlin), Android 10`

thanakij:
> Did this just came back in 7.4.1? Experiencing same crashes. Disabling acceleration solves it...
> Also disabling react-navigation animations seems to fix the problem also.

First of all, my thoughts on this are:
- By disabling the hardware acceleration, it is more like trying to hide the problem. Although I am not saying that the fix is perfect (like what the `NullPointerException` has already been reported, which will be addressed independently), but I am convinced that by allowing the hardware mode in WebView (instead of _LAYER_TYPE_NONE_) will bring this WebView to the same standard view type being used consistently by other views in the app. In short, this is the challenge after we allow _LAYER_TYPE_HARDWARE_ for WebView.
Nevertheless, please disabling the hardware acceleration for now if this helps you to prevent the incoming crash reports on your side.
- Next, like what you have found that when you disable the "animation" in react-navigation library, the crash disappeared, so the issue is actually may be on the react-navigation side too (ie, some bugs when animation is enabled).
I will try to go through the library, but it could take some time for me.

Next, did the crash saying something like _"Unable to create layer for RNCWebViewManager$e @bdf4ece, size 840x16416 exceeds max size 16384"_? If yes, then it probably relate to #906.

@KestasVenslauskas Could you help us verify if this is the case for WebView by:
(1) You have some `<ScrollView>` wrapping around `<WebView>`, can you help change <ScrollView> to <View> and see if the crash disappeared.
(2) Next, please get back the `<ScrollView>` and try to replace the (probably large) content from `<WebView>` with plain `<Text>` to see if the same crash still happen without WebView.

The reason for asking this is because I have Googled and found that there is [a similar issue](https://stackoverflow.com/questions/26626344/scene-transition-with-hero-elements-throws-layer-exceeds-max-dimensions-support) in case of native Android when using large texts with animation. It is helpful if you could help verify if `hasOverlappingRendering` or `transitionGroup` resolves your issue?

thanakij:
> I'm also seeing a crash here from GooglePlay:
> 
> ```
> Caused by: java.lang.NullPointerException: 
>   at com.reactnativecommunity.webview.RNCWebViewManager.setHardwareAccelerationDisabled (RNCWebViewManager.java:284)
> ```
> 
> It says the device is `Google Pixel XL (marlin), Android 10`
> 
> I'm on `"react-native-webview": "^7.4.1",`

Thanks for reporting. I will make another pull request for fixing this.

Titozzz:
the default is https://developer.android.com/guide/topics/graphics/hardware-accel#layers NONE iirc, so I'm not sure

thanakij:
> Did this just came back in 7.4.1? Experiencing same crashes. Disabling acceleration solves it...
> Also disabling react-navigation animations seems to fix the problem also.

Hi, can you let me know which version of react-navigation you are using. Also if you are using version > 2, please let me know the versions for react-navigation-stack, react-navigation-tabs and react-navigation-drawer also.

indapublic:
```react-native-webview@7.2.5```

Have this exception in my Sentry

```
JSApplicationIllegalArgumentException
Error while updating property 'androidHardwareAccelerationDisabled' of a view managed by: RNCWebView
```

But my affected Android versions are Android 7.1.1 (SM-T350), Android 8.0.0 (SM-G930F), Android 9 (
ELE-L09). But cannot reproduce it for my Oneplus with Android 9.0.9

Can it be related?

**Comment by thanakij** [comment]
> `react-native-webview@7.2.5`
> 
> Have this exception in my Sentry
> 
> ```
> JSApplicationIllegalArgumentException
> Error while updating property 'androidHardwareAccelerationDisabled' of a view managed by: RNCWebView
> ```
> 
> But my affected Android versions are Android 7.1.1 (SM-T350), Android 8.0.0 (SM-G930F), Android 9 (
> ELE-L09). But cannot reproduce it for my Oneplus with Android 9.0.9
> 
> Can it be related?

Hi, my guess here is that it is pretty much a different issue because it is not really the `NullPointerException` which is fixed in [7.4.3](https://github.com/react-native-community/react-native-webview/releases/tag/v7.4.3). Being said that, please see if this update resolves your issue.

Next, it helps if you can show us how parameters, especially `androidHardwareAccelerationDisabled`, are being passed to `<WebView>`.

**Comment by indapublic** [comment]
Thanks for quick reply! I will update version but unfortunetaly I can't reproduce it so perhaps investigation can take a long time.

```
<WebView 
	source={{
		uri: url,
		headers: {
			referer: getBaseRootUrl(),
		},
	}}
	ref={this.setRef}
	userAgent={config.userAgent}
	javaScriptEnabled
	startInLoadingState
	sharedCookiesEnabled
	thirdPartyCookiesEnabled
	onNavigationStateChange={onNavigationStateChange}
	onError={this.handleError}
	renderError={this.renderError}
	renderLoading={this.renderLoading}
	onMessage={onMessage}
	mixedContentMode="compatibility"
	allowUniversalAccessFromFileURLs
	style={styles.webView}
	injectedJavaScript={javascript}
	onShouldStartLoadWithRequest={this.shouldStartLoadWithRequest}
	allowsInlineMediaPlayback
	mediaPlaybackRequiresUserAction={false}
/>
```

`androidHardwareAccelerationDisabled` not used

**Comment by ItsNoHax** [comment]
@thanakij I am still getting crashes from having hardware acceleration enabled in Webviews that are nested inside ScrollViews

**Comment by thanakij** [comment]
> @thanakij I am still getting crashes from having hardware acceleration enabled in Webviews that are nested inside ScrollViews

@ItsNoHax 
(1) Which version are you using?
(2) What is the URL trying to view?
(3) When you say "hardware acceleration enabled", you simply mean you simply omit "androidHardwareAccelerationDisabled" prop, right?
(4) I guess you are using some kind of navigation library. Can you let me know what are them and the versions also?

**Comment by ItsNoHax** [comment]
Hi @thanakij ,

First of all, the reason I need hardware acceleration is because sometimes the html we render seems to run into a limit and logcat spits out the warning: 

```
2020-01-14 14:59:41.191 6736-6736/nl.basenet.appreactnative W/View: RNCWebView not displayed because it is too large to fit into a software layer (or drawing cache), needs 20599320 bytes, only 17049600 available
```

but when we do enable hardware acceleration it crashes with the following error:

```
020-01-14 14:53:30.294 15359-15401/nl.basenet.appreactnative A/libc: Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x1c in tid 15401 (RenderThread), pid 15359 (.appreactnative)
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG: Build fingerprint: 'google/sdk_gphone_x86/generic_x86:10/QSR1.191030.002/5978551:userdebug/dev-keys'
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG: Revision: '0'
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG: ABI: 'x86'
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG: Timestamp: 2020-01-14 14:53:30+0100
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG: pid: 15359, tid: 15401, name: RenderThread  >>> nl.basenet.appreactnative <<<
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG: uid: 10168
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG: signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x1c
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG: Cause: null pointer dereference
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG:     eax 00000000  ebx f286b9ac  ecx f2863548  edx f3376460
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG:     edi c6a5b560  esi 00000000
2020-01-14 14:53:30.370 19031-19031/? A/DEBUG:     ebp c6a5b458  esp c6a5b440  eip f236e9ba
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG: backtrace:
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #00 pc 002ad9ba  /system/lib/libhwui.so (SkSurface::getCanvas()+26) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #01 pc 0037cd4f  /system/lib/libhwui.so (android::uirenderer::skiapipeline::GLFunctorDrawable::onDraw(SkCanvas*)+2191) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #02 pc 003f4c34  /system/lib/libhwui.so (SkCanvas::onDrawDrawable(SkDrawable*, SkMatrix const*)+612) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #03 pc 00379552  /system/lib/libhwui.so (SkCanvas::drawDrawable(SkDrawable*, SkMatrix const*)+306) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #04 pc 0062770a  /system/lib/libhwui.so (android::uirenderer::$_22::__invoke(void const*, SkCanvas*, SkMatrix const&)+42) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #05 pc 001f8f90  /system/lib/libhwui.so (android::uirenderer::skiapipeline::RenderNodeDrawable::drawContent(SkCanvas*) const+2080) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #06 pc 003f4b8d  /system/lib/libhwui.so (SkCanvas::onDrawDrawable(SkDrawable*, SkMatrix const*)+445) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #07 pc 00379552  /system/lib/libhwui.so (SkCanvas::drawDrawable(SkDrawable*, SkMatrix const*)+306) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #08 pc 0062770a  /system/lib/libhwui.so (android::uirenderer::$_22::__invoke(void const*, SkCanvas*, SkMatrix const&)+42) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #09 pc 001f8f90  /system/lib/libhwui.so (android::uirenderer::skiapipeline::RenderNodeDrawable::drawContent(SkCanvas*) const+2080) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #10 pc 003f4b8d  /system/lib/libhwui.so (SkCanvas::onDrawDrawable(SkDrawable*, SkMatrix const*)+445) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #11 pc 00379552  /system/lib/libhwui.so (SkCanvas::drawDrawable(SkDrawable*, SkMatrix const*)+306) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #12 pc 0062770a  /system/lib/libhwui.so (android::uirenderer::$_22::__invoke(void const*, SkCanvas*, SkMatrix const&)+42) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #13 pc 001f8f90  /system/lib/libhwui.so (android::uirenderer::skiapipeline::RenderNodeDrawable::drawContent(SkCanvas*) const+2080) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #14 pc 003f4b8d  /system/lib/libhwui.so (SkCanvas::onDrawDrawable(SkDrawable*, SkMatrix const*)+445) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #15 pc 00379552  /system/lib/libhwui.so (SkCanvas::drawDrawable(SkDrawable*, SkMatrix const*)+306) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #16 pc 0062770a  /system/lib/libhwui.so (android::uirenderer::$_22::__invoke(void const*, SkCanvas*, SkMatrix const&)+42) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #17 pc 001f8f90  /system/lib/libhwui.so (android::uirenderer::skiapipeline::RenderNodeDrawable::drawContent(SkCanvas*) const+2080) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #18 pc 003f4b8d  /system/lib/libhwui.so (SkCanvas::onDrawDrawable(SkDrawable*, SkMatrix const*)+445) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #19 pc 00379552  /system/lib/libhwui.so (SkCanvas::drawDrawable(SkDrawable*, SkMatrix const*)+306) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #20 pc 0062770a  /system/lib/libhwui.so (android::uirenderer::$_22::__invoke(void const*, SkCanvas*, SkMatrix const&)+42) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #21 pc 001f8f90  /system/lib/libhwui.so (android::uirenderer::skiapipeline::RenderNodeDrawable::drawContent(SkCanvas*) const+2080) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #22 pc 003f4b8d  /system/lib/libhwui.so (SkCanvas::onDrawDrawable(SkDrawable*, SkMatrix const*)+445) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #23 pc 00379552  /system/lib/libhwui.so (SkCanvas::drawDrawable(SkDrawable*, SkMatrix const*)+306) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #24 pc 0062770a  /system/lib/libhwui.so (android::uirenderer::$_22::__invoke(void const*, SkCanvas*, SkMatrix const&)+42) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #25 pc 001f8f90  /system/lib/libhwui.so (android::uirenderer::skiapipeline::RenderNodeDrawable::drawContent(SkCanvas*) const+2080) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #26 pc 003f4b8d  /system/lib/libhwui.so (SkCanvas::onDrawDrawable(SkDrawable*, SkMatrix const*)+445) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #27 pc 00379552  /system/lib/libhwui.so (SkCanvas::drawDrawable(SkDrawable*, SkMatrix const*)+306) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #28 pc 0062770a  /system/lib/libhwui.so (android::uirenderer::$_22::__invoke(void const*, SkCanvas*, SkMatrix const&)+42) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #29 pc 001f8f90  /system/lib/libhwui.so (android::uirenderer::skiapipeline::RenderNodeDrawable::drawContent(SkCanvas*) const+2080) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #30 pc 002584d2  /system/lib/libhwui.so (android::uirenderer::skiapipeline::RenderNodeDrawable::forceDraw(SkCanvas*)+258) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #31 pc 002b02c2  /system/lib/libhwui.so (android::uirenderer::skiapipeline::SkiaPipeline::renderLayersImpl(android::uirenderer::LayerUpdateQueue const&, bool)+626) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #32 pc 002ad40d  /system/lib/libhwui.so (android::uirenderer::skiapipeline::SkiaPipeline::renderFrame(android::uirenderer::LayerUpdateQueue const&, SkRect const&, std::__1::vector<android::sp<android::uirenderer::RenderNode>, std::__1::allocator<android::sp<android::uirenderer::RenderNode>>> const&, bool, android::uirenderer::Rect const&, sk_sp<SkSurface>, SkMatrix const&)+93) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #33 pc 002ad0b0  /system/lib/libhwui.so (android::uirenderer::skiapipeline::SkiaOpenGLPipeline::draw(android::uirenderer::renderthread::Frame const&, SkRect const&, SkRect const&, android::uirenderer::LightGeometry const&, android::uirenderer::LayerUpdateQueue*, android::uirenderer::Rect const&, bool, android::uirenderer::LightInfo const&, std::__1::vector<android::sp<android::uirenderer::RenderNode>, std::__1::allocator<android::sp<android::uirenderer::RenderNode>>> const&, android::uirenderer::FrameInfoVisualizer*)+352) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #34 pc 002ba04c  /system/lib/libhwui.so (android::uirenderer::renderthread::CanvasContext::draw()+332) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #35 pc 002b9098  /system/lib/libhwui.so (_ZNSt3__110__function6__funcIZN7android10uirenderer12renderthread13DrawFrameTask11postAndWaitEvE3$_0NS_9allocatorIS6_EEFvvEEclEv$c303f2d2360db58ed70a2d0ac7ed911b+296) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #36 pc 002cf38f  /system/lib/libhwui.so (android::uirenderer::WorkQueue::process()+255) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #37 pc 002cf0ff  /system/lib/libhwui.so (android::uirenderer::renderthread::RenderThread::threadLoop()+127) (BuildId: 10d1e316b7aa14d35052a448bd944fde)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #38 pc 00014166  /system/lib/libutils.so (android::Thread::_threadLoop(void*)+342) (BuildId: 288ba3aff5b46dbd7e74be954af88b83)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #39 pc 00013aa6  /system/lib/libutils.so (thread_data_t::trampoline(thread_data_t const*)+246) (BuildId: 288ba3aff5b46dbd7e74be954af88b83)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #40 pc 0011a8e5  /apex/com.android.runtime/lib/bionic/libc.so (__pthread_start(void*)+53) (BuildId: 76290498408016ad14f4b98c3ab6c65c)
2020-01-14 14:53:30.394 19031-19031/? A/DEBUG:       #41 pc 000af6a7  /apex/com.android.runtime/lib/bionic/libc.so (__start_thread+71) (BuildId: 76290498408016ad14f4b98c3ab6c65c)
2020-01-14 14:53:30.647 1789-2287/? E/GnssHAL_GnssInterface: gnssSvStatusCb: a: input svInfo.flags is 8
2020-01-14 14:53:30.647 1789-2287/? E/GnssHAL_GnssInterface: gnssSvStatusCb: b: input svInfo.flags is 8
2020-01-14 14:53:30.654 1837-1837/? E//system/bin/tombstoned: Tombstone written to: /data/tombstones/tombstone_47
```

1. Latest (8.0.3)
2. Local html, (sometimes it's extremely basic html when it crashes e.g: <p> This is a paragraph <p>)
3. Correct
4. react-native-navigation 4.0.3

Let me know if you need anything else!

**Comment by thanakij** [comment]
@ItsNoHax, I have done some more research but the log is hinting that it might relate to [something else](https://github.com/invertase/react-native-firebase/issues/2935#issuecomment-559096409) this time.

Can you quickly patch [RNCWebViewManager.java](https://github.com/react-native-community/react-native-webview/blob/master/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L712) with the following:
```
import android.webkit.RenderProcessGoneDetail;
protected boolean onRenderProcessGone(WebView view, RenderProcessGoneDetail detail) {
    return true;
}
```

ItsNoHax:
@thanakij I figured it out. It seems you can't have hardware acceleration and set a height of 0 on a WebView. Our use case was that the webview would load with 0 height and calculate it's the html's height, send it back over the bridge and then we update the webview's height to match it's html content. This used to work fine < RN 0.59.5. Basically we now do the same but initial height set to 1.

ItsNoHax:
So I'm still running into issues with this @thanakij :/

Bigger pieces of HTML are causing it to crash now. (WebView with height 0 also crashes as mentioned in my comment above)

KestasVenslauskas:
This stopped once I removed screens dependency only left react navigation

thanakij:
> So I'm still running into issues with this @thanakij :/
> 
> Bigger pieces of HTML are causing it to crash now. (WebView with height 0 also crashes as mentioned in my comment above)

And stack trace for your latest crash?

Razorholt:
OK, I just tried and played with height and it does work for a while but eventually the app crashes.
The only solution I found is to disable animation in React Navigation (animationEnabled: false). I know it makes the UX crappy - and Android didn't need that - but that's all I could find so far.

gaoyiming:
@ItsNoHax I have same question with you . How do you resolve this problem .
when i close  hardware  , show log "NCWebView not displayed because it is too large to fit into a software layer "
when i open hardware , crash " signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x1c"

PR #865, opened 2019-09-13
Allow custom CA to be used on webview requests

eaceto:
# Summary

This commit allows us to use custom Certificate Authorities (aka self-signed certificates) with WebView without disabling SSL check.
It's also valid for those wanting to implement SSL Pinning over the WebView requests.

I implemented this by modifying webView:didReceiveAuthenticationChallenge:completionHandler in order to look into a Map of hosts/Certificates for the right certificate. If we have the same certificate the server presents, then it's a trusted cert. 

This features impacts only on RNCWebView.

## Test Plan

Create a self-signed certificate and assign it to a website. 
Then in your iOS App copy the DER file of the certificate into the bundle, load it and set it in your AppDelegate. For example:

```
  NSBundle *bundle = [NSBundle mainBundle];
  NSMutableDictionary* certMap = [NSMutableDictionary new];

  // Get Certificate from Bundle file
  {
    NSData *rootCertData = [NSData dataWithContentsOfFile:[bundle pathForResource:@"example.com" ofType:@"der"]];
    SecCertificateRef certificate = SecCertificateCreateWithData(NULL, (CFDataRef) rootCertData);
    OSStatus err = SecItemAdd((CFDictionaryRef) [NSDictionary dictionaryWithObjectsAndKeys:(id) kSecClassCertificate, kSecClass, certificate, kSecValueRef, nil], NULL);
    [certMap setObject:(__bridge id _Nonnull)(certificate) forKey:@"example.com"];
  }
 
  // Set a Map of hosts / certificates
  [RNCWebView setCustomCertificatesForHost:certMap];
```

* Having a SelfSigned Certificate

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


dulmandakh:
@eaceto could you please revert back spacing changes, it would improve readability and makes review easy.

eaceto:
Hello @dulmandakh 

I made the requested changed.

Titozzz:
Would this potentially be a breaking change?

eaceto:
Not at all. This change introduces a new (optional) feature. It doesn’t break the existing API, and introduces a new method that may be called from the AppDelegate in case it’s needed. 
Most apps don’t implement SSL Pinning, but this is a security standard in enterprises.

eaceto:
I will submit a PR for the Readme ASAP.
Thanks

samitha9125:
@Titozzz @eaceto 

Recently I faced an issue with Microsoft login where using Charles and custom certificate, can extract username and the password without any problem by monitoring the traffic even through it used HTTPS. So I'm curious to know, after using this method in iOS, can we stop such sensitive data exposure using SSL Pinning?  Also love to see updated doc soon. Great work guys.

eaceto:
Hello @samitha9125 , excuse me for the delaying in replaying, I was on vacations.
Using SSL Pinning, the custom SSL certificate that Chales Proxy exposes to your (react native) app will be rejected by the app. Leading in not loading the login page.
I'm curious how you implemented the Microsoft Login as it should only work using HTTPS and a valid certificate.

flexaddicted:
@eaceto I'm wondering why you use `SecItemAdd` in this example.

Thanks, Lorenzo

eaceto:
Hello @flexaddicted. In this example is not needed, I just extracted it from a more complex example where then the certificate was used (from the KeyChain). Sorry for that, hope it didn't produce you a problem.

flexaddicted:
@eaceto not at all but it should be removed since unnecessary.

eaceto:
I commented it out in the example. Thanks!!!

romixch:
I wonder why the hostname is used as key here. Now I have to know all the hosts which use my custom certificate authority in advance. Wouldn't it be easier to have a list of trusted authorities and test against those? Just the way a browser does.

theMasix:
Hi @eaceto . Thanks for the immediate response after my previous comment. Any good news?

hpelitebook745G2:
any updates on this @eaceto ? appreciate your time and response.

eaceto:
Hello! I thought about implementing a solution similar to the iOS one, with a fine-grained control, but I found it not useful when you can use Android Network Security Config for the same purpose.

https://developer.android.com/training/articles/security-config

You can add trusted CAs to your app, and perform SSL pinning easily by describing a XML like this:

<?xml version="1.0" encoding="utf-8"?>
<network-security-config>
    <domain-config>
        <domain includeSubdomains="true">secure.example.com</domain>
        <domain includeSubdomains="true">cdn.example.com</domain>
        <trust-anchors>
            <certificates src="@raw/trusted_roots"/>
        </trust-anchors>
    </domain-config>
</network-security-config>

In contrast, iOS does not provide built-in SSL Pinning, but Android does.

PR #868, opened 2019-09-16
fix inconsistent backgroundColor

wsliaw:
# Summary

Fix inconsistent background color when loading html & html loaded.
I implement this by setting background color right after webview init.

## Test Plan

1.
```
<Webview source={html:'<html></html>'} style={{backgroundColor:'transparent'}} />
```
we can see background color is always transparent (before applying this fix it will show a white background when loding)

2.
```
<Webview source={html:'<html></html>'} style={{backgroundColor:'white'}} />
```
set backgroundColor to white or without setting any backgroundColor, the result will be totally the same as current version, which is white.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


Titozzz:
Shouldn't we remove setBackgroundColor from other places if we add it here as it might be redundant?

wsliaw:
> Shouldn't we remove setBackgroundColor from other places if we add it here as it might be redundant?

Agree, removed redundant setBackgroundColor and tested on both device & simulator

brandons:
This fixes the white background issue mentioned here. https://github.com/react-native-community/react-native-webview/pull/587 @Titozzz

kefyru:
how it works? I use <WebView style={{backgroundColor='transparent'.... or any other, but it still white...
in WebView@7.2.3
rn 0.60.5

liquidvisual:
Yeah, this doesn't work.

PR #885, opened 2019-09-20
Http error callback

### LINKED ISSUE #189: I hope listening to the loading fail （500 or 404 or 401 ）

### LINKED ISSUE #807: Check requested page network status and errors
I have a angular panel and I need to change it to web App. In some of pages i have a API call and the server response. Is there any way that I can check the page network call to reload the page if the response have problem ?

hojason117:
# Summary

#### This PR adds an `onHttpError` callback which notifies the host app that an error http status code is received.

After finishing the code I found out there's already a closed PR for this feature but it never get merged, so I figure I'll just mention it here and I've also addressed to most of the comments in that PR.
[Http error callback (Catch http error statuscode)](https://github.com/react-native-community/react-native-webview/pull/558) #558 

It's useful to be able to get the error status code in the webview, so that we can have different error handlings based on the status code (e.g. refresh token on `401`).

Requests for nested frames are filtered out, so `onHttpError` will only be called on the main frame.

The native event returned from this callback will contain:
`statusCode`: The http error status code.
`description`: Description of the error status code (only on `Android`).

Implementation:
`iOS`: [`decidePolicyForNavigationResponse`](https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455643-webview?language=objc)
`Android`: [`onReceivedHttpError`](https://developer.android.com/reference/android/webkit/WebViewClient.html#onReceivedHttpError(android.webkit.WebView,%20android.webkit.WebResourceRequest,%20android.webkit.WebResourceResponse)) (introduced in API level 23)

Fixes #189 
Fixes #807 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅ (only on >= 6.0)    |

## Checklist


Titozzz:
What is the difference between that and onLoadingError ?

hojason117:
@Titozzz I assume you are talking about the `onError` callback. `onError` is called when the webview failed to load the content, usually means there's no internet connection. The error code in the `onError` callback is not an http status code, but something like [this on iOS](https://developer.apple.com/documentation/foundation/1448136-nserror_codes) and [this on Android](https://developer.android.com/reference/android/webkit/WebViewClient.html#constants_2).
On the other hand `onHttpError` catches the http error status code (e.g. 4xx, 5xx).

PR #889, opened 2019-09-23
fix(android scroll): revert #281 that caused new issues

### LINKED ISSUE #878: vertical scroll in ScrollView (with WebView component inside) stopped works after upgraiding webview from version 7.0.5 to 7.0.6 on Android-Platform
vertical scroll in ScrollView (with WebView component inside) stopped works after upgraiding webview from version 7.0.5 to 7.0.6 on Android platform. When webview cover screen at bottom and mostly cover whole screen, we can't scroll via webview to elements located under webview.
It related that in new version 7.0.6 webview has own scroll, which disable scroll inside scrollview.
It was broken in commit a5cb7ba65c65a68ba714272df21fcef16728e56a,
Probably to fix it we need something like scrollEnabled props for android platform (currently it exist only for ios platform)

platform: Android
react-native: 0.60.5
react-native-webview: 7.0.6

Titozzz:
# Summary
Fix #878 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -16,7 +16,6 @@ import android.os.Environment;
 import androidx.annotation.RequiresApi;
 import androidx.core.content.ContextCompat;
 import android.text.TextUtils;
-import android.view.MotionEvent;
 import android.view.Gravity;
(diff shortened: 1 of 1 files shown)


kefyru:
WORKS!!!

Issue #802, opened 2019-08-23
lint failed


vinurs:
> Task :react-native-webview:lint FAILED
Ran lint on variant release: 11 issues found
Ran lint on variant debug: 11 issues found
Wrote HTML report to file:///Users/vinurs/system-configuration/work/glm/GlmApp/node_modules/react-native-webview/android/build/reports/lint-results.html
Wrote XML report to file:///Users/vinurs/system-configuration/work/glm/GlmApp/node_modules/react-native-webview/android/build/reports/lint-results.xml

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':react-native-webview:lint'.
> Lint found errors in the project; aborting build.

  Fix the issues identified by lint, or add the following to your build script to proceed with errors:
  ...
  android {
      lintOptions {
          abortOnError false
      }
  }
  ...

  The first 3 errors (out of 4) were:
  /Users/vinurs/system-configuration/work/glm/GlmApp/node_modules/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java:369: Error: Call requires API level 21 (current min is 16): android.webkit.CookieManager#removeAllCookies [NewApi]
      CookieManager.getInstance().removeAllCookies(null);
                                  ~~~~~~~~~~~~~~~~
  /Users/vinurs/system-configuration/work/glm/GlmApp/node_modules/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java:780: Error: Call requires API level 21 (current min is 16): android.webkit.PermissionRequest#getResources [NewApi]
        String[] requestedResources = request.getResources();
                                              ~~~~~~~~~~~~
  /Users/vinurs/system-configuration/work/glm/GlmApp/node_modules/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java:804: Error: Call requires API level 21 (current min is 16): android.webkit.PermissionRequest#deny [NewApi]
          request.deny();
                  ~~~~

vinurs:
this need to set the min sdk version to 21

TakaKeiji:
> 
> 
> this need to set the min sdk version to 21

Same issue here, i've changed the minSdkVersion to 21, and still getting the error...

TakaKeiji:
> 
> 
> > this need to set the min sdk version to 21
> 
> Same issue here, i've changed the minSdkVersion to 21, and still getting the error...

[...]

PR #894, opened 2019-09-24
fix(android): Update minSdkVersion to 21 to resolve linting issues

silverchen:
# Summary

Some methods used require API level 21 but it's currently set to 16, which causes`:react-native-webview:lint` to fail as reported here #802 
## Test Plan

`:react-native-webview:lint` should run successfully.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅  |

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -29,7 +29,7 @@ android {
   compileSdkVersion getExtOrIntegerDefault('compileSdkVersion')
   buildToolsVersion getExtOrDefault('buildToolsVersion')
   defaultConfig {
-    minSdkVersion 16
+    minSdkVersion 21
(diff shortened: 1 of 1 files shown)


Titozzz:
We need to support 16. We can guard theses method if we want to fix lint instead of updating minVersion. We will drop 16 as soon as RN does

tomtargosz:
@Titozzz when you say "guard these methods" do you mean supplying a `TargetApi` for the methods or adding a version check inside the method?

tomtargosz:
@silverchen @Titozzz I opened #940 to address this issue while maintaining support for 16.

PR #912, opened 2019-09-26
feat(new prop): containerStyle

marcorm:
# Summary

Me and my team need to use the `WebView` to display error from server (in HTML format) inside modals.
The `flex: 1` in the container style doesn't allow us to style the modal as the UX/UI team wants to.

With this `prop` we can set `flex: 0` and style the container and the `WebView` as we want.

## Test Plan

`yarn ci` runs without any issue.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [X] I updated the typed files (TS)

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -25,6 +25,7 @@ This document lays out the current public properties and methods for the React N
 - [`onShouldStartLoadWithRequest`](Reference.md#onshouldstartloadwithrequest)
 - [`startInLoadingState`](Reference.md#startinloadingstate)
 - [`style`](Reference.md#style)
+- [`containerStyle`](Reference.md#containerStyle)
 - [`decelerationRate`](Reference.md#decelerationrate)
 - [`domStorageEnabled`](Reference.md#domstorageenabled)
 - [`javaScriptEnabled`](Reference.md#javascriptenabled)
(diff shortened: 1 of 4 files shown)


Titozzz:
I'm curious to understand what you can't do, as you could just wrap webview inside a view with fixed height and achieve exactly what you need, can you elaborate more ?

marcorm:
Hi @Titozzz thanks for watching this.
You're right. I should elaborate more:
The real "issue" comes in conjunction with react-native-autoheight-webview.

I created a very [small repository to show this problem](https://github.com/marcorm/webviewIssue/tree/master) (I don't know if you've got time to watch into this).
You can uncomment the indicated 3 lines inside the App.js to see what's the desired behaviour.

We were already thinking about wrapping the webview and changing its size accordingly and it's what we've done inside the repo.
But letting us setting `flex: 0` or other styles to `webview container` seemed much clearer and 
more flexible to me, because there's no more need to set the container size when it changes.

Anyway, the solution implemented in the repository (uncommenting 3 lines), could be enough for us. So, if you think this PR is not useful don't worry ;)

marcorm:
Hi @Titozzz,
[here's the repository](https://github.com/marcorm/webviewIssueResolved/tree/master) where I use my fork and get the desired result, without setting height 'manually'.

Issue #942, opened 2019-10-09
onPermissionRequest not called when requesting camera permissions on Android


siddhantbenadikar:
**Bug description:**
While uploading a file, in the startPhotoPickerIntent, the onPermissionRequest listener isn't called when asking for ACTION_IMAGE_CAPTURE and ACTION_VIDEO_CAPTURE actions. As a result, permissions to the camera aren't given thus the chooser doesn't respond to the camera option.

**To Reproduce:**
Add a `<input accepts="image/*" />` field in your DOM and try accessing it from the webview. On clicking the camera action, the chooser closes without navigating to the camera

**Expected behavior:**
Camera permission request should be asked to the user

**Environment:**
 - OS: Android
 - OS version: API 28
 - react-native version: 0.59.10
 - react-native-webview version: 7.0.5

Issue #573, opened 2019-05-16
check camera permission


M45ter:
**Bug description:**
html use camera , in android log error
java.lang.SecurityException: Permission Denial: starting Intent { act=android.media.action.IMAGE_CAPTURE...

**To Reproduce:**
html code like
<input type="file" accept="image/*" capture="camera">

**Expected behavior:**
open camera , take a pic and return pic to html

**Environment:**
 - OS: android 
 - OS version: 9
 - react-native version: 0.59.8
 - react-native-webview version:5.8.1

ddenisen:
Dealing with the same issue here. The problem is that Android dictates that _if_ your app declares the "CAMERA" permission in its manifest, _then_ you must explicitly request camera permission from the user before using the IMAGE_CAPTURE intent, which the web view does. See more details [here](https://stackoverflow.com/questions/32789027/android-m-camera-intent-permission-bug) and [here](https://developer.android.com/reference/android/provider/MediaStore.html#ACTION_IMAGE_CAPTURE).

However, if the web view is the only place where you're relying on the camera, then you can simply remove the camera permission from your manifest and it will just work. Kind of stupid API design, I know.

sakina1403:
@ddenisen I tried removing the camera permissions but it still doesn't work. Like the chooser opens for whether to select camera/gallery, but when I press on Camera, the chooser closes and nothing happens after that.

sakina1403:
It should ask permission at runtime, instead it just denies silently. I can confirm this because when I go to settings and change it from 'Ask' to 'Allow' then the camera opens and I am able to upload image.

lfalkner:
I am seeing the same behavior as @sakina1403 with the following environment.
      react: 16.8.6 => 16.8.6 
      react-native: 0.59.9 => 0.59.9 
and 
        API Levels: 28
        Build Tools: 28.0.3
        System Images: android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom

In addition, I can see that the permissions for the camera are granted in the emulator, when I go to the settings section.

lfalkner:
Previously, my team was using this library:
 https://github.com/andreipfeiffer/react-native-webview-android-file-upload
I'd be open to making a PR with changes included in this library which properly handle permissions for the camera if there is interest

Issue #508, opened 2019-04-17
How to use camera through Webview?


KamilGromov1122:
Hi.

I am a new React Native developer.
Now, I am implementing some function using React Native.

I want to use Camera and Mic through Webview.
My site works well on Mobile Browser, but when I call that url in Webview, I can't use Camera and Mic.

Please help me.

Thanks.

KamilGromov1122:
I can't find any proper answer for this question.

Titozzz:
#231 is maybe related?

hartatovich:
> #231 is maybe related?

i'm also having the same issue...
tried to review #231  as well but didn't find a clear answer about how to solve that 

"react-native-webview": "^5.7.0"

@Titozzz @KamilGromov1122

mharis161:
@hartatovich  did you find any solution yet?

hartatovich:
@mharis161 not yet...
my guess we just need some one to merge a pull request from #231  
but waiting for someone to reply :)

cc - @Titozzz

mharis161:
@hartatovich i commited @Javascript-Ninja's file 'RNCWebViewManager.java' to my code but got this error:

> Task :react-native-webview:compileDebugJavaWithJavac FAILED
D:\Android\workspace\demoApp\node_modules\react-native-webview\android\src\main\java\com\reactnativecommunity\webview\RNCWebViewManager.java:157: error: incompatible types: String cannot be converted to WritableMap
      dispatchEvent(view, new TopShouldStartLoadWithRequestEvent(view.getId(), url));
                                                                               ^
D:\Android\workspace\demoApp\node_modules\react-native-webview\android\src\main\java\com\reactnativecommunity\webview\RNCWebViewManager.java:165: error: incompatible types: String cannot be converted to WritableMap
      dispatchEvent(view, new TopShouldStartLoadWithRequestEvent(view.getId(), request.getUrl().toString()));
                                                                                                        ^
D:\Android\workspace\demoApp\node_modules\react-native-webview\android\src\main\java\com\reactnativecommunity\webview\RNCWebViewManager.java:788: error: cannot find symbol
    return this.aPackage.getModule();
                        ^
  symbol:   method getModule()
  location: variable aPackage of type RNCWebViewPackage
Note: D:\Android\workspace\demoApp\node_modules\react-native-webview\android\src\main\java\com\reactnativecommunity\webview\RNCWebViewManager.java uses or overrides a deprecated API.
[...]

PR #944, opened 2019-10-10
(fix android) camera permissions request on file upload

siddhantbenadikar:
# Summary
Fix for issue #942 related to issues #573 and #508 
Proposed an alternate chooser for the user to select the action on file upload for android in the webview. The issue with the current implementation is that we don't ask the user for Camera permissions if the user selects either ACTION_IMAGE_CAPTURE or ACTION_VIDEO_CAPTURE. This fix would explicitly ask Camera permissions from the user when the user selects to take an image even if they have previously denied it.

Add a `<input accepts="image/*" />` field in your DOM and try accessing it from the webview. 

You should receive a `Take a photo...` option that on selection would ask camera permission from the user. Look at #942 for current behavior

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
@@ -199,33 +199,6 @@ public class RNCWebViewModule extends ReactContextBaseJavaModule implements Acti
     }
   }
 
-//  @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
-//  public boolean startPhotoPickerIntent(final ValueCallback<Uri[]> callback, final Intent intent, final String[] acceptTypes, final boolean allowMultiple) {
-//    filePathCallback = callback;
-//
-//    ArrayList<Parcelable> extraIntents = new ArrayList<>();
-//    if (acceptsImages(acceptTypes)) {
-//      extraIntents.add(getPhotoIntent());
-//    }
-//    if (acceptsVideo(acceptTypes)) {
-//      extraIntents.add(getVideoIntent());
-//    }
-//
-//    Intent fileSelectionIntent = getFileChooserIntent(acceptTypes, allowMultiple);
-//
-//    Intent chooserIntent = new Intent(Intent.ACTION_CHOOSER);
-//    chooserIntent.putExtra(Intent.EXTRA_INTENT, fileSelectionIntent);
-//    chooserIntent.putExtra(Intent.EXTRA_INITIAL_INTENTS, extraIntents.toArray(new Parcelable[]{}));
-//
-//    if (chooserIntent.resolveActivity(getCurrentActivity().getPackageManager()) != null) {
-//      getCurrentActivity().startActivityForResult(chooserIntent, PICKER);
-//    } else {
-//      Log.w("RNCWebViewModule", "there is no Activity to handle this Intent");
-//    }
-//
-//    return true;
-//  }
-
   @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
   public boolean startPhotoPickerIntent(final ValueCallback<Uri[]> callback, final Intent intent, final String[] acceptTypes, final boolean allowMultiple) {
     filePathCallback = callback;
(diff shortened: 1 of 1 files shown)


siddhantbenadikar:
**Note:** If this approach is agreed upon by the maintainers then I will spend time on fixing the `public void startPhotoPickerIntent(ValueCallback<Uri> filePathCallback, String acceptType)` (the function that is meant for API level < Lollipop) to use the new implementation. This is not the final version. Wanted to get it out there so maintainers can agree on this solution.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
```diff
@@ -32,19 +35,29 @@
...
+  private static final int OPEN_CAMERA = 4;
+  private static final int SELECT_FILE = 5;
   private static final int FILE_DOWNLOAD_PERMISSION_REQUEST = 1;
+  private static final int CAMERA_PERMISSION_REQUEST = 2;
+
   final String DEFAULT_MIME_TYPES = "*/*";
+  final String TAKE_PHOTO = "Take a photo…";
+  final String TAKE_VIDEO = "Record a video…";
+  final String CHOOSE_FILE = "Choose an existing file…";
+  final String CANCEL = "Cancel";
```
Titozzz: How would you translate those?

andreipfeiffer:
@siddhantbenadikar the initial implementation I did was exactly like yours.

The problem was we couldn't find an elegant way to translate the hard-coded strings you have in native code (Take a photo, Record a video, etc):

Options:

1. force the consumers to pass the strings that they want to display
2. create a comprehensive dictionary in native code in multiple languages
2.1. do you read the language from the phone?
2.2. do you let the consumer pass the language which should be used?

All these questions, without a clear and easy answer, led us to the current implementation.

siddhantbenadikar:
@andreipfeiffer I agree. We need to get find out how to add internalization here. My only concern with the current implementation is that it does not make a request to the user to grant camera permissions. onPermissionRequest should ideally be handling that but for a lot of devs a breakpoint in that function is never hit. I still haven't found out a reason why that is happening, but my guess would be it's got something to do with the different min sdk versions. More details are mentioned in issue #942.

siddhantbenadikar:
@andreipfeiffer could you think of a solution here. The only way my team and I get around this permission issue is by asking the user for the camera permission on log in. But I would like the permission only to be asked if he decides to use the camera. In our case it is taking a picture for file upload while in the webview

andreipfeiffer:
@siddhantbenadikar I'm not an Android dev, but I have discussed this issue with some Android devs.

If you want to capture an image from the camera in your App, you have 2 choices:

1. implement your own Camera API, using the low-level APIs, where you need to control focus, etc, you need to handle the UI, like buttons, etc ... and ask for permissions, because you App uses the Camera API.

2. delegate using Intents to another App (like the default Camera app). Another application uses the Camera API, thus needing permissions, of course.

We went with the second approach, because it was easier to implement.

Please tell us what the problem is with the current approach?
Why do you need permissions?
What API level(s) do you have issues with?

andreipfeiffer:
@siddhantbenadikar 

1. Do you test this on a simulator? or a real device?
2. What device is it? brand, api level, etc.
3. Is the device rooted?
4. Do you have a default camera app?
5. Can you take pictures with it, as a standalone app?

siddhantbenadikar:
@andreipfeiffer 
Do you test this on a simulator? or a real device?
Both

What device is it? brand, api level, etc.
Pixel 3 XL (Android 9, Api level 28)

Is the device rooted?
No

Do you have a default camera app?
Yes

Can you take pictures with it, as a standalone app?
Yes

andreipfeiffer:
@siddhantbenadikar hmmm, interesting. Just out of curiosity, is it working as expected on a lower API level simulator, like 26/27?

I have a OnePlus 7, also with Android 9, and I just checked an app built with the same package.
I don't have any Camera Permissions on that App, and I'm able to capture the photo.

Can you reproduce the bug with the same device/simulator starting from an empty RN project?

We should determine if this is an issue with:
1. the library itself
2. the library in conjunction with other libraries
3. your project in particular

hslavax:
hi @siddhantbenadikar , may I know why this PR had been closed?

PR #955, opened 2019-10-16
Update Reference.md

torjantai:
incognito prop does not work on android


Titozzz:
Maybe it 's not working but that would be a big because it is implemented https://github.com/react-native-community/react-native-webview/blob/master/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L399

PR #958, opened 2019-10-16
Enable postmessage

khuongTiet:
# Summary
Took inspiration from the original webview implementation on the `react-native` repo before the lean core and move to `react-native-community`

In the current implementation, on the android side, it never exposes the `window.postMessage` to the `window.webkit.messageHandlers.ReactNativeWebView`, so messages that are being posted never get handled. On the iOS side, the implementation was updated to use `window.ReactNativeWebView` as opposed to `window.postMessage` to pass the data through. I'm not sure if this was on purpose? Some clarification could help @Titozzz 

Previously, during a webview with a `window.addEventListener("message", function (event) ...` pattern, the message would not be handled by the library.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅    |

## Checklist

- [X ] I have tested this on a device and a simulator (iPhone 8 physical, iPhone X simulator / Samsung S9 physical )

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -700,6 +700,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
         RNCWebView reactWebView = (RNCWebView) webView;
 
         reactWebView.callInjectedJavaScript();
+        reactWebView.linkBridge();
 
         emitFinishEvent(webView, url);
       }
@@ -1033,6 +1034,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
       if (enabled) {
         addJavascriptInterface(createRNCWebViewBridge(this), JAVASCRIPT_INTERFACE);
+        linkBridge();
       } else {
         removeJavascriptInterface(JAVASCRIPT_INTERFACE);
       }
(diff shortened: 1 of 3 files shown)


Titozzz:
Hello, you are not supposed to use webViewRef.postMessage but instead you should use webViewRef.injectJavascript() to communicate from react-native to the webView.

Maybe see my answer here : https://github.com/react-native-community/react-native-webview/issues/809

If this is ok for you please close this PR

khuongTiet:
@Titozzz , do you mean to update the `postMessage` on the javascript in the webview? I'm currently using this with a certain third party integration that posts a message from their javascript that I am capturing in this webview by listening for a message with `onMessage`.

Titozzz:
Webview is not intercepting postMessage on window directly as it caused issues, if you want this behavior you can inject javascript that will proxy anything from window.postMessage to the new one.

jamonholmgren:
As an example of what @Titozzz is saying, you could do something like this:

```js
const interceptPostMessage = `
var _originalPostMessage = window.postMessage
window.postMessage = function(msg) {
  if (_originalPostMessage) { _originalPostMessage(msg) }
  window.ReactNativeWebView.postMessage(msg)
}
true`;

// in render
<WebView
  injectedJavaScript={interceptPostMessage}
  // ...
/>
```

jamonholmgren:
This guide is helpful to understand how this works: https://github.com/react-native-community/react-native-webview/blob/master/docs/Guide.md#communicating-between-js-and-native

khuongTiet:
Thanks @jamonholmgren!! I'll try it out!

PR #963, opened 2019-10-17
iOS 13 - Cannot open PDF file when Content-type is wrong

eladgel:
Added support for opening pdf files in ios 13

# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |   ❌     |

## Checklist

- [ X ] I have tested this on a device and a simulator


Titozzz:
@jamonholmgren please see my comment in the issue

alishaevn:
@jamonholmgren @Titozzz is there an update on this PR? I'm currently experiencing [this issue](https://github.com/react-native-community/react-native-webview/issues/1072) and I'm wondering if this PR would fix it.

Titozzz:
I'm pasting what I wrote in the issue and is the reason this is not merged:

>This pull request means trusting file extension over mime type, are we sure that is a good idea ?

Titozzz:
Related to: https://developer.apple.com/documentation/ios_ipados_release_notes/ios_13_release_notes?language=objc

> To enhance security, NSURLSession no longer sniffs the MIME type when the server sends Content-Type: application/octet-stream. (7820658)

shirakaba:
I think that inferring the MIME type from the file extension is essentially doing "the best of a bad job". If anything, the Webview should *first* inspect the NSURLRequest's `Content-Type` header, and if it's `application/pdf`, then happy days – we simply load it as a PDF. If not, it's probably not a PDF, except for the ambiguous case where the `Content-Type` header is `application/octet-stream`. Only in *that* case should we attempt to infer from the MIME type from the file extension (and as @Titozzz says, as an option behind a flag). This approach would hopefully catch far more cases of genuine PDFs with incorrect file extensions.

Also: can `pathExtension` be `nil`? e.g. when the URL doesn't end in any file extension at all? It's not clear from [the `pathExtension` docs](https://developer.apple.com/documentation/foundation/nsurl/1410208-pathextension?language=objc).

shirakaba:
Related Android issue: https://github.com/react-native-community/react-native-webview/issues/1255

bova80:
@shirakaba will this be added for android?

shirakaba:
@bova80 if it’s not working that way already (I guess it’s not), then no, nobody’s working on a PR for it for Android right now.

bova80:
@shirakaba I opened an issue, but it's along the same lines, trying to get android to understand links with attribute target="_blank". iOS seems to handle it, android not so much.

Liqiankun:
Not working on iOS12

mananjain149:
Is the issue gonna be resolved by this PR waiting for it to be resolved.

jamonholmgren:
I'd like to get more input from the community on this, especially if someone is familiar with webview security.

mdabees:
Hey is supporting pdf on webview, possible anytime soon?

TheAlmightyBob:
We've been using `react-native-webview` to load PDFs on iOS for a while without issue. I gather this PR is specifically addressing the scenario where the `Content-Type` is not `application/pdf`?

Titozzz:
Exactly @TheAlmightyBob and just taking .pdf files as pdf isn't safe

shirakaba:
> Exactly @TheAlmightyBob and just taking .pdf files as pdf isn't safe

Agreed – it's the server's job to serve the right `Content-Type` response header. It's not the client's job to guess what content is being served to it, and it certainly shouldn't blindly trust the file extension. I think if anyone wants that kind of behaviour, it should be opt-in (or otherwise left for them to implement entirely via their own forks).

PR #965, opened 2019-10-17
Fixing possible NullPointerException in setHardwareAccelerationDisabled

thanakij:
# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -280,13 +280,8 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
   @ReactProp(name = "androidHardwareAccelerationDisabled")
   public void setHardwareAccelerationDisabled(WebView view, boolean disabled) {
-    ReactContext reactContext = (ReactContext) view.getContext();
-    final boolean isHardwareAccelerated = (reactContext.getCurrentActivity().getWindow()
-        .getAttributes().flags & WindowManager.LayoutParams.FLAG_HARDWARE_ACCELERATED) != 0;
-    if (disabled || !isHardwareAccelerated) {
+    if (disabled) {
       view.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
-    } else {
-      view.setLayerType(View.LAYER_TYPE_HARDWARE, null);
     }
   }
 


thanakij:
This fix is trying to fix a crash issue NullPointerException in #854.

@Titozzz @dulmandakh What is your thought on this?

dulmandakh:
I would like to suggest a fix, instead of feature removal. So I would like to look into stack trace if possible, and find solution to it.

thanakij:
> I would like to suggest a fix, instead of feature removal. So I would like to look into stack trace if possible, and find solution to it.

From the outside, it does look really like some kind of feature removal but it has some reasons:

(1) From the doc, _"Hardware acceleration is enabled by default if your Target API level is >=14, but can also be explicitly enabled."_ and because using `react-native` means `API level >= 16`. My conclusion here is that, by default, hardware acceleration is enabled and to explicitly set it on is optional.
However, according to the doc, because using hardware acceleration is not perfect (not all 2D operations are supported), Android allows you to control the enabled/disabled from the `Application` level down to the `View` level.

```
public void setHardwareAccelerationDisabled(WebView view, boolean disabled) {
    ReactContext reactContext = (ReactContext) view.getContext();
    final boolean isHardwareAccelerated = (reactContext.getCurrentActivity().getWindow()
        .getAttributes().flags & WindowManager.LayoutParams.FLAG_HARDWARE_ACCELERATED) != 0;
    if (disabled || !isHardwareAccelerated) {
      view.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
    } else {
      view.setLayerType(View.LAYER_TYPE_HARDWARE, null);
    }
}
```

To me, it just means that the "else" part is optional.

(2) Next, when you think more about it, removing the else part from WebView (`View`) can play nicely with the flow from higher levels (`Window` -> `Activity` -> `Application`) already. That is the default flag is inherited from `Window` and as a result the protection code `isHardwareAccelerated` in the "if" part is unnecessary.

This second point is just my hypothesis, but @KestasVenslauskas in #854 can help confirm this by applying the patch from this pull request. And if the crash still happen for him, it just means that the power for this patch is the same as the current one, but without the `NullPointerException`.

Titozzz:
the default is https://developer.android.com/guide/topics/graphics/hardware-accel#layers NONE iirc, so I'm not sure

thanakij:
> the default is https://developer.android.com/guide/topics/graphics/hardware-accel#layers NONE iirc, so I'm not sure

_"The view is rendered normally and is not backed by an off-screen buffer. This is the default behavior."_

I agree that the above statement from doc creates a lot of confusion but because it is quite clear (compared to quoted unclear statement) that the hardware mode is now turned on by default.
So if we did nothing, logically `View` itself should be safely assumed to be `LAYER_TYPE_HARDWARE`.

So why is the statement is not consistent? Perhaps, one possible interpretation of the sentence is that the default behavior is to not utilize off-screen buffers when this layer type is set. Then it starts to make sense that the statement only applies when the mode is set to be `LAYER_TYPE_NONE`.

It is now hardware enabled by default from `Application` -> `Activity` -> `Window` -> `View`. Developers have a chance to disable it when necessary (ie, found some problems when using hardware acceleration) at different levels. To me, this just confirms that `LAYER_TYPE_NONE` is just a legacy mode that we can totally get rid of and we already did remove it from our current _react-native-webview_'s code base.

Maybe, we should just focus on the other two modes: (1) `LAYER_TYPE_HARDWARE` when we want to explicitly enable the hardware mode for the `View` and (2) `LAYER_TYPE_SOFTWARE` when we want to do the fine-grained control and disable the hardware acceleration for this particular `View`.

WoLewicki:
Why was this change merged in the end? I am a maintainer of `react-native-screens` and there is an issue with the WebView causing crash: https://github.com/software-mansion/react-native-screens/issues/214. It seems to be caused by the fact that when the WebView does not explicitly set the layer type to `hardware`, it has the default `none` option, which causes the crash mentioned in the issue. Can someone elaborate on this?

thanakij:
Hi, @WoLewicki @Titozzz,

There are 3 possible layers but `setHardwareAccelerationDisabled()` is taking "boolean" which has only two possible values. What if we replace this method with something more explicit like `setLayerType()`?

Next, we can discuss **the default value** for this. Maybe it is convenient for many users if we make the default to be `LAYER_TYPE_HARDWARRE` and let the users who encounter an error to `setLayerType(LAYER_TYPE_NONE)`.

Lastly, one thing that makes this difficult in the first place is because we actually tried with different layer types but they ended up with somebody will complain (or report a crash issue):

- `LAYER_TYPE_HARDWARE`: crash if HTML is big and animation enabled
- `LAYER_TYPE_SOFTWARE`: performance degradation
- `LAYER_TYPE_NONE`: crash with react-native-screens

WoLewicki:
I think changing this method to `setLayerType()` is a good solution. About the default option, `LAYER_TYPE_NONE` seems like the most reasonable choice, since it is stated to be the default one on Android (which makes the method name "disable hardware acceleration" misleading IMO). Then the users of e.g. `react-native-screens` could be told that the `LAYER_TYPE_HARDWARE` or `LAYER_TYPE_SOFTWARE` is necessary to use WebView with the library. What do you thing about it?

thanakij:
Then, it is like we have the solution then. Let me create a pull request for this.

thanakij:
Btw, just to confirm here. @Titozzz This is going to be breaking change though.

Titozzz:
I'm not sure why we need to change anything,
None has been noted to cause bug for most use cases, and we decided to enforce for these reasons software or hardware, see the conversation above the (merged) PR

WoLewicki:
Maybe I am doing something wrong, but when I test this code, not setting `disabled` to true, makes the view not change the layer type, and it is of type `none` then, while, if I understand correctly, you suggest that it should be `hardware` then.

thanakij:
Hi, before me, it was `LAYER_TYPE_NONE`. It is actually me who created a pull request [#854](https://github.com/react-native-community/react-native-webview/pull/854/files), changing it to `LAYER_TYPE_HARDWARE`.

However, people still face some issues and yes, it is me again creating this pull request switching it back to `LAYER_TYPE_NONE`.

Eventually, that is why now I am thinking about supporting 3 values. If it is required to stick with this method's name, how about we change from `boolean` (primitive type) to `Boolean` (object), which can support _null_?

WoLewicki:
So why was the method called androidHardwareAccelerationDisabled if the `false` value made it obtain `LAYER_TYPE_NONE`? It is not the same as `LAYER_TYPE_HARDWARE`. Am I missing something?

thanakij:
It was my misunderstanding (as Titozzz suggested above). So I was thinking that we should keep the code simple by removing the "else" case and leave it to the system default. However, it turns out the default is `LAYER_TYPE_NONE`.

I realized later but it already got released. Anyway, it turns out that this version crashes less (compared to the version `LAYER_TYPE_HARDWARE`) for me so I didn't bother back then.

thanakij:
I have found that the base "View" component from react-native is doing something similar.
It has a props called "[renderToHardwareTextureAndroid](https://reactnative.dev/docs/view#rendertohardwaretextureandroid)".

Its implementation in Java is shown below (RN 0.62.2):

Image of Yaktocat

RNCWebViewManager -> SimpleViewManager -> BaseViewManager

So does this mean react-native-webview is having somewhat duplicate code with react-native? @Titozzz @dulmandakh 

_default case:_
react-native = `LAYER_TYPE_NONE`
react-native-webview = `LAYER_TYPE_NONE`

_"true" case_ (renderToHardwareTextureAndroid={true}, androidHardwareAccelerationDisabled={true})
react-native = `LAYER_TYPE_HARDWARE`
react-native-webview = `LAYER_TYPE_SOFTWARE`

Titozzz:
So there is a bug with the current code... 

Interesting that there is already another prop to do that... But can you set software with the other prop?

thanakij:
From the code, I don't think it is possible. `renderToHardwareTextureAndroid` would allow only `LAYER_TYPE_NONE` or `LAYER_TYPE_HARDWARE`.

PR #980, opened 2019-10-24
chore(docs): README.md translated into Brazilian portugues

Rubensmp:
I was using this solution a lot, so I decided to translate it into my language.

# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |

## Checklist


Titozzz:
I would gladly merge that, but I won't be able to maintain it, would that still be a win? Maybe we should specify it somewhere

Rubensmp:
I could try to maintain as I can. 
I was looking to translate other docs. I need to find some time. haha

But if you think its not a win. Its ok!  Glad to try to help anyway! 😄

Best regards

Titozzz:
Let's merge it as an experiment!

We will remove it if it becomes too outdated

PR #986, opened 2019-10-28
Implementing minimumFontSize for Android

tmaly1980:
# Summary

Android WebView enforces a default minimum font size of 8 (see https://developer.android.com/reference/android/webkit/WebSettings#setMinimumFontSize(int)), causing scaling issues when using smaller font sizes. Content that may be deliberately scaled down to fit 100% width of the device may overflow. This property allows for customizing the minimum font size so scaling appears similar as on iOS. 

In cases where grids of content/text are displayed (such as a seating chart), font size may need to be scaled down to fit the entire grid onto one screen width.

Running on Android.

Reference HTML code that uses font sizes < 8. For example:

`
<WebView source={{ html: '<html><body><span style="font-size: 4px;">Font size 4</span></body></html>' }} minimumFontSize={1} />
`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [X] I added the documentation in `Reference.md`

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -232,6 +232,11 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
     view.getSettings().setJavaScriptEnabled(enabled);
   }
 
+  @ReactProp(name = "minimumFontSize")
+  public void setMinimumFontSize(WebView view, int fontSize) {
+    view.getSettings().setMinimumFontSize(fontSize);
+  }
+
   @ReactProp(name = "showsHorizontalScrollIndicator")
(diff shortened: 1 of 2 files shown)


Titozzz:
Please add types, but everything else looks good :)

tmaly1980:
@Titozzz Can you clarify what you mean? I'm just coding it in the style like everything else is coded, I don't see that I'm missing anything semantically.

jamonholmgren:
@tmaly1980 You would need to add the type to the TypeScript interface, somewhere around here:

https://github.com/react-native-community/react-native-webview/blob/5ddd6c6fc277b348d1e6725c76f98ab55ea282c1/src/WebViewTypes.ts#L229

Titozzz:
@tmaly1980 Sorry I did not see your message
@jamonholmgren Thanks for clarifying

PR #987, opened 2019-10-28
Fix ClassCastException

nikhilcn:
# Summary
A ClassCastException occurs when setWebViewClient() is called on Android native.
I access the RN webview on Android native by casting to an Android webview to add custom functionality. Calling setWebViewClient() thereafter caused a ClassCastException.

Running on Android

1. Access the RN webview object on Android
2. Cast to Android WebView
3. Call setWebViewClient()

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -1002,7 +1002,9 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
     @Override
     public void setWebViewClient(WebViewClient client) {
       super.setWebViewClient(client);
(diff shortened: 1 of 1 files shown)


Titozzz:
What happens if you then don't have an mRNCWebViewClient?

nikhilcn:
Since RNCWebViewClient extends Android WebView and a super call exists it should be ok.
Calling setWebViewClient() from Android Native is setting properties on the parent class. 

For ReactNative only development using RNCWebViewClient no change is observed.
For ReactNative + Native Android development RNCWebViewClient provides access to vanilla  WebView class.

marcelo-muniz:
We are running into a ClassCastException, this check fixes it. Can we add this to the main distribution?

PR #989, opened 2019-10-29
Allow the use of WKWebViewConfiguration with RNCWKWebView

franck-nadeau:
# Summary

I needed to be able to configure the WKWebView with a [WKWebViewConfiguration](https://developer.apple.com/documentation/webkit/wkwebviewconfiguration?language=objc), but this was not possible with the current code.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ❌     |

## Checklist


jamonholmgren:
@franck-nadeau Can you give an example of how this might be used?

Is there anywhere in the documentation that could be updated to reflect this usage?

franck-nadeau:
@jamonholmgren Sorry for the delay in my response.

My use of this was for adding a [URLSchemeHandler](https://developer.apple.com/documentation/webkit/wkwebviewconfiguration/2875766-seturlschemehandler?language=objc) to the WebView which I was using in my react-native app.

So within my native class:

````
#import <react-native-webview/RNCWKWebViewManager.h>

@interface RCTCustomWebViewManager : RNCWKWebViewManager
@end

RCTCustomWebViewManager () <RNCWKWebViewDelegate...
...
- (UIView *)view {
  self.webView.defaultConfiguration = [WKWebViewConfiguration new];
  self.customUrlHandler = [CDVLocalUrlHandler new];
  self.customUrlHandler.delegate = self;
  [self.webView.defaultConfiguration setURLSchemeHandler:urlHandler forURLScheme:@"cdvfile"];
...
}
````

As for the documentation, I could add a section to your [guide](https://github.com/react-native-community/react-native-webview/blob/master/docs/Guide.md) for this.

franck-nadeau:
@jamonholmgren I forgot to mention that this is already supported on Android. 

Its use is not fully [documented](https://github.com/react-native-community/react-native-webview/blob/5cc3c965ed3f6848bf11e70c2ebdb7363720081e/docs/Custom-Android.md), as it is a rather advance  use I guess, but I am able to do this inside my native Android class to intercept any url which starts with **cdvfile**:

````
public class CustomWebViewManager extends ReactWebViewManager {

     protected static class CustomWebViewClient extends ReactWebViewClient {

        public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceRequest request)
        {
            String url = request.getUrl().toString();

            if (url.startsWith("cdvfile:"))
....
````

This PR allows me to do the same thing on iOS.

PR #991, opened 2019-10-30
fix(android): redirected URL cannot be loaded normally.

hanlh1:
…eturn true，

# Summary
Since it always returns true in RNCWebViewClient shouldOverrideUrlLoading, URLs that need to be redirected will not be loaded normally. To fix this issue, it needs to return super.shouldOverrideUrlLoading(view, url) instead.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist


zcesur:
This fixes #1150 as well

jackmu95:
Since this PR a `return false;` in `onShouldStartLoadWithRequest()` has no effect at all!
This is a absolutely breaking change and is a misbehavior.

swallville:
> Since this PR a `return false;` in `onShouldStartLoadWithRequest()` has no effect at all!
> This is a absolutely breaking change and is a misbehavior.

In my case the webview is working as should be with `onShouldStartLoadWithRequest()`, but the misbehavior described by @jackmu95  should be checked with other test cases to guarantee that everything is working properly.

hojason117:
> Since this PR a `return false;` in `onShouldStartLoadWithRequest()` has no effect at all!
> This is a absolutely breaking change and is a misbehavior.

I'm seeing the same thing. Returning `false` in `onShouldStartLoadWithRequest` for any arbitrary url doesn't seem to work,  the page still loads.

dvicory:
@hojason117 and @jackmu95 are right. This is a breaking change. As it is currently architected, we must return false from `shouldOverrideUrlLoading` to prevent loading the page, as we can not decide synchronously whether or not to allow the URL to load. We send an event to JS, run our `onShouldStartLoadWithRequest` then based on that do nothing or [manually `loadUrl` if we want to allow the navigation](https://github.com/react-native-community/react-native-webview/blob/d4efa1e7f50e2cae211dbef0319695b95a964d94/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L608-L612).

Eyeballing this change alone means we will actually double load URLs and can never block them - once because we allowed it natively, and another when the JavaScript allows it.

I'll open a revert PR. As for the original bug, #1073, I'm afraid I don't understand it enough to propose a fix there, but I don't see any way of keeping this either.

safaiyeh:
Thanks for the revert @dvicory. This did fix a use of case of mine, ill do more of an investigation

dvicory:
Sure thing @safaiyeh. Do you happen to have a small test case showing the issue?

I was actually thinking about doing some stuff with JSI/TurboModules as a learning exercise as the current asynchronous `shouldOverrideUrlLoading` approach bothers me. It should be possible to have some C++ that calls into JS to delegate the decision and get the result synchronously to return true/false. Not sure if that would help your use case.

PR #999, opened 2019-11-07
Move destructor into invalidate

magicismight:
# Summary

WebView instance won't release properly after React component been unmounted, and the `[NSNotificationCenter defaultCenter]` observers are never been removed.

changes:

* I have moved `[NSNotificationCenter defaultCenter]` observer to didMoveToWindow, it is unnecessary to add those observers in initWithFrame.
* remove dealloc, the `[NSNotificationCenter defaultCenter]` observers should be removed before dealloc, or the `dealloc` will never be called.
* move `removeFromSuperview`'s code into `invalidate`, somehow the `removeFromSuperview` is not been called while the React component is unmounted, `RCTInvalidating` protocol is solid in this case.

## Test Plan

Mount a WebView React component on screen and unmount the React component after a while, check if the WebView instance can be accessed from a Safari remote debugger.

A Safari remote debugger is required, it will display the valid WebViews in simulator.

An unmounted WebView instance can be access by Safari remote debugger.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅   |
| Android |    ❌     |

## Checklist

- [x]  I have tested this on a device and a simulator


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -420,6 +420,16 @@ public void setIncognito(WebView view, boolean enabled) {
...
+    // INFO: ensure all other props has been set
+    view.post(new Runnable() {
+      @Override
+      public void run() {
+        loadSource(view, source);
+      }
+    });
+  }
+
+  private void loadSource(WebView view, @Nullable ReadableMap source) {
```
jamonholmgren: Could you submit this in a different PR, since this PR is about iOS?

safaiyeh:
@magicismight any update on separating this PR?

Issue #988, opened 2019-10-29
Android not loading a url  - inconsistent api


gilador:
**Bug description:**
there is a different api for iOS and Android 
          source={{ url: url }}   //works only on iOS
          source={{ uri: url }}   //works on Android & iOS

**Expected behavior:**
the api for both platform should be consistent   

**Environment:**
 - OS:Android
 - OS version: sdk 23+
 - react-native version: 0.59.9
 - react-native-webview version: 6.0.2

tianweicc:
can you fixed？

1nfility:
+1 - This bug made me debug for over 3 hours until i found this.

Titozzz:
We use react-native default loading capabilities, (same as for images). Using url as a parameter is not documented anywhere, has no benefits, and could be removed anytime, because no one should use it. See https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#source

PR #1002, opened 2019-11-07
Fixed Android's source 'url'

1nfility:
# Android source 'url' fix

I decided to make this pull request, as I struggled with the issue for over 3 hours until i found this issue #988

When loading a url in your source props, both ``uri`` and ``url`` works on iOS, but on android only ``uri`` works. There were no error message telling me that, so i fixed the compatibility issue.

## Compatibility with both ``uri`` and ``url``

Before my Pull Request:

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅✅     |
| Android |    ✅❌   |

With my Pull Request:

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅✅     |
| Android |    ✅✅   |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -421,8 +421,8 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
         view.loadDataWithBaseURL(baseUrl, html, HTML_MIME_TYPE, HTML_ENCODING, null);
         return;
       }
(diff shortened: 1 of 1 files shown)


Titozzz:
We use react-native default loading capabilities, (same as for images). Using url as a parameter is not documented anywhere, has no benefits, and could be removed anytime because no one should use it. See https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#source 

I'd rather not encourage using undocumented APIs.
Please if you feel the docs were not clear about using uri instead of url, submit a pull request! I'd be happy to merge it.

PR #1018, opened 2019-11-12
Focus webview after source changes

jakelazaroff:
# Summary

Implements the suggestion from https://github.com/react-native-community/react-native-webview/issues/366 (in which you would be unable to interact with the webview with a remote on Android TV, because the webview wasn't focused). This fix should focus the webview whenever the `source` prop changes.

This fix is only for Android. I'm not sure whether or not this is an issue on iOS, and I don't have the means to test on an Apple TV. I've tested this on a device (NVIDIA Shield TV).

- Android TV (I'm using an NVIDIA Shield TV, but this probably works with others as well)
- TV remote

Once the webview loads, you should be able to interact with it with the remote (left and right tab through links/inputs as if on a keyboard, "OK" button clicks)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [x] I have tested this on a device ~~and a simulator~~

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -468,6 +468,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
           }
         }
         view.loadUrl(url, headerMap);
+        view.requestFocus();
         return;
       }
     }


Titozzz:
Instead of requesting focus for everyone, you can use requestFocus on the webview ref whenever you feel like you need it. https://github.com/react-native-community/react-native-webview/pull/567

jakelazaroff:
@Tito Oof, I’m sorry, I missed that in the documentation 😔 Thank you for the tip!

PR #1038, opened 2019-11-21
Run the injectedJavaScript on DocumentStart for iOS

SRandazzo:
# Summary

## * What issues does the pull request solve? Please tag them so that they will get automatically closed once the PR is merged
We have been migrating to react-native-webview from [react-native-wkwebview](https://github.com/CRAlpha/react-native-wkwebview) which has been relatively straight forward so far. 

One inconsistency we've discovered is that the `injectedJavaScript` script is called at different times.

- In `react-native-wkwebview`, the `injectedJavaScript` value is passed into the `WKWebView` prior to loading starting. Specifically, `after the document element is created, but before any other content is loaded.` (taken from the Apple documentation here: https://developer.apple.com/documentation/webkit/wkuserscriptinjectiontime/wkuserscriptinjectiontimeatdocumentstart?language=objc)

- In `react-native-webview`, the `injectedJavaScript` value is passed into the `WKWebView` after loading has finished. This is effectively identical to a consumer utilizing the `onLoadEnd` prop to call `webview.injectJavaScript(...)` 

I'm not seeking this change to match consistency, but rather found that react-native-webview does not allow us to inject any JS *prior* to the content loading. 

This has left us unable to do any configuration prior to the content being loaded.
Our specific use case is that we are passing values in `localStorage` that our web app needs in order to be configured properly. 

## * How did you implement the solution?
In iOS only, the WKWebView will be initialized with an additional WKUserScript *if* the user has passed in the `injectedJavaScript` prop.
Additionally, I have removed the `injectedJavaScript` from being run after the WebView has finished loading. 

To fill the need of executing code after loading finished, consumers of the library should use the `onLoadEnd` prop which achieves the same result.

## * What areas of the library does it impact?
The iOS codebase only is affected.

## Test Plan

I have tested this by running the example in the readme/api reference section for `injectedJavaScript` and verified that the alert still pops up as expected. 
https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#injectedjavascript

An iOS project

Follow the example in the API reference:
https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#injectedjavascript

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ❌     |

## Checklist

**I cannot find CHANGELOG.md in the source**
- [n/a] I updated the typed files (TS and Flow)


jamonholmgren:
@SRandazzo Thanks, this looks really good.

However, this could cause problems as I'm sure a lot of code is relying on the content being loaded in this property. I wouldn't want to break backwards compatibility in what is probably a very subtle and hard to debug way.

One option would be to introduce another property that conforms to the way you expect it to work. It makes sense that there would be a way to inject JS before the content is loaded. Perhaps something like `injectedJavaScriptBeforeContentLoaded` to make it very explicit.

SRandazzo:
@jamonholmgren thanks for taking a look!

This definitely could end up being a breaking change depending on how people are using it.

I'm definitely open to adding a new prop. The name you suggested sounds more clear in it's behavior which is a plus. 

On the flip side, the way RNWebView currently executes `injectedJavaScript` can already be achieved by calling `webRef.injectJavaScript` from within the `onLoadEnd` prop - so the current implementation of `injectedJavaScript` is convenient but redundant.

All that being said, I appreciate your time looking at this PR and am happy to go either way with this PR so we can get it moved into master! I will update this PR with the `injectedJavaScriptBeforeContentLoaded`

SRandazzo:
@jamonholmgren Pushed up changes as suggested! Thanks for taking a look

congnt171:
Hi
I try install the new react-native-webview 7.6.0 and try on snack
`import React, { Component, useState, useEffect } from 'react';

export default () => {
  const [js, setJs] = useState(`window.addEventListener('load', () => {
	window.ReactNativeWebView.postMessage('loaded');
});
window.ReactNativeWebView.postMessage('load');
`);
  const onMessage = ({ nativeEvent: { data } }) => {
    console.log('request', data);
  };
  const url = 'https://www.google.com/';
  return (
    <WebView
      source={{
        uri: url,
      }}
      injectedJavaScript={js}
      injectedJavaScriptBeforeContentLoaded={`window.ReactNativeWebView.postMessage('load before');`}
      onMessage={onMessage}
      userAgent="Mozilla/5.0 (iPhone; CPU iPhone OS 12_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148"
      javaScriptEnabled
      useWebKit
      sendCookies
      style={{ flex: 1 }}
    />
  );
};
`
But only load catch by onMessage. load before and loaded not running!

SRandazzo:
Hi @congnt171, is this being run on an iOS device or Android?

congnt171:
@SRandazzo i'm testing on iOS device and android simulator.

SRandazzo:
I believe this a race condition going on here with how WKWebView injects the scripts that make this functionality possible. 

On the native side, we inject two scripts at startup:
- One that configures `window.ReactNtaiveWebView` 
- If supplied, the `injectedJavaScriptBeforeContentLoaded` script

It's up to WKWebView when and how the two scripts are injected, so if the `injectedJavaScriptBeforeContentLoaded` is setup first, `window.ReactNativeWebView` will not have been setup yet. 

We should update the docs to recommend against using `window.ReactNativeWebView.postMessage` in the `injectedJavaScriptBeforeContentLoaded` prop since `window.reactNativeWebView` may not have been initialized yet on the native side.


`injectedJavaScriptBeforeContentLoaded` can be used for anything else 

If you have a more specific use case in mind I can help you figure out how to achieve it given the current limitations

congnt171:
I try injectedJavascript on webview `window.onload = () => {something eles}`
I try many time but it not running.
How can i do this with both ios and android.
Thanks!

sholzmayer:
Sounds cool - was looking exactly for that!

But actually doesn't have any effect for me: 

```
      <WebView
        source={{uri}}
        injectedJavaScriptBeforeContentLoaded={`(function() {
          window.alert("Hi there");
      })();`}
        javaScriptEnabled={true}
      />
```

fungilation:
This is awesome. Thanks for this PR @SRandazzo! With this, I'm able to process partially loaded google search results in [my app](https://medium.com/wonderswipe/rethink-mobile-search-10-100x-faster-introducing-wonderswipe-6f2ff0d0e667) before `onload` fires, making first results available much sooner over slow mobile networks.

Would love to see the same `injectedJavaScriptBeforeContentLoaded` implemented on Android also.

congnt171:
@SRandazzo I try using injectedJavaScriptBeforeContentLoaded but nothing run.
I test on iPhoneX and simulator.
When i change injectedJavaScriptBeforeContentLoaded to injectedJavaScript it run.
Thanks!
My snack example: https://snack.expo.io/@louisnguyen171/webview

innocces:
this api is not work for android

SRandazzo:
Happy Holidays 🎉 https://github.com/react-native-community/react-native-webview/pull/1099

I do not have prior experience with Android/Java but gave it a shot and it works in my app. 

If you are familiar with Android please have a look at the PR above and provide feedback, thanks!

SRandazzo:
Hey @sholzmayer, please make sure you are on the latest version and rebuild the native app as this has native code changes. I have just verified using your window.alert example that this works as expected. 

image

SRandazzo:
Please see https://github.com/react-native-community/react-native-webview/pull/1166 for a bugfix update related to this feature

PR #1044, opened 2019-11-26
add allowFileAccessFromFileURLs property on iOS

salockhart:
# Summary

Similar to the work done in https://github.com/react-native-community/react-native-webview/pull/831, it is not possible to render a local HTML file on iOS. For this to be possible `allowFileAccessFromFileURLs` has to be set to true.

This PR adds a new property, and sets it on the `WKPreferences` object before the webview is created.

- Attempt to load a local HTML file that in turn loads other files in the bundle (JS, etc)

- Attempt to load a local HTML file that in turn loads other files in the bundle (JS, etc)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅ in this PR     |
| Android |    ✅ previously    |

## Checklist

-- Is there a CHANGELOG?
-- Is there an example project?

### DIFF
--- a/ios/RNCWebViewManager.m
+++ b/ios/RNCWebViewManager.m
@@ -128,6 +128,10 @@ RCT_CUSTOM_VIEW_PROPERTY(keyboardDisplayRequiresUserAction, BOOL, RNCWebView) {
   view.keyboardDisplayRequiresUserAction = json == nil ? true : [RCTConvert BOOL: json];
 }
 
+RCT_CUSTOM_VIEW_PROPERTY(allowFileAccessFromFileURLs, BOOL, RNCWebView) {
+    view.allowFileAccessFromFileURLs = json == nil ? false : [RCTConvert BOOL: json];
+}
+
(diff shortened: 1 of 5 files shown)


ios/RNCWebViewManager.m
```diff
@@ -128,6 +128,10 @@ - (UIView *)view
   view.keyboardDisplayRequiresUserAction = json == nil ? true : [RCTConvert BOOL: json];
 }
 
+RCT_CUSTOM_VIEW_PROPERTY(allowFileAccessFromFileURLs, BOOL, RNCWebView) {
+    view.allowFileAccessFromFileURLs = json == nil ? false : [RCTConvert BOOL: json];
```
jamonholmgren: This appears to be a private API and could cause apps to be rejected.

https://stackoverflow.com/a/44365081/204044

Any thoughts on that?

salockhart: I’ve used this in the past with production applications, and I’ve not had an issue with apps being rejected. 

However - you’re absolutely right, and I hadn’t considered that when I put up the PR. If this isn’t a risk we want to introduce to the project, I completely understand and we can close this out.

PR #1078, opened 2019-12-12
Two tweaks to how URLs from un-whitelisted origins are handled

aarondail:
# Summary

This PR does two pretty small and related things (in two commits):

## 1. Preventing an Unhandled Promise Rejection
When the WebView tries to load a URL from an un-whitelisted origin, react-native-webview calls `Linking.openURL` and lets the OS handle it.  But, sometimes there is an error on the native side of `Linking.openURL`, for example on Android if the URL is like `intent://....` but there is no Activity that can handle that intent installed.  This results in an unhandled promise rejection because no code is `catch`ing the error. 

We track and handle unhandled promise rejections in our app, and send them to our crash/analytics service as errors.  This is not _really_ an error though, so its just noise.  Actually another concern is that it possibly has user info (in the URL) that we don't want to be storing either.  

Anyways, the simple fix is to add a `.catch(() => {})` when we call `Linking.openURL` so errors, if any, are simply ignored and not sent to our unhandled promise rejection handler.

## 2. Prevent both the library and the app from handling un-whitelisted URLs

Another thing about the scenario described above, is that react-native-webview will call `Linking.openURL` and then _still_ call our `onShouldStartLoadWithRequest` handler.  The result of  `onShouldStartLoadWithRequest` then overrides whether or not the URL should get loaded.

This results in a scenario where a URL from an un-whitelisted origin is handled by the OS (via the call to `Linking.openURL`) *AND* loaded into the WebView because the custom `onShouldStartLoadWithRequest` returns `true`.

The app can work around this by doing the whitelist origin checks itself in `onShouldStartLoadWithRequest` but that feels like a worse solution because every client that uses `onShouldStartLoadWithRequest` would have to do it or have this subtle bug (albeit its definitely an edge-case).

Anyways, the fix is to just say that if a URL is from an un-whitelisted origin, _dont_ call `onShouldStartLoadWithRequest`.  Just call `Linking.openURL` and let the OS handle it, and _dont_ load it in the WebView.

This might be more controversial since it is technically a breaking change (and I had to update a test for it).

## Test Plan

For the first thing just exercising loading a non-whitelisted URL in the WebView that causes an OS error (e.g. the `intent://...` style URLs on Android).  Without this change you will get an unhandled promise rejection error.  With it you won't.

I did that and verified there was no error.

For the second thing, also implement a `onShouldStartLoadWithRequest` function that gets passed to the WebView and that returns true for everything.  Without this change following a URL to an un-whitelisted origin, e.g. un-whitelist `http://` and follow a link to a `http://` URL, will result in the OS opening the system browser for the URL as well as it getting loaded in the WebView.  With this change it will just get opened by the OS in the system browser.

I did that as well, and verified that the URL is just handed off to the OS and not `onShouldStartLoadWithRequest`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [sorta] I have tested this on a device and a simulator // I tested on an Android simulator
- [no] I added the documentation in `README.md` // Not sure it needs to be documented?  Happy to update it though.
- [not yet] I mentioned this change in `CHANGELOG.md`// Wanted to make sure this seemed ok first 
- [no] I added a sample use of the API in the example project (`example/App.js`) // Didn't seem necessary.

### DIFF
--- a/src/WebViewShared.tsx
+++ b/src/WebViewShared.tsx
@@ -44,7 +44,15 @@ const createOnShouldStartLoadWithRequest = (
     const { url, lockIdentifier } = nativeEvent;
 
     if (!passesWhitelist(compileWhitelist(originWhitelist), url)) {
-      Linking.openURL(url);
+      Linking.canOpenURL(url).then((supported) => {
+        if (supported) {
+          return Linking.openURL(url);
+        }
+        console.warn(`Can't open url: ${url}`);
+        return undefined;
+      }).catch(e => {
+        console.warn('Error opening URL: ', e);
+      });
       shouldStart = false;
     }
 
--- a/src/WebViewShared.tsx
+++ b/src/WebViewShared.tsx
@@ -54,9 +54,7 @@ const createOnShouldStartLoadWithRequest = (
         console.warn('Error opening URL: ', e);
       });
       shouldStart = false;
-    }
-
-    if (onShouldStartLoadWithRequest) {
+    } else if (onShouldStartLoadWithRequest) {
       shouldStart = onShouldStartLoadWithRequest(nativeEvent);
     }
 
--- a/src/__tests__/WebViewShared-test.js
+++ b/src/__tests__/WebViewShared-test.js
@@ -102,7 +102,9 @@ describe('WebViewShared', () => {
       await flushPromises();
 
       expect(Linking.openURL).toHaveBeenLastCalledWith('invalid://example.com/');
-      expect(loadRequest).toHaveBeenLastCalledWith(true, 'invalid://example.com/', 1);
+      // We don't expect the URL to have been loaded in the WebView because it
+      // is not in the origin whitelist
+      expect(loadRequest).toHaveBeenLastCalledWith(false, 'invalid://example.com/', 1);
     });
 
     test('loadRequest with false onShouldStartLoadWithRequest override is called', async () => {
--- a/src/__tests__/WebViewShared-test.js
+++ b/src/__tests__/WebViewShared-test.js
@@ -5,6 +5,31 @@ import {
   createOnShouldStartLoadWithRequest,
 } from '../WebViewShared';
 
+Linking.openURL.mockResolvedValue(undefined);
+Linking.canOpenURL.mockResolvedValue(true);
+
+// The tests that call createOnShouldStartLoadWithRequest will cause a promise
+// to get kicked off (by calling the mocked `Linking.canOpenURL`) that the tests
+// _need_ to get run to completion _before_ doing any `expect`ing. The reason
+// is: once that promise is resolved another function should get run which will
+// call `Linking.openURL`, and we want to test that.
+//
+// Normally we would probably do something like `await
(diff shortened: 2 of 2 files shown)


src/WebViewShared.tsx
```diff
@@ -44,11 +44,9 @@ const createOnShouldStartLoadWithRequest = (
     const { url, lockIdentifier } = nativeEvent;
 
     if (!passesWhitelist(compileWhitelist(originWhitelist), url)) {
-      Linking.openURL(url);
+      Linking.openURL(url).catch(() => {});
```
safaiyeh: Would this be cleaner with `async/await`?

Not a fan of silent errors, should at least warn in dev mode what happened.

aarondail: Re: `async/await` I don't _think_ so because the function this is in is not `async` currently (or returning a promise), and we probably(?) don't want to block the rest of the function (calling `loadRequest`) on `await Linking.openURL`... that said I am not 100% sure and am not familiar with the code in any detail so I could totally be wrong.

Re: silent errors, yeah that makes sense.  I uh, borrowed this "pattern" from a markdown library that handles `Linking.openURL` errors this way (eating it) :) I see a handful of places in this library where we call `console.warn`.  Would just doing that be sufficient (e.g. `.catch(e => console.warn('Error calling Linking.openURL', e))`)?

safaiyeh: Sorry about that, for some reason I thought I read `reject` 😂 ignore `async/await`.

So there is a way to pre-check if the app can open a url: https://facebook.github.io/react-native/docs/linking#opening-external-links, this might be useful.

`console.warn` looks good to me, that triggers the YellowBox.

aarondail: Good find, I updated the code to use `canOpenURL` before calling `openURL.   This required some changes to the tests because they need to wait for the Promise returned by `canOpenURL` to resolve.  I added a pretty good sized comment to the tests hopefully explaining that.  Hopefully that is not too gross.

I also added the code to call `console.warn` on errors instead of eating them.  I wrapped it in a `if (__DEV__)` check so they don't get run on release builds, though its probably not a big deal.  I had to updated `.eslintrc` to make `ESLint` ok with using `__DEV__`.  I think that is standard practice but I am happy to revert if it seems wrong.

safaiyeh: Thanks for all this work!

I would say remove `__DEV__`, that should be developers who use this package responsibility to configure their metro configs to handle.

With that change, LGTM. Will ping the team to check it out

aarondail: OK updated.  Thanks for the feedback.

BTW for my education can you tell me what you mean by `configure their metro configs to handle.`?  Is there an option to auto remove calls to the `console.*` functions?

safaiyeh: Yup! Thanks for the change.

Check this out:
https://facebook.github.io/react-native/docs/performance.html#using-consolelog-statements

PR #1096, opened 2019-12-26
fix:iOS WKWebView RetainCycle

sunzhongliangde:
# Fix iOS WKWebView retain cycle

 

When close the webview, dealloc function does not execute, there have a retain cycle in RNCWebView.m file

NO
Hit berakpoint in dealloc function，after the webview close，this function is not excute
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RNCWebView.h
+++ b/ios/RNCWebView.h
@@ -19,6 +19,11 @@
 
 @end
 
+@interface RNCWeakScriptMessageDelegate : NSObject<WKScriptMessageHandler>
+@property (nonatomic, weak) id<WKScriptMessageHandler> scriptDelegate;
+- (instancetype)initWithDelegate:(id<WKScriptMessageHandler>)scriptDelegate;
+@end
+
 @interface RNCWebView : RCTView
 
 @property (nonatomic, weak) id<RNCWebViewDelegate> _Nullable delegate;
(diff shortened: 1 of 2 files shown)


binhndicts:
@Titozzz @sunzhongliangde I don't understand why the PR was merged.

The retain `self` problem is because we `addScriptMessageHandler` for `HistoryShimName` but forget to `removeScriptMessageHandlerForName:HistoryShimName` in `removeFromSuperview`.
Just add this line 
```
[_webView.configuration.userContentController removeScriptMessageHandlerForName:MessageHandlerName];
```
in `removeFromSuperview` will be fine!

(I already tried to test and confirm it works)

sunzhongliangde:
@binhndicts 
Calling `removeScriptMessageHandlerForName:HistoryShimName` method can solve。
but this method is not a good solution, because the `removeScriptMessageHandlerForName` method does not have the right time to call.
In the `ViewWillDisappear` method，If you call `removeScriptMessageHandlerForName` method, you can't use these interaction methods on other pages

binhndicts:
> In the ViewWillDisappear method

I don't understand why do you mention this method. What I said is adding `removeScriptMessageHandlerForName:HistoryShimName` in `removeFromSuperview`. It's not good?

sunzhongliangde:
Sorry, I forget it, I will pull a merge request to remove it.

PR #1099, opened 2019-12-27
Add support for injectedJavaScriptBeforeContentLoaded on Android

SRandazzo:
# Summary
This is the Android counterpart to https://github.com/react-native-community/react-native-webview/pull/1038

This is my first time writing Java/Android so any feedback is much appreciated.

## Test Plan

I am running an iOS + Android ReactNative application and have verified that the injected JS behaves the same on both platforms

An Android RN app with a WebView and some JS to inject

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| Android |    ✅      |

## Checklist


brickglobal:
Good job bro. I found my issuse from expo using build-in old version of react-native-webview. 
Thanks!

congnt171:
Up vote. I found my issue because expo using old version react-native-webview. 
Thanks!

SRandazzo:
Found a bug in this - will update when it's resolved!

SRandazzo:
Ok bug is resolved. I could use some extra eyes on whether or not placing the execution of the code inside of `onPageStarted` is the correct approach.

I originally gated the execution with a boolean flag so it would only run "once", but that caused it to not execute at the proper time

Titozzz:
Issue is that won't work on android > N :( https://developer.android.com/reference/android/webkit/WebView.html#evaluateJavascript(java.lang.String,%20android.webkit.ValueCallback%3Cjava.lang.String%3E)

SRandazzo:
Thanks @Titozzz  - I admit I don't fully understand here as I am new to Android. I tried doing some reading but could you elaborate for my knowledge?

Is this the part of the docs you are referring to:
```
Compatibility note. Applications targeting Build.VERSION_CODES.N or later, JavaScript state from an empty WebView is no longer persisted across navigations like loadUrl(java.lang.String). For example, global variables and functions defined before calling loadUrl(java.lang.String) will not exist in the loaded page. Applications should use addJavascriptInterface(Object, String) instead to persist JavaScript objects across navigations.
```

congnt171:
You can refer to this here: https://github.com/brunobar79/react-native-webview/commit/8cb4e9f463592db7cb1452b2f88fe7555f9c32de
I think this is helpful for you.
Thanks!

Titozzz:
> Is this the part of the docs you are referring to
Yes exactly this says that it won't work :/

shirakaba:
> Yes exactly this says that it won't work :/

I'm not reading it like that. It just says that persisted state won't survive a navigation (just as, if you evaluate any JS in a browser's web inspector that mutates the `window`, any state imparted upon the `window` would be lost upon refreshing the page).

This is fine for the use-case of `injectedJavaScriptBeforeContentLoad` – our script would be able to store state into the page upon `onPageStarted` (which I believe is after the navigation has been committed to), and that state would only be lost upon navigating away from the page. To be honest, that's the behaviour I'd expect and desire, and I'm sure that it's consistent with the behaviour of iOS WKUserScripts.

One way to characterise the behaviour: Use my Test Plan in https://github.com/react-native-community/react-native-webview/pull/1119. Injecting into all frames is out-of-scope for this PR, but you can do the test concerning injecting into just the main frame.

shirakaba:
> Ok bug is resolved. I could use some extra eyes on whether or not placing the execution of the code inside of onPageStarted is the correct approach.

Docs for [onPageStarted](https://developer.android.com/reference/android/webkit/WebViewClient#onPageStarted(android.webkit.WebView,%20java.lang.String,%20android.graphics.Bitmap)):

> Notify the host application that a page has started loading. This method is called once for each main frame load so a page with iframes or framesets will call onPageStarted one time for the main frame. This also means that onPageStarted will not be called when the contents of an embedded frame changes, i.e. clicking a link whose target is an iframe, it will also not be called for fragment navigations (navigations to #fragment_id).

So this is indeed the appropriate lifecycle event for `injectedJavaScriptBeforeContentLoaded`, but only if injecting into the main frame.

Unfortunately, there are no lifecycle methods that mention that they are called for frames other than the main frame 🙁 I can't find any mention online of any Android WebView native methods for injecting JS into non-main frames (only JSObjects), and there's no reliable and secure way to work around that lack of functionality.

congnt171:
Any update ?, i hope this feauture soon avaiable.

safaiyeh:
@SRandazzo Could you update the docs with this prop and make sure to note the capability issue with versions >= Android N?

safaiyeh:
@congnt171 I'm aiming to merge this in with @shirakaba's iOS PR https://github.com/react-native-community/react-native-webview/pull/1119 to have parity between iOS + Android for `injectedJavaScriptBeforeContentLoaded`

shirakaba:
@muthuspark has an interesting workaround for injecting JS into iframes. It may only work post-page-load, i.e. `injectedJavaScript` (and is therefore irrelevant to this present PR, which concerns pre-page-load – `injectedJavaScriptBeforeContentLoaded`), but it's interesting to learn of and would be worth investigating for a follow-up PR:

https://github.com/react-native-community/react-native-webview/issues/824

> For android, I am overriding the `public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceRequest request)` method to intercept my HTML and add the required Javascript into it.

SRandazzo:
Thanks all for the helpful insight!
@safaiyeh i've updated the docs with a note on Android compatibility for this API

congnt171:
Most injectJavascript needs to be done before it is fully loaded to get the best results. So I think this is a problem that should be carefully reviewed and quickly included in order to improve.
Thanks @SRandazzo

shirakaba:
I suppose that injecting JS after content loading could be achieved crudely by listening for the document “loaded” event firing from before the content has loaded. We’d best keep that in mind as an option.

congnt171:
> I suppose that injecting JS after content loading could be achieved crudely by listening for the document “loaded” event firing from before the content has loaded. We’d best keep that in mind as an option.

I think that's not a good idea. Doing so we can not be sure "loaded inject javascript" or not. This will result in heterogeneous results between each download.

shirakaba:
> I think that's not a good idea. Doing so we can not be sure "loaded inject javascript" or not. This will result in heterogeneous results between each download.

I'm not sure the results would be different between each download; could you explain your concerns?

Incidentally, it looks like there are two lifecycle hooks we could use.

> The `DOMContentLoaded` event fires when the initial HTML document has been completely loaded and parsed, without waiting for stylesheets, images, and subframes to finish loading.

[`DOMContentLoaded`](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event)

> The `load` event is fired when the whole page has loaded, including all dependent resources such as stylesheets and images. This is in contrast to `DOMContentLoaded`, which is fired as soon as the page DOM has been loaded, without waiting for resources to finish loading.

[`load`](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event)

On iOS, [`WKUserScriptInjectionTimeAtDocumentEnd`](https://developer.apple.com/documentation/webkit/wkuserscriptinjectiontime/wkuserscriptinjectiontimeatdocumentend?language=objc), which is how [`react-native-wkwebview`](https://github.com/CRAlpha/react-native-wkwebview) implements `injectedJavaScript` (and how my current PR is implementing it) has the following meaning:

> Inject the script after the document finishes loading, but before other subresources finish loading.

If I'm understanding correctly, WebKit injects `WKUserScriptInjectionTimeAtDocumentEnd` scripts at the [`FrameLoader::finishedParsing()` point of the codebase](https://github.com/WebKit/webkit/blob/622d995699a96cafea1023c171471ba766c06361/Source/WebCore/loader/FrameLoader.cpp#L772). Both that and the above description seem to point to 
the `DOMContentLoaded` being the more appropriate lifecycle event of the two.

If we injected specifically upon the `DOMContentLoaded` event, do you still think that would be a bad idea?

shirakaba:
Now that Microsoft have introduced an examples app, I've forked this PR to [ shirakaba/updatable-user-scripts-android](https://github.com/shirakaba/react-native-webview/tree/shirakaba/updatable-user-scripts-android) and have added an example to test injection:

## Partial test pass

Only a partial pass because Android can only inject into the main frame at present. Options for injecting into all frames are limited, but [I've raised one possible approach](https://github.com/react-native-community/react-native-webview/pull/1099#issuecomment-574685371) to try to implement it.

image

## Test fail

Of concern: upon hot reload, injection sometimes works, but sometimes doesn't. This possibly indicates that the android implementation isn't handling props changes well. Or a race condition with loading. Hard to tell.

image

---

[Screenshots](https://github.com/react-native-community/react-native-webview/pull/1119#issue-359340245) of the expected behaviour for a full test pass are visible in my related PR for iOS.

shirakaba:
I played around with the  `shouldInterceptRequest` API, but sadly I don't see any way it (nor any other available `android.webkit.WebViewClient` API) could be used to inject on all frames. I think Android is fundamentally limited to single-frame JS injection.

@safaiyeh if @SRandazzo updated this branch to merge the latest changes, would anything else be needed to merge this?

Note: equally, if @SRandazzo is busy, I could open a PR from [`shirakaba/updatable-user-scripts-android`](https://github.com/shirakaba/react-native-webview/tree/shirakaba/updatable-user-scripts-android), which – as I have failed to add multi-frame support – is just a branch that has merged our two PRs together and is up-to-date with master.

SRandazzo:
Thank you all for pushing this along! This has been some A++ OSS support 


I've updated the branch with master 

The unit test results from @shirakaba are really insightful - thanks for taking the time to do those.  For the potential hot reloading issue is that something we should try to tackle in this PR or as a followup?

SRandazzo:
Also - I am going to go ahead and grant @shirakaba write permissions to this branch/PR since there is a lot of great discussion. 

@shirakaba feel free to push commits to this branch and move it along - I think you have more context/knowledge on this than I do!

safaiyeh:
@SRandazzo & @shirakaba thanks for the team work :D, Let me know when it is ready.

shirakaba:
@SRandazzo Thanks for giving me push access!

I've brought the branch into feature parity with that of my iOS PR.

> For the potential hot reloading issue is that something we should try to tackle in this PR or as a followup?

I'm a bit concerned actually that – although the test failure does happen to reproduce on hot reload – actually the same test failure occurs even from a fresh build, where there's no hot reloading involved. It fails about 75% of the time (see [video](https://gfycat.com/windyuntimelykrill)), and I don't know how to get a browser debugger attached to the Android simulator to look into it...

While the implementation looks perfect on paper, I can't explain this behaviour just at first glance. :(

shirakaba:
I've made the tests more granular and have investigated things a bit more.

* injectedJavaScript works just fine.
* injectedJavaScriptBeforeContentLoaded works about 25% of the time; raising to 100% of the time [if you add in a breakpoint to try to investigate it](https://twitter.com/LinguaBrowse/status/1230621793016254464?s=20).

Does the webview start loading before it's received all its props or something? 🤔

shirakaba:
I've exhausted all avenues I can think of to debug this, at this point – maybe, given its weird breakpoint behaviour, it works better on a real device than on a virtual one. Unfortunately, I don't own an Android device to test this hypothesis.

All I can say to the maintainers @safaiyeh is:

* the code looks perfect
* all the expected code-paths are taken
* the `onLoadStart` lifecycle method is always reliably called
* the `injectedJSBeforeContentLoaded` prop is always set by the time that `callInjectedJavaScriptBeforeContentLoaded()` is invoked
* adding a callback handler into `evaluateJavaScript()` shows that the `injectedJSBeforeContentLoaded` script is indeed run to completion and returns an evaluated value even in visibly failing test cases

I think the implementation is as good as can be and that the reliability problems are possibly down to bugs in `android.webkit.webview`, or my virtual device. The best I can suggest is that users be wary that this API may have device-specific reliability issues.

jamonholmgren:
@shirakaba Thanks so much for your extensive documentation of your exhaustive research into this problem. You've done pretty much all you can and more than we'd ask for. Hopefully someone from the community can find a solution to this problem soon. :heart:

safaiyeh:
This is great @shirakaba. Hot reload issues happen to me as well and I would be okay with ignoring them as it does not affect production.

I love your write ups with everything involved gives me awesome context. Mind finding a home of what you documented somewhere where people who use injectedJavaScriptBeforeContentLoaded can understand what the limitations are.

shirakaba:
@jamonholmgren @safaiyeh thank you both very much for the kind words! 

> Hot reload issues happen to me as well and I would be okay with ignoring them as it does not affect production.

I agree there – However, the issue of `injectedJavaScriptBeforeContentLoaded` failing also reproduces without hot reload involved, namely when switching tabs on the test app to initialise the WebView for the first time. So I can't in good confidence call it "production-ready", unfortunately.

> Mind finding a home of what you documented somewhere where people who use `injectedJavaScriptBeforeContentLoaded` can understand what the limitations are.

I'd be happy to write up a dedicated docs page on JS injection (or add to the existing one on WebView <-> RN communication) to characterise it at length.

# Where to go from here

## Further debugging the issue

### What's the underlying problem

By all measures, the injection does run even upon test failure. And upon test failure, it certainly evaluates the script in *a* JSContext – we just don't see any evidence of it influencing the JSContext belonging to the window we were expecting to load.

The pattern would be explained by `window` (for some reason) being overwritten/discarded shortly after `onPageStarted` fires. I wonder vaguely whether there's some odd treatment of `window` when loading in a static HTML string (as we're doing now) rather than visiting a real hosted webpage.

### Benefits of running the test upon a real hosted webpage 

I note that I'm currently loading the test as a static HTML string, whose URL is officially `about:blank` – it would be worth pointing it at a real hosted webpage instead to compare.

Using the Chrome debugger on the Android device, you can execute `location.reload()` to observe the console logs and the final state of the page (and importantly, this doesn't involve re-instantiating the WebView to repeat the test). As our page is currently loaded from a static HTML string, however, calling `location.reload()` simply revisits `about:blank`, throwing away your static webpage, and breaking the test. So by hosting the test on a real webpage, we could try a little investigation in this manner.

### Why it's worth characterising the test behaviour in response to `location.reload()`

This would allow us to see whether `injectedJavaScriptBeforeContentLoaded` is merely unreliable soon after instantiation, or consistently unreliable on each and every page load.

### The significance of unreliability only shortly after WebView instantiation

It shows that we need to pay attention to the instantiation process (something that I admit I haven't looked into).

Note also that I've been re-running the test either by hot-reloading or by switching test tabs – this may result in some unexpected off-screen rendering magic happening (the iOS implementation has some).

## Whether to merge as-is or not

### This current PR #1099

Apart from the last mile of solving the reliability problem, I think this PR is technically solid – if no solution is found, it would nonetheless be a shame for it to go stale and fall out of sync if the master branch keeps moving ahead.

It may be worth merging it in just with a warning in the documentation saying that it's currently "provisional support" for Android and will be regarded as "experimental" until a follow-up PR gives confidence of stability.

Or alternatively the implementation could be merged in but with no prop for Android exposed in the typings until the reliability aspect is solved. 😇

Either way, I'd be happy to document the understood behaviour in detail (best to do so after characterising with `location.reload()` as mentioned above).

### Its related PR for iOS, #1119 

My iOS PR is pending this Android PR being merged, however, it concerns rather different aspects: multi-frame injection and prop-updating. If this Android PR is going to need a bit more work before being merged, could we try to push the iOS one over the line in advance? It'd be ready once I've cherry-picked in the tests that I wrote for this Android PR.

safaiyeh:
@shirakaba Let's get the iOS over the line first, to support the multiple frames. I am okay with having this partially in if the documentation is clear and people can easily understand the limitations of the Android implementation.

safaiyeh:
One more to go @shirakaba could you resolve the conflicts and I'll test out the Android as well.

shirakaba:
# iOS behaviour

Injection into all frames works both before and after content loaded.

image


# Android behaviour

Injection only works into the top frame. And even then, it only works on afterContentLoaded. Not very exciting.

image

## Last revisit of Android bug

### ✅ `beforeContentLoaded` works for static HTML websites

I quickly uncommented the static HTML in the top of the test to confirm this, and used that as the source instead of a real website.

image

image


### ❌ `beforeContentLoaded` doesn't work for real websites

As planned, the injection tests now load from a genuine web page rather than a static HTML snippet, allowing me to test without the interference of hot loading or the recreating of the WebView, simply via `location.reload()`.

I have confirmed that:

1. The `onPageStarted()` lifecycle event runs `reactWebView.callInjectedJavaScriptBeforeContentLoaded()`, and then runs the `injectedJS` script.
2. The `onPageFinished` lifecycle event runs `reactWebView.callInjectedJavaScript()`, and then runs the `injectedJSBeforeContentLoaded` script.
3. The `WebView.android.tsx` is passing the correct props on to the correspondingly named Java props.
4. The `webView` instance beforeContentLoaded is the same one as afterContentLoaded.

And yet, what I find, is that, for a real website, the `beforeContentLoaded` script doesn't run at all (it logs nothing in the console).

image

... That is, until you turn on "Preserve log" and find that it actually populates some console logs but evidently for a previous(?) page that has since been navigated away from. Annoyingly, if you try to visit the script (`debugger:///VM221:2`), it is empty and you can't set breakpoints on it for next time.

image

So I think that although this PR is correct, there may be some other part of the React Native WebView implementation that is causing the page to be thrown away on early load if it's a real hosted website (or something). I'm not familiar enough with the Android WebView's lifecycle to get any further on this.

---

I have documented the limitations of `injectedJavaScriptBeforeContentLoaded` in the API docs, so I think this is ready for merge; a follow-up PR can try to address what's going wrong now that I have some tests in place.

This PR furthermore brings the docs up to date to clarify what sort of injection is possible in Android vs. iOS (i.e. only main frame injection; and if `beforeContentLoaded`, then only for static HTML sources), and puts in the infrastructure for future PRs related to JS injection.

Ready for review @safaiyeh !

shirakaba:
Just spotted this issue, which might be the cause of the failure (again, out of scope for this current PR): https://github.com/react-native-community/react-native-webview/issues/1250#issue-577271485

Hugo-Meiring:
@shirakaba thanks! Can confirm your PR changes fixed race conditions on our side.

safaiyeh:
Will be testing out this today, hopefully we can land <3

siphiuel:
@shirakaba I think we've ran into similar racing conditions (failing to properly inject in ~50% cases) when using `onPageStarted` for before-load JS injection (details in https://github.com/status-im/status-react/pull/10273).

Using `shouldInterceptRequest` for JS injection solved the problem for us. `shouldInterceptRequest` then inserts a `<script ...>` tag into the page input stream once `<head>` tag is found (see this [commit](https://github.com/status-im/react-native-webview/commit/5578e63fc609b770408ddbdead5a4a211c45218b)).

neerajbuccam:
@SRandazzo Thanks for this useful PR, and code works as well. 

@safaiyeh @jamonholmgren @Titozzz When can we expect this to be merged?

congnt171:
Any update ?

sapjax:
Any update ? 😲

jamonholmgren:
On behalf of the RN WebView maintainers, thanks everyone for your patience and all your help! I'm excited to finally ship this!

SRandazzo:
Wow this is great! Thank you all for your work on this, big thanks to @shirakaba who made huge contributions that made this possible. Open source win!

panda0603:
injectedJavaScriptBeforeContentLoaded is now available for android now, am I correct? 
API reference still says it is only available for iOS, macOS only.

sapjax:
This PR works on some Android device, but don't work on some other device.

image

shirakaba:
@sapjax I believe the JS injection is implemented correctly (this PR works perfectly for HTML pages created as a string but faces problems for real hosted web pages, as I’ve noted in the documentation and in this thread). I think that the issue you’re seeing is the unsolved issue regarding lifecycle race conditions, which I hope someone can solve in a follow-up PR.
See: https://github.com/react-native-community/react-native-webview/issues/1250

shirakaba:
@sapjax I’ve just checked the Chromium issue that you linked. This WebView is based on WebKit, but the pattern that they mention there – namely of erasing the JS Context stores on Window – may also match what I was finding in my debug above. So rather than a race condition, it could also be explained by a WebKit bug of the context being cleared. Unfortunately, I wasn’t able to find any solution myself. I think I noted that attaching a debugger actually caused the problem to disappear, which made it impossible for me to dig any deeper. 😕

sapjax:
@shirakaba 
You are right, It's not a race condition,  It caused by erased JS Context after `onLoadStart`,  On my Android device,  I can recurrence this problem with attaching debugger every time, But on the android simulator and other Android devices, there's no problem.

KingAmo:
@sapjax  @shirakaba in my case, `injectedJavaScriptBeforeContentLoaded ` does not work anyway, even attaching a debugger. i test it on  a HuaWei phone with EMUI 10.0.0

kilmas:
I have the same problem，Most of the time, injectedJavaScriptBeforeContentLoaded doesn't want the effect on IOS. 

> @shirakaba I think we've ran into similar racing conditions (failing to properly inject in ~50% cases) when using `onPageStarted` for before-load JS injection (details in [status-im/status-react#10273](https://github.com/status-im/status-react/pull/10273)).
> 
> Using `shouldInterceptRequest` for JS injection solved the problem for us. `shouldInterceptRequest` then inserts a `<script ...>` tag into the page input stream once `<head>` tag is found (see this [commit](https://github.com/status-im/react-native-webview/commit/5578e63fc609b770408ddbdead5a4a211c45218b)).

PR #1118, opened 2020-01-05
Extract wkWebViewConfig setup into setUpWkWebViewConfig function

shirakaba:
# Summary

I intend to implement updatable user scripts (i.e. user scripts such as injected JS would be updated upon props change), and this will involve removing and re-adding all scripts on the `wkWebViewConfiguration`.

A necessary first step is to extract all the config setup from `didMoveToWindow()` into a separate function. This is beneficial anyway as `didMoveToWindow()` is enormous and hard to analyse at the moment due to all this logic.

This purely a function extraction and will not impact the library.

## Test Plan

None; minor change.

None; minor change.

None; minor change.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    N/A     |

## Checklist


safaiyeh:
@shirakaba Looks good to me!

PR #1119, opened 2020-01-05
feat(iOS): WKUserScripts (e.g. injectedJavaScript) can now update upon props change; and can be configured to inject into all frames.

### LINKED ISSUE #824: How to enable forMainFrameOnly option in iOS webview
I am using react-native-webview to load a URL which has multiple iframes. I am injecting javascript into the loaded website using the `injectedJavaScript` method. The code is injected only in the top frames and not in the iframes.

For android, I am overriding the `public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceRequest request)` method to intercept my HTML and add the required Javascript into it. 

For iOS, after some google search I found some results pointing to setting `forMainFrameOnly` to NO. I am a complete noob to Objective C and have absolutely no idea how to set this. I had a look at RNCWKWenView.m file which had `forMainFrameOnly:YES` at 2 places. I replaced YES with NO and then rebuilt my iOS project in xcode. But still not able to get my javascripts in iframes.

**To Reproduce:**
App.js file

```

import React, { Component } from 'react';
import { BackHandler, Platform } from 'react-native';

const WEBVIEW_REF = 'webview';

    webView = {
        canGoBack: false,
        ref: null
    }

    onAndroidBackPress = () => {
        if (this.webView.canGoBack) {
            this.refs[WEBVIEW_REF].goBack();
            return true;
        }
        return false;
    }

    /**
     * generate the javascript needed to be embedded inside the react native code. 
     */
    generateJSCode() {

        var jsCode = "console.log('my script shows');";

        return jsCode;
    }

    componentDidMount() {
        if (Platform.OS === 'android') {
            BackHandler.addEventListener('hardwareBackPress', this.onAndroidBackPress);
        }
    }

    componentWillUnmount() {
        if (Platform.OS === 'android') {
            BackHandler.removeEventListener('hardwareBackPress');
        }
    }

    render() {

        const uri = "http://localhost:8000";

        return (
            <WebView
              ref={WEBVIEW_REF}
              source={{uri}}
              useWebkit={true}
              injectedJavaScript={this.generateJSCode()}
              javaScriptEnabled={true}
              onNavigationStateChange={(navState) => { this.webView.canGoBack = navState.canGoBack; }}
            />
        );
    }
}
```

**The HTML file:**
```
<!DOCTYPE html>
<html>
<head>
	<title>test</title>
</head>
<body>
	<h1>this is a sample page</h1>
	<iframe src="https://wikipedia.org/" style="width:100%;min-height: 400px;"></iframe>
	<iframe src="https://wikipedia.org/" style="width:100%;min-height: 400px;"></iframe>
</body>
</html>
```

**Expected behaviour:**

On iOS: When I set `forMainFrameOnly:YES` as `forMainFrameOnly:NO`, I am expecting the script to get injected in my iframes as well.

**Environment:**
 - OS: MacOS
 - OS version:
 - react-native version: 0.60.5
 - react-native-webview version: 6.9.0

shirakaba:
Closes #824 (which was closed only due to inactivity)

# Summary

Note: depends upon the minor PR #1118.

* In `react-native-wkwebview`, it is possible to update `injectedJavascript` and `injectedJavaScriptBeforeContentLoad` upon props change. But in `react-native-webview`, it is committed upon mount (as they are only initialised upon `didMoveToWindow`.
* Equally, `react-native-wkwebview` supports injecting JS into the main frame or all frames while `react-native-webview` injects only into the main frame. This means that the WebView has no influence over iframes, for example.

I implemented the solution by:

1. extracting the WKWebViewConfiguration initialisation routine from `didMoveToWindow` into a function called `setUpWkWebViewConfig()` (depends upon PR #1118), and;

2. extracting the WKUserConfig initialisation code from there into a function called `resetupScripts()`.

`resetupScripts()` removes all user scripts (the HTML5 history shim; the cookie-sharing shim; the `injectedJavaScript` source; the `injectedJavaScriptBeforeContentLoaded` source; and the (postMessage) messaging source), then restores all which still apply, given the latest props.

`resetupScripts()` is called upon `didMoveToWindow` (the post-initialisation life-cycle method) and upon any change to the user scripts – be the change to the script itself or the frames to which they apply.

## Test Plan

```tsx
<WebView
  source={{ uri: "https://birchlabs.co.uk/linguabrowse/infopages/obsol/iframeTest.html" }}
  /* Must be populated in order for `messagingEnabled` to be `true` to activate the
   * JS injection user scripts, consistent with current behaviour. This is undesirable,
   * so I'll address it in a follow-up PR. */
  onMessage={(e) => {}}
  
  injectedJavaScriptBeforeContentLoadedForMainFrameOnly={true}
  /* We set this property in each frame */
  injectedJavaScriptBeforeContentLoaded={`window.self.colourToUse = "orange";`}
  
  injectedJavaScriptForMainFrameOnly={true}
  /* We read the colourToUse property in each frame to recolour each frame */
  injectedJavaScript={`window.self.document.body.style.backgroundColor = window.self.colourToUse;`}
/>
```

This should change each frame to have a background colour of orange upon loading. There are three frames here: yellow, pink, and – just out of curiosity – https://www.eBay.co.uk (which cannot be displayed because its X-Frame-Options is set to `SAMEORIGIN`). The eBay frame seems orange post-injection only because it has a transparent background; it blocks script injection by the user content controllers, and this is expected behaviour.

## The page without any injected JS

image

## The page after injected JS (main frame only)

image

## The page after injected JS (all frames)

image

Make assertions based on the background colour of a site with iframes successfully being changed, given the above props.

N/A (feature)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [ ] I have tested this on a device
- [x] I have tested this on a a simulator


shirakaba:
I had to make a few extra bugfix commits and a few edits to the text and screenshots of the initial post, but I think I'm satisfied with my edits now so will stop changing it.

safaiyeh:
@shirakaba Would this need to be implemented for Android as well?

shirakaba:
@safaiyeh This would not technically need to be implemented for Android as I have only changed the props for the iOS NativeWebView component. I would of course eventually like to have this feature for Android, but I will be focusing on the iOS side of my app for the time being.

Android is rather far behind:

1. Although `CommonNativeWebViewProps` and `WebViewSharedProps` includes `injectedJavaScriptBeforeContentLoaded`, `injectedJavaScriptBeforeContentLoaded` is not implemented at all for Android to begin with.

2. `injectedJavaScript` has only a very basic implementation for Android. During the `onPageFinished` function, [`reactWebView.callInjectedJavaScript()` is called](https://github.com/react-native-community/react-native-webview/blob/master/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L725). The current implementation only acts on the main frame and does not consider other frames. I don't know whether Android has an equivalent to the WKUserScript APIs for injecting into all frames at certain page load times. If Android WebView doesn't have a hook for frame load start and end, then things may get tricky to implement. Does anyone know whether it has these hooks?

However, on Android the `injectedJavaScript` does at least update on each prop change.

safaiyeh:
@shirakaba Thanks for the updates. We usually like to wait for features to support both platforms. Since it is not possible on Android from your research, I'll review this PR sometime today.

docs/Reference.md
```diff
@@ -174,6 +176,30 @@ const INJECTED_JAVASCRIPT = `(function() {
 
 ---
 
+### `injectedJavaScriptForMainFrameOnly`
+
+If `true` (default), loads the `injectedJavaScript` only into the main frame.
```
safaiyeh: It may be a bit better to rename to something that says "I am enabling the injected JavaScript on the other frames"

`injectedJavaScript` is always on the  main frame. 

Something like `injectedJavaScriptOnAllFrames` default false. Adding the feature to all frames is an opt-in rather than an opt-out of the main frame only.

safaiyeh:
@shirakaba Partial review, could you resolve conflicts so the diff is easier to read?

shirakaba:
> @shirakaba Partial review, could you resolve conflicts so the diff is easier to read?

Done.

docs/Reference.md
shirakaba: I chose that naming for consistency with `react-native-wkwebview`:

https://github.com/CRAlpha/react-native-wkwebview/blob/master/WKWebView.ios.js#L206-L215

But I agree that it's not a great name. The opposite of "ForMainFrameOnly" is not necessarily "for all frames"; logically it could also be "for more than one frame".

`injectedJavaScriptOnAllFrames` is problematic too, because the opposite of that is not necessarily "just the main frame". The name also sounds like it may take a string (your JS script) rather than a boolean.

I think it's hard to find a property name for which a boolean value makes total sense.

On the other hand, `forMainFrameOnly` is exactly the wording used by WKUserScript, which furthermore takes a boolean:

https://developer.apple.com/documentation/webkit/wkuserscript/1537856-formainframeonly

> A Boolean value indicating whether the script should be injected only into the main frame (YES) or into all frames (NO).

What do you think?

safaiyeh: Sounds good! Naming is always hard, but since it is consistent with iOS I'm fine with it!

safaiyeh:
@shirakaba Looks good to me, would you consider this a breaking change or just a feature?

shirakaba:
@safaiyeh A breaking change, because:

## Breaking changes

1. Users may previously have accidentally been relying on the behaviour that any props updates to `injectedJavaScript` remained immutable after first being set.
2. This section: https://github.com/react-native-community/react-native-webview/pull/1119/files#diff-e2c6890fad2eec1786f059cac2d6d839L1012-L1019 `injectedJavaScript` no longer attaches a `jsEvaluationValue` property to the `onLoadingFinish` event.

## Retaining the behaviour of collecting a JS evaluation value after `injectedJavaScript`

I would say that this could still be achieved by manually calling `injectJavaScript()` upon the `onLoadedFinish` event, but apparently that [no longer returns an evaluation result for security reasons](https://github.com/react-native-community/react-native-webview/blob/master/src/WebView.ios.tsx#L154-L159) (which admittedly I don't understand). It's quite possible that the `jsEvaluationValue` recorded upon that the `evaluateJS()` call that this PR removes was actually a security risk for the same reason and is therefore worth removing. Either way, it did not handle evaluation errors gracefully.

So instead, the best way to retain this behaviour would be to collect the evaluation result of your JS by ensuring that your `injectedJavaScript` source string posts the result that it has evaluated:

```jsx
<WebView
  source={{ uri: "https://birchlabs.co.uk/linguabrowse/infopages/obsol/iframeTest.html" }}
  /* Must be populated in order for `messagingEnabled` to be `true` to activate the
   * JS injection user scripts, consistent with current behaviour. */
  onMessage={(e: WebViewMessageEvent) => {
    const data = JSON.parse(e.nativeEvent.data);
    const { error, evaluation } = data;
    if(error === null){
      // The JS evaluation succeeded!
      console.log(`The expression evaluated as: ${evaluation}`);
    } else {
      // The JS evaluation failed
      console.log(`The expression evaluation failed, with message: ${error}`);
    }
  }}
  /* You could still inject it on all frames, but you'll see we're evaluating 1+1 so we only need it once. */
  injectedJavaScriptForMainFrameOnly={true}
  injectedJavaScript={`
try {
  const expressionToEvaluate = 1 + 1;
  // We can only post strings.
  window.ReactNativeWebView.postMessage(
    JSON.stringify({ error: null, evaluation: expressionToEvaluate })
  );
} catch (e) {
  // Error objects don't stringify well, so we'll just report the error message.
  window.ReactNativeWebView.postMessage(
    JSON.stringify({ error: e.message, expressionToEvaluation: null })
  );
}
`}
/>
```

safaiyeh:
@shirakaba Seems like a lot of friction for the end dev to get the old behavior to work. Is the removal necessary, is it possible to respond to the prop changes; however, and maintain attaching `jsEvaluationValue` to `onLoadingFinish`.

My concern is that the most common use is a static inject with no prop updates. I would not want to create friction with those developers.

shirakaba:
> Seems like a lot of friction for the end dev to get the old behavior to work.

Note the code comment on the [injectJavaScript](https://github.com/react-native-community/react-native-webview/blob/master/src/WebView.ios.tsx#L154-L159) method:

> Injects a javascript string into the referenced WebView. **Deliberately does not return a response** because using `eval()` to return a response breaks this method on pages with a Content Security Policy that disallows `eval()`. **If you need that functionality, look into postMessage/onMessage.**

`injectJavaScript` wraps the same `evaluateJavaScript` method used in the old behaviour – I think that the old behaviour (of returning a response upon evaluating some JS on `onLoadFinish`) is not secure, but has remained up until now due to not being noticed.

The documentation comment recommends migrating to `postMessage` instead, so I think it would make sense to take this opportunity to deprecate that method.

>  Is the removal necessary, is it possible to respond to the prop changes

Although we can certainly make the old method respond to prop changes, the old method only supports injecting JS into the main frame; the only way to inject into all frames is to migrate to WKUserScripts (as in this PR).

So I think that the added friction (which is inconvenient but at least surmountable) should be weighed up against the fact that there is simply no way, using the old method, to inject JS into non-main frames.

safaiyeh:
Sounds good, glad we got to talk about the pro/cons of these changes. I agree that the friction is warranted given the limitations & concerns.

Thanks for being very thorough with all of your responses @shirakaba! You've been super helpful helping me understand the changes and thank you for creating this PR. 

I tested your changes on my production app and everything was smooth. 

Will prep the Android PR for `injectedJavaScriptBeforeContentLoaded`, so we can merge these in at the same time.

ios/RNCWebView.m
```diff
@@ -162,51 +169,8 @@ - (WKWebViewConfiguration *)setUpWkWebViewConfig
 
   // Shim the HTML5 history API:
   [wkWebViewConfig.userContentController addScriptMessageHandler:[[RNCWeakScriptMessageDelegate alloc] initWithDelegate:self]
-                                                            name:HistoryShimName];
```
safaiyeh: This cut off a piece of the function, could you restore this?

shirakaba: Now resolved – sorry for the trouble!

rekmarks:
Hello @shirakaba and @safaiyeh,

At MetaMask, we're looking to rebase [our fork of `react-native-webview`](https://github.com/MetaMask/react-native-webview). We have particular needs for JavaScript injection (as indicated by e.g. [here](https://github.com/react-native-community/react-native-webview/pull/1099#issuecomment-570980669)), and would like to wait for this and #1099 before rebasing.

May I ask what's the status of these two PRs, and if there's a timeline for merging them?

shirakaba:
@rekmarks I've just brought this iOS PR back up to date with upstream's master branch. I'll run it on a simulator to make sure that the merge was successful (though from the diff, it looks good). It sounds like that'll be enough to make it eligible for merge.

After that, I'll document the limitations of the Android implementation, then that will probably be eligible for merge, too.

I've designed the Android one to augment the iOS one's docs, so the iOS one should be merged before the Android one.

I plan to have these both ready within a few hours 👌

rekmarks:
@shirakaba amazing, 简直太棒了, thank you!

shirakaba:
@rekmarks 不用谢！

All tests pass :) requesting review now.

image

safaiyeh:
Awesome @shirakaba going to build it on my lunch break then approve if all goes well :)

safaiyeh:
@shirakaba  Could you add this to the docs: https://github.com/react-native-community/react-native-webview/pull/1119#issuecomment-574919464

"Retaining the behavior of collecting a JS evaluation value after injectedJavaScript" so people understand migration steps needed.

safaiyeh:
Screen Shot 2020-03-17 at 12 37 10 PM
Verified it worked! Just pending adding the migration docs and I'll land this as v9

shirakaba:
@safaiyeh Thanks for building and testing!

As alluded to in my hasty edit, I've updated the test (the one you've run is my older, broken test). This one's more thorough (and more importantly, not broken):

image

For some reason, injection into iframes upon the `beforeContentLoaded` timepoint isn't working. As this mirrors the general behaviour of the Android implementation, it may simply be that my understanding has been wrong all along and we're looking at a WebKit limitation. By that reasoning, `beforeContentLoaded` may refer to `window.top.document` rather than each frame's `window.self.document`, so maybe the iframes don't exist at this point so we wouldn't expect them to be injectable `beforeContentLoaded` anyway.

> Verified it worked! Just pending adding the migration docs and I'll land this as v9

Good idea – I'll get on that next.

safaiyeh:
That makes sense that the frames do not exist before content is loaded; therefore, nowhere to inject into! Since this would bring Android & iOS parity, all the better. 

I'll check out your updates

shirakaba:
@safaiyeh Docs updated.

safaiyeh:
Landed! @shirakaba thank you so much for your persistence from start to end. ❤️

rekmarks:
I feel pampered. Thanks guys! 🙌

shirakaba:
Thank you @safaiyeh for the review and merge, and @rekmarks for the motivation!

redpandatronicsuk:
This PR brakes  `injectedJavaScript` for me, I opened an issue here: https://github.com/react-native-community/react-native-webview/issues/1260

PR #1132, opened 2020-01-10
fix(types): Add missing applicationNameForUserAgent type in WebViewSharedProps

williamchong007:
# Summary

In Typescript, using `applicationNameForUserAgent` in `<WebView>` would throw a type error
```
  Property 'applicationNameForUserAgent' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<WebView> & Readonly<WebViewProps> & Readonly<{ children?: ReactNode; }>'.
```
This is due to https://github.com/react-native-community/react-native-webview/pull/707 moved the type definition from `WebViewSharedProps` to `CommonNativeWebViewProps`. We should probably define the type in both props.

## Checklist

### DIFF
--- a/src/WebViewTypes.ts
+++ b/src/WebViewTypes.ts
@@ -736,4 +736,9 @@ export interface WebViewSharedProps extends ViewProps {
    * Should caching be enabled. Default is true.
    */
   cacheEnabled?: boolean;
+
+  /**
+   * Append to the existing user-agent. Overriden if `userAgent` is set.
+   */
+  applicationNameForUserAgent?: string;
 }


williamchong007:
Not sure who to ping or if my thoughts are correct

TheAlmightyBob:
@jamonholmgren @Titozzz How can we get this merged? This is a straightforward error in the types with a simple fix.

TheAlmightyBob:
@safaiyeh You seem to have been active on PRs lately, could you help get this tiny fix in? :)

ludgerey:
I tried to temporary fix this with a custom declaration file. But I can't get it to work. Any ideas what's wrong?

```
// globals.d.ts
// NOT WORKING 

import 'react-native-webview';

declare module 'react-native-webview' {
    interface WebViewSharedProps {
        /**
         * Append to the existing user-agent. Overriden if `userAgent` is set.
         */
        applicationNameForUserAgent?: string;
    }
}
```

PR #1144, opened 2020-01-14
fix(iOS): Meta method 'UIScrollViewContentInsetAdjustmentBehavior:' conflict warning

tido64:
# Summary

This converter is already defined in
[RCTScrollViewManager.m](https://github.com/facebook/react-native/blob/master/React/Views/ScrollView/RCTScrollViewManager.m#L40).
Redefining it in react-native-webview can cause undefined behaviour when
used. Since ScrollView is unlikely to go away, we can take a
dependency on it.

## Test Plan

All current tests should pass.

When building a React Native app with `react-native-webview`, the following
warning can be seen in the build logs:

```
Meta method 'UIScrollViewContentInsetAdjustmentBehavior:' in category from /~/Library/Developer/Xcode/DerivedData/App-geojljebaazfrrgvxfkjjpiwmdlg/Build/Products/Debug-iphonesimulator/React-Core/libReact-Core.a(RCTScrollViewManager.o) conflicts with same method from another category
```

The warning should go away after removing the offending function.


safaiyeh:
@tido64 This might be necessary given: https://github.com/react-native-community/react-native-webview/commit/493b65de21b79194b697bb839690339cd826404a

Does this affect this functionality?

tido64:
@safaiyeh This patch does not remove any properties. It only removes a duplicate function for serializing/deserializing `UIScrollViewContentInsetAdjustmentBehavior`. The exact same code already exists in [core](https://github.com/facebook/react-native/blob/master/React/Views/ScrollView/RCTScrollViewManager.m#L40), so there is no reason for us to redefine it here.

Xcode warns us about the duplicate because it may cause undefined behaviour; the runtime doesn't know which one to call and will end up calling one on random. It just so happens that the two methods are exactly the same in our case.

PR #1156, opened 2020-01-20
Allows extra WebKit postMessage handlers to be registered outside of the default 'ReactNativeWebView' handler.

jcarioti:
# Summary

In our port from a native iOS project in Swift to React-Native, we noticed that ReactNativeWebview only registered a single postMessage handler. Our existing implementation relied on multiple different handlers being registered, and the presence of those handlers in the window.webkit.messageHandlers would indicate to the web app whether or not the native app could service the request. 

The ReactNativeWebview approach limits us from registering more than a single message handler, which then restricts our web app from knowing what the native app is capable of handling. 

This pull request allows users to register extra messageHandlers in the WebView props which, when called, trigger the onMessage callback just as the current implementation does. Since multiple handlers call this callback, the 'event' object gets an extra 'handler' property to identify which messageHandler was called.

The solution is implemented as an optional array of messageHandlers passed as a prop on the WebView. In the background, the extra messageHandlers are registered and are wired up alongside the default 'ReactNativeWebView' handler. Custom handlers are piped through the same 'didReceiveScriptMessage' as the historyShim and defaultHandler. They are treated as a third-option in the conditional. Data for the messageHandler is optional since calling the postMessage alone is enough for certain scenarios ('logout', for example, does not require any extra data to be passed. Just the acknowledgement that 'logout' messageHandler was called is enough to perform the action).

## Test Plan

I've tested by passing expected and unexpected types into the 'messageHandlers' prop. Expected is an empty array, null, undefined, or an array of strings. Unexpected values for 'messageHandlers' would be numbers, an object, an array of non-string values. Unexpected types cause runtime errors, expected types don't.

```
<WebView
  ...
  messageHandlers={ [ ] }
  messageHandlers={ [ "messageHandler1", "handler 2" ] }
  messageHandlers={ [ "messageHandler1", { } ] }
  messageHandlers={ undefined }
  messageHandlers={ null }
  messageHandlers={ 1234 }
  messageHandlers={ "fail" }
/>
```

I've also tested from the web app -> native boundary by calling postMessage with string data, an object, null, and undefined. The custom message handlers have a condition that allows for an empty or undefined payload.

`window.webkit.messageHandlers['registeredHandler'].postMessage();`
`window.webkit.messageHandlers['registeredHandler'].postMessage('string data');`
`window.webkit.messageHandlers['registeredHandler'].postMessage({ data: 'object' });`
`window.webkit.messageHandlers['registeredHandler'].postMessage(null);`

Required for testing would be a React Native WebView project that registers custom postMessage handlers and a web app that would call these custom handlers with various types of data. 

## Compatibility

This is only compatible with the iOS project for now. I would like to make this change universal for the Android project as well, but would be open to discussion before committing to that piece. 

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


shirakaba:
I've worked around lack of postMessage handlers just by specifying a controller in the posted message:

## JS to be called from the WebView's JSContext

```js
window.ReactNativeWebView.postMessage(
    JSON.stringify({ controller: "logger", message: "hello" })
);

window.ReactNativeWebView.postMessage(
    JSON.stringify({ controller: "adder", x: 123 })
);
```

## Message handler to set up on the WebView

```ts
const accumulator: number = 0;

function onWebViewMessage(e: WebViewMessageEvent){
  const data: any = JSON.parse(e.nativeEvent.data);
  
  if(!data.controller){
    console.error("Expected WV message to specify controller.");
    return;
  }

  switch(data.controller){
    case "logger":
      console.log(`[From WV] ${data.message}`);
      break;
    case "adder":
      accumulator += data.x;
      break;
  }
}
```

## How to set up the message handler on the WebView component

```ts
<WebView onMessage={onWebViewMessage}/>
```

---

Doing it this way means that no further APIs need to be added, nor does an Android implementation need to be kept in step. Is there any benefit to having genuinely separate message handlers rather than just switching based on some feature of the message?

jcarioti:
Thanks for the response. What you've described is how I originally decided to handle the situation.

> Is there any benefit to having genuinely separate message handlers rather than just switching based on some feature of the message?

Yes, absolutely. The issue is one-way communication vs two-way communication.

### Our web app has logic that works like this:

```
if (window.webkit.postMessage[handler]) {
  window.webkit.postMessage[handler](data);
  return true;
}
return false;
```

This only works if the native app registers separate handlers for each message that it has logic for. If the native app doesn't register a handler with that name, then the native app is not capable of handling that message and shouldn't be sent.

If a message handler is NOT present, then our web app handles the logic in a more web-esqe manner. Essentially the logic is "let the native app tell us if it's capable of handling the request, and if not, let the web app handle the request instead". 

### The RN approach only gives us a single handler that's always present

This means our web app never knows what the native app is capable of handling. There is no return value from a postMessage, so the web app has to blindly fire the message and ALWAYS rely on the native app to service the request. We don't always plan on having the exact same featureset from Android to iOS (think SiriKit stuff), so this single handler implementation really hinders what we've already established as a working pattern.

shirakaba:
I see – I'm in two minds if that's your use-case. Should the poster know about the available messageHandlers? Arguably so – it's admittedly hard to provide a strong argument against that premise. However, there's an alternative approach you could take that doesn't involve implementing anything new.

You could configure the WebView to, before contentLoad, inject an array into the global scope that announces to the web app which controllers (as I've described `controller` in my code snippet) will be available for communication with.

This keeps the implementation of message-handling simple for both iOS and Android.

jcarioti:
I've already come to this solution. The issue isn't whether or not I can make my code work again as much as it's about making React-Native work as closely to our native app counterparts as possible. It doesn't make sense that we need to make server-side changes to support a migration to a RN app... the app is still iOS or Android underneath and should be using the same mechanisms that the native OS uses. 

The perk of RN is that you're still using native views and native APIs. This suggestion introduces a new paradigm (announcing the supported messageHandlers/controllers in the global scope) that is already supported by the native strategy (adding supported messageHandlers to the webkit postmessage global scope). It's just a roundabout way of doing it instead of simply supporting what the OS is capable of in the first place. 

> This keeps the implementation of message-handling simple for both iOS and Android.

Should this project be focused on keeping things simple or should it try to be as transparent as possible with the features that the OS provides? The API for adding multiple messages is there... let's just wire into it...

jcarioti:
**I should also mention that I intend on supporting Android as well as iOS with this change. The original commit only supported iOS with the intention of having this discussion first.** If the contention is the fact that I only added iOS support, then I would totally understand the hesitation to make this change. I would like to have both iOS and Android supporting multiple messages with the exact same RN interface as an end goal.

PR #1164, opened 2020-01-23
Create a macOS implementation of WebView and a macOS example app

tom-un:
# Summary

This change adds a macOS implementation of the <WebView> component.  It shares the same Objective C source as the iOS implementation and as such is also based on Apple's WKWebView class.   

The Apple WKWebView is nearly identical between iOS and macOS but is missing an important property: `scrollView`.  The biggest impact is that the <WebView> `onScroll` events cannot be handled on mac.   Similarly, the `<WebView>` `backgroundColor` property can't be fully implemented.  On iOS `backgroundColor` sets the color of the `scrollView`.   In lieu of access to scrollView, it is possible to make the mac WKWebView background transparent, so if `backgroundColor`'s alpha channel is fully transparent I make the WKWebView transparent.  Also on iOS, if the `<WebView>` `scrollEnabled` prop is false, it disables scrolling on the `scrollView`.  On macOS I implemented scroll disabling by subclassing WKWebView and overriding `scrollWheel:`.

The PR introduces something new: it takes a dependency on a "out-of-tree" platform for the first time.   The FB react-native repo of course contains android and ios implementations, but the macOS implementation is in a separate repo.   Having two react-native implementations in node_modules creates some challenges for the metro bundler.  The metro issue is being discussed here https://github.com/react-native-community/discussions-and-proposals/issues/182, but in lieu of an "official" solution I'm using a workaround involving `react-native.config.js` and a pair of `metro.config.js` files for the two react-native implementations.

For the first iteration of the PR I'm consuming react-native-macos via the `git:` url to the repo, but I'm separately working on getting react-native-macos published as its own npm package and will switch to using the real package shortly.

react-native-macos doesn't yet have a CLI package so it's not yet possible to do `react-native run-macos`.   It's on our near term list of things to develop but in the meantime testing is a bit more manual than ios or android: it involves opening the macOS example XCode project manually and starting the bundler via `yarn start:macos`.

## Test Plan

The macOS implementation was tested using a new macOS test app parallel to the recently added ios and android test apps.   

WebView-macos

The prerequisites for testing macOS are identical to iOS: https://facebook.github.io/react-native/docs/getting-started.

```bash
$ cd react-native-webview
$ yarn install
$ open example/macos/example.xcodeproj
$ yarn start:macos
```

In XCode, select the `example-macos` target and Run.

The three test pages in the example app function except that you'll notice on the "Scrolling" page that scrolling the WebView's contents do not trigger `onScroll` events as occur on iOS and Android.

In addition, iOS and Android were re-tested via `yarn start:ios` and `yarn start:android`.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| macOS     |    ✅❌     |

## Checklist


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-native-webview",
-  "description": "React Native WebView component for iOS, Android, and Windows 10 (coming soon)",
+  "description": "React Native WebView component for iOS, Android, macOS, and Windows 10 (coming soon)",
```
jamonholmgren: We should remove the Windows 10 reference, as we no longer plan to support it (unless someone steps up to champion it.)

shirakaba:
Just came across this PR! Very excited to see macOS activity in community repos.

I like the approach of using compiler directives for adding macOS as a platform as it greatly reduces the maintenance burden, given how similar macOS's WKWebView is to that of iOS.

It's a fossil now, but I [ported](https://github.com/shirakaba/react-native-wkwebview) `react-native-wkwebview` to ptmt's `react-native-macos` a couple of years ago. I found for that port that there was nothing to change, relative to the iOS version, beyond addressing any references to the `scrollView` and `backgroundColor`.

Sadly, I [did ask](https://twitter.com/LinguaBrowse/status/1011676685580865539?s=20) the WebKit team at the time to expose the scroll view and they confirmed that they were "tracking it internally", but evidently still no progress. They're [asking for enhancement requests again](https://twitter.com/ddkilzer/status/1209170163339149312?s=20) as of December, but it's a bit too late to help the `react-native-macos` effort.

I also found that the macOS version of WKWebView uniquely has a [`allowsMagnification`](https://developer.apple.com/documentation/webkit/wkwebview/1414983-allowsmagnification) property that needs to be set to `TRUE` if one wants to support pinch-to-zoom on macOS just as in Safari.
We'll have to take care about how the lack of a scrollView interacts with React Native Reanimated, but as my PR for Reanimated support #1102 is still pending, we can revisit that later.

shirakaba:
> The biggest impact is that the onScroll events cannot be handled on mac. 

I wonder if we could forward scroll events some other way. For instance, would it be acceptable to inject a WKUserScript, `atDocumentStart` and `forMainFrameOnly` that attaches a scroll event listener to the `window` and fires events to the WKWebView's script message handler? This message handler could then forward the event to React Native.

Current scroll offset, if users need to access it as a property, can be determined by `window.pageXOffset` and `window.pageYOffset`. The main thing I haven't found is a way to read the current zoom factor.

tom-un:
> > The biggest impact is that the onScroll events cannot be handled on mac.
> 
> I wonder if we could forward scroll events some other way. For instance, would it be acceptable to inject a WKUserScript, `atDocumentStart` and `forMainFrameOnly` that attaches a scroll event listener to the `window` and fires events to the WKWebView's script message handler? This message handler could then forward the event to React Native.
> 
> Current scroll offset, if users need to access it as a property, can be determined by `window.pageXOffset` and `window.pageYOffset`. The main thing I haven't found is a way to read the current zoom factor.

Thanks @shirakaba !   That sounds like a reasonable way to work around the lack of `scrollView`.   Although I would prefer it be added in a later PR.   Same for `allowsMagnification` prop.

src/WebViewTypes.ts
```diff
@@ -281,6 +289,27 @@ export interface IOSNativeWebViewProps extends CommonNativeWebViewProps {
...
+  allowsInlineMediaPlayback?: boolean;
+  allowsLinkPreview?: boolean;
+  automaticallyAdjustContentInsets?: boolean;
+  bounces?: boolean;
+  contentInset?: ContentInsetProp;
+  contentInsetAdjustmentBehavior?:
+    | 'automatic'
+    | 'scrollableAxes'
+    | 'never'
+    | 'always';
```
safaiyeh: This would be nicer as a enum. So I can import these values and make sure I'm using the correct string.

src/WebView.macos.tsx
```diff
@@ -0,0 +1,347 @@
...
+  static defaultProps = {
+    javaScriptEnabled: true,
+    cacheEnabled: true,
+    originWhitelist: defaultOriginWhitelist,
+    useSharedProcessPool: true,
+  };
+
+  static isFileUploadSupported = async () => {
+    // no native implementation for macOS, depends only on permissions
+    return true;
```
safaiyeh: Should this be false?

safaiyeh:
Also is the `example` target needed in the macOS example? Could it be removed and just have the macOS target?

src/WebViewTypes.ts
tom-un: Sure.   But this is identical to iOS.   I can add an enum but I should make both the iOS and macOS conform to the same enum

src/WebView.macos.tsx
tom-un: I assumed it would be the same as iOS.   On Android there is an actual native module method that gets called.   On iOS (and now macOS) there is no native method, so it just returns true.

tom-un:
> Also is the `example` target needed in the macOS example? Could it be removed and just have the macOS target?

I added the iOS target to the mac xcodeproj just because it makes it really easy to test that iOS still builds and runs when making macOS changes.

src/WebView.macos.tsx
safaiyeh: Sounds good.

safaiyeh:
> > Also is the `example` target needed in the macOS example? Could it be removed and just have the macOS target?
> 
> I added the iOS target to the mac xcodeproj just because it makes it really easy to test that iOS still builds and runs when making macOS changes.

That sounds good!

PR #1184, opened 2020-02-03
WebView removed from RN-0.61

flyingcircle:
# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist


flyingcircle:
This is causing an error with react-native-web since they recently upgraded to RN-0.61 and this import was removed in that version of RN.

PR #1194, opened 2020-02-10
Convert RNCWebViewPackage to Kotlin

dulmandakh:
# Summary

Make RN implementation dependency, because RNW won't export RN API, but used as RN module. Convert RNCWebViewPackage to Kotlin, step forward to become Kotlin.

There is no visible user change, but build performance.

## Test Plan

Will build and run as expected.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist


dulmandakh:
I'll be making bit sized changes and would like to migrate to https://developer.android.com/reference/androidx/webkit/package-summary, which would solve many compatibility issues.

shirakaba:
@dulmandakh I'm very interested in `androidx.webkit`, though a migration to something like that should be raised in a separate PR.

I'm concerned that I don't see certain WKWebView APIs in the `androidx.webkit` documentation, e.g. `WKUserScript`, `WKUserContentController`, `WKScriptMessage`, `WKScriptMessageHandler`, `WKNavigationDelegate`, etc. These APIs are critical to implementing functionality equivalent with the iOS port. Do you know where the documentation for these APIs might be, and how to call them from `androidx.webkit`?

dulmandakh:
I’m not sure about androidx.webkit implementation details, but would like to convert to Kotlin first to reduce code size and make code more maintainable.

safaiyeh:
Looks good! Thanks for this @dulmandakh

PR #1203, opened 2020-02-12
Implement direct communication between Android code and JS

### LINKED ISSUE #1120: Communication via onMessage is slow on Android

Our app moves relatively large sets of data from WebView to native JS. During testing, we encountered noticeable delays in event delivery which seemed to scale linearly with message size. As the delay seemed to be higher on Android than iOS, I took a quick look into the code.

Looking at iOS implementation, https://github.com/react-native-community/react-native-webview/blob/07a87131dcfccb51465859986a78fd4ed29b5caf/ios/RNCWebView.m#L50 https://github.com/react-native-community/react-native-webview/blob/07a87131dcfccb51465859986a78fd4ed29b5caf/ios/RNCWebView.m#L442-L446 it shows iOS runs the handler implementation in native JS directly.

Looking at Android, https://github.com/react-native-community/react-native-webview/blob/07a87131dcfccb51465859986a78fd4ed29b5caf/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L150-L155 https://github.com/react-native-community/react-native-webview/blob/07a87131dcfccb51465859986a78fd4ed29b5caf/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L1088-L1107 https://github.com/react-native-community/react-native-webview/blob/07a87131dcfccb51465859986a78fd4ed29b5caf/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L1153-L1156 it shows Android implementation utilising full React event queue.

Quick (unscientific) measurements for a 6.5MB payload (stringified JSON) resulted in ~5200ms on Android and ~1000ms on iOS.

I am curious if this difference is intended, and if it could be possible to move towards a more unified implementation?

I did a quick proof-of-concept implementation for Android to also call the native JS handler directly, replacing the injected method as such:
```
@JavascriptInterface
public void postMessage(String message) {
  ReactContext reactContext = (ReactContext) mContext.getContext();

  if (reactContext != null) {
    CatalystInstance catalystInstance = reactContext.getCatalystInstance();
    WritableNativeArray params = new WritableNativeArray();
    params.pushString(message);
    catalystInstance.callFunction("MessageHandlerModule", "handleMessage", params);
  }
}
```

The handler was exposed in JS like this:
```
class MessageHandler {
  handleMessage = message => {
    // handle message
  };
}

const messageHandler = new MessageHandler();
BatchedBridge.registerCallableModule('MessageHandlerModule', messageHandler);
```

This implementation brought the delay on Android down to ~1800ms (same payload).

Thoughts? Comments?
CC: @Titozzz, @jamonholmgren

ivari:
# Summary
**Changes onMessage handling on Android from React Event Dispatcher to direct call over BatchedBridge, similar to how iOS is implemented.**

Looking at RNCWebViewManager.java, it implements message passing from WebView contents to React Native code over React Event Queue, which for large messages can be noticeably slow.

Switching to direct calls over BatchedBridge (something similar to how iOS handles it) displays considerable improvement in speed.

For local tests of messages in ~6.5MB in size, I noticed improvements of ~65%, from 5200ms to 1800ms. This is more comparable to iOS implementation, which took ~1s.

Fixes #1120.

Any suggestions and questions welcome.

## Test Plan
Testing done and before/after is detailed under #1120.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    not applicable     |
| Android |    ✅     |

## Checklist

Below are not applicable as the PR implements no feature changes, only internal behaviour.


macobo:
cc @safaiyeh, could you take a look at this and provide us feedback where to proceed with this patch?

As mentioned in the PR description, this improves latency for large messages immensely on android, so we'd be super interested in rolling this out to users.

safaiyeh:
This is great! @ivari @macobo, what is the best way for me to test this? Is there a concern of this breaking current functionality?

src/WebView.android.tsx
```diff
@@ -10,6 +10,8 @@ import {
   findNodeHandle,
 } from 'react-native';
 
+import BatchedBridge from 'react-native/Libraries/BatchedBridge/BatchedBridge';
```
safaiyeh: This would make webview dependent on a specific version of React Native. Peer dependencies need to be updated & docs to make sure users know this fix can only support a certain RN version

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -1096,16 +1102,38 @@ public void run() {
...
+        if (reactContext != null) {
+          this.sendDirectMessage(reactContext, eventData);
+        } else {
+          dispatchEvent(this, new TopMessageEvent(this.getId(), eventData));
+        }
       }
     }
 
+    protected void sendDirectMessage(ReactContext reactContext, WritableMap data) {
+      CatalystInstance catalystInstance = reactContext.getCatalystInstance();
```
safaiyeh: Can the instance be a class variable? To avoid initializing this instance ever time onMessage is called.

src/WebView.android.tsx
ivari: I'm not sure I follow. BatchedBridge has been there since RNv0.6.0, no?

ivari:
> This is great! @ivari @macobo, what is the best way for me to test this? Is there a concern of this breaking current functionality?

Hey! Apologies for the slightly delayed response.

I assume you mean how to test the speed? I used something like this:

    import React, { Component } from 'react'

      render() {
        const html = '<html><head></head><body>dummy html</body></html>'

        const run = () => "window.ReactNativeWebView.postMessage('${Date.now()}')"

        setTimeout(() => {
          this.webref.injectJavaScript(run);
        }, 3000)

        return (
          <View style={{ flex: 1 }}>
            <WebView
              ref={r => (this.webref = r)}
              source={{ html }}
              onMessage={event => {
                console.log({
                  start: event.nativeEvent.data,
                  end: Date.now(),
                })
              }}
            />
          </View>
        )
      }
    }

`injectJavaScript` should for all intents and purposes be instant, so this should give you a way to get the time from calling `ReactNativeWebView.postMessage` to when it was handled with `onMessage`.

For larger payloads, simply adjust the code to include more data in the `postMessage` call.

As for breaking current functionality, I don't see any specific risks here, unless there's some inherent instability in using BatchedBridge of which I am not aware.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
ivari: Changed accordingly.

Fr33maan:
Hi there, I have a very naive question.
As injectJavascript takes a string, does it means it get `eval()` in the webview ? 
If so, as eval is pretty slow, how could we improve performances ?

ivari:
> Hi there, I have a very naive question.
> As injectJavascript takes a string, does it means it get `eval()` in the webview ?
> If so, as eval is pretty slow, how could we improve performances ?

I would assume that it gets handled as any other `<script>` tag or a `.js` file by the WebView. There is no faster way to include scripts than that.

Fr33maan:
And post message was doing do too? Or using another mechanism? I'm making a
game and keeping the joystick stuff in the UI thread but still need to send
commands to the webview. I throttle the messaging so it's ok but I was
curious if the deprecated postMessage would be more performant or not.

src/WebView.android.tsx
jamonholmgren: It's probably fine, but if React Native changes where the BatchedBridge lives, this could break. I'm not too worried about it. Try changing it to `/foo` and see how it breaks, and perhaps we could check for that and provide a nicer error message.

jamonholmgren:
Is there any documentation that needs to be updated for this?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -1095,16 +1112,36 @@ public void run() {
             }
             WritableMap data = mRNCWebViewClient.createWebViewEvent(webView, webView.getUrl());
             data.putString("data", message);
-            dispatchEvent(webView, new TopMessageEvent(webView.getId(), data));
+
+            if (mCatalystInstance != null) {
```
luancurti: Could you create another method to remove the duplicated code?

index.js
```diff
@@ -1,4 +1,4 @@
-import WebView from './lib/WebView';
+import WebView from './src/WebView';
```
jamonholmgren: Why is this changed?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
jamonholmgren: I wouldn't worry about the duplicated code in this instance. It's a pretty minor bit of duplication, close together.

ivari: I tried to keep the additions close to existing code there.

index.js
ivari: Because it complained on compile for me. You are right though, that should not be part of this PR. Will revert.

src/WebView.android.tsx
ivari: Thank you, I will give it a try and see if I can add an error message.

acerbastimur:
Waiting for that feature...

cristianoccazinsp:
Trying to educate myself a little about this subject. Could someone point me into some docs about how this can be used to send events from native to JS? I can't find any of this on the RN docs.  Where did you learn this from?

soroushm:
Good Job!
it will be grate have this on ios

fungilation:
@ivari @safaiyeh This PR is bugged on Android. `onMessage` prop on different WebView gets overridden.

ex. I have 2 components that wrap 2 WebView, with prop `onMessage={this.onBridgeMessage}` on each that calls `onBridgeMessage` within each comp. I'll call these comp A and B. onMessage from within A's webview would call B.onBridgeMessage.

This was not the case on Android, up to react-native-webview 9.0.2. As such, I have to downgrade to it in the mean time.

ivari:
> @ivari @safaiyeh This PR is bugged on Android. `onMessage` prop on different WebView gets overridden.
> 
> ex. I have 2 components that wrap 2 WebView, with prop `onMessage={this.onBridgeMessage}` on each that calls `onBridgeMessage` within each comp. I'll call these comp A and B. onMessage from within A's webview would call B.onBridgeMessage.
> 
> This was not the case on Android, up to react-native-webview 9.0.2. As such, I have to downgrade to it in the mean time.

Thank you @fungilation for noticing this! I will see what I can do to get this fixed.

ivari:
> Trying to educate myself a little about this subject. Could someone point me into some docs about how this can be used to send events from native to JS? I can't find any of this on the RN docs. Where did you learn this from?

@cristianoccazinsp, it's not in the docs. I just googled it and read the code. The preferred way to do this is using React Events (https://reactnative.dev/docs/native-modules-android#sending-events-to-javascript), but due to this method copying the message around 6-7 times it gets *very* slow for large messages. You can read more at https://github.com/react-native-community/react-native-webview/issues/1120

> Good Job!
> it will be grate have this on ios

@soroushm, thanks! On iOS it already works like this, so no changes were necessary.

shirakaba:
Does this happen to improve (or even fix) https://github.com/react-native-community/react-native-webview/issues/1250 in any way?

ivari:
> Does this happen to improve (or even fix) #1250 in any way?

@shirakaba, perhaps the addition of componentDidMount delayed the race condition somewhat? Other than that, I am not familiar with https://github.com/react-native-community/react-native-webview/issues/1250, but on the first look I don't see any actual connection between the two.

PR #1210, opened 2020-02-16
fix(Android): Don't show camera options for a file upload when they can not be used

dvicory:
# Summary

Don't show camera options for a file upload that would result in nothing happening for the user.

On Android, if the application declares the camera permission, then even intents that use the camera require permission to be granted. This is a problem for apps that combine an in-app camera with a WebView that has file uploading and the user has not given permission for the camera.

Note, this will not request permission for camera. This will simply prevent showing the camera options that would be a no-op action for users. It does this by checking if the camera permission is declared, and if so, checks that the user has granted permission.

[More information on this camera permission/intent issue.](https://blog.egorand.me/taking-photos-not-so-simply-how-i-got-bitten-by-action_image_capture/)

## Test Plan

We have had this patch in our app for some time. I also tested it by modifying the example app.  First I modified the WebView to go to a page that has file inputs that would trigger the camera options (an example being https://addpipe.com/html-media-capture-demo/). I verified the camera options are present.

Then I declared the camera permission by adding this to the `AndroidManifest.xml`:

```xml
<uses-permission android:name="android.permission.CAMERA" />
```

Then I went back to the demo site and verified the camera options were missing. In fact, we go straight to the file picker.

###  Before patch with camera permission declared

device-2020-02-16-150140

### After patch with camera permission declared

device-2020-02-16-145914

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [x] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
@@ -32,6 +32,7 @@ import com.facebook.react.modules.core.PermissionListener;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 
 import static android.app.Activity.RESULT_OK;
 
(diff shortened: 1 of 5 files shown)


safaiyeh:
This is awesome! @dvicory I'll review this tomorrow

safaiyeh:
Now that we have an example app, would you mind adding an example for this use case as well? @dvicory

dvicory:
> Now that we have an example app, would you mind adding an example for this use case as well? @dvicory

I was thinking that, but I wasn't sure what the best way to show an example would be. If for demo purposes, toggling the permissions in system/app settings is fine (otherwise we'd be talking about adding `react-native-permissions` potentially...), then I could add a screen with just a simple file input that would normally have camera options shown. If the permission is granted from settings, then they are there, if not, they will be missing.

What do you think?

safaiyeh:
@dvicory Hmm, adding `react-native-permissions` seems like too much work (unless it is not, this would be ideal). With toggling w/ settings, just add a note on the example on how to toggle it.

safaiyeh:
Approved pending example, @dvicory

dvicory:
Thanks @safaiyeh. I've updated the PR with an example and extra information to the documentation.

safaiyeh:
This is awesome. I really appreciate the work you put in this! @dvicory merging it now!

Issue #1187, opened 2020-02-04
Cant Download Zip Files on iOS 


Krypternite:
**Bug description:**
Not able to download **zip** and other file formats on iOS .

**To Reproduce:**

**Expected behavior:**
When I click the download link, the download should start. The download starts on Android but on iOS the webview tries to open the zio file and shows the encoded data instead. When I tried this on safari, I was able to download the file and save it on the iPad.

**Screenshots/Videos:**
![image](https://user-images.githubusercontent.com/9376390/73755673-642ebe80-478c-11ea-8f3a-a10012ed060a.png)
Upon clicking the download link I am getting this. 
![image](https://user-images.githubusercontent.com/9376390/73755705-6db82680-478c-11ea-9a07-d9b951b5a1c7.png)

**Environment:**
 - OS: iOS
 - OS version: > 12.0
 - react-native version:  > 0.61
 - react-native-webview version: latest

dvicory:
Unfortunately, the underlying WKWebView APIs are extremely limiting here. There are no true file downloads in WKWebView. WKWebView will basically attempt to preview any URL given to it. Android has a dedicated handler you can implement when WebView detects there is a download, for which there is no equivalent on iOS. The only true file download for WKWebView is long pressing on an image and saving it.

Our team has struggled off and on for months with various workarounds to overcome this. I've come to the conclusion there does not exist a technique that can universally be applied to this library with the WKWebView API as it is.

What I can tell you is ways we overcame it.

The first strategy we took was using `onShouldStartLoadWithRequest` that compares the URL we're trying to load to a list of regex URLs that are actually downloads. If it matches that list, we `return false`, blocking the loading of it, and have other code that does the download, and shows a file preview (we used [react-native-file-viewer](https://github.com/vinzscam/react-native-file-viewer)) that allows for saving to the device.

Our evolution of that technique is our web app is aware that we're sending a download, and instead sends a custom protocol (like `rn-app://`) with a serialized message indicating a download, which we are able to similarly intercept, block, and handle ourselves.

After a long time looking at the WKWebView API, I feel like this is the best that can be currently done. At least until the API allows for getting at raw response headers and/or data without having to use a custom protocol...

Krypternite:
@

[...]

Issue #1126, opened 2020-01-09
Not Possible download on webiview


arunreenamohan-creator:
Hi.
I cant download any files on webview

I cant download any files on webview even in google page

arunreenamohan-creator:
pls help me iam struggled with this issue

arunreenamohan-creator:
<View style={{ flex: 1 }}>
 <WebView
 onLoad={() => this.hideSpinner()}
 style={{ flex: 1 }}

 source={{ uri:'https://www.google.com/'}}
  javaScriptEnabled = {true}





       />

dvicory:
What platform? What's the symptom?

Did you make sure to [follow the instructions here](https://github.com/react-native-community/react-native-webview/blob/361529630fbe7e0f743eaa71d72ba3f542553ec2/docs/Guide.md#add-support-for-file-download)?

For iOS, there is no true downloading of arbitrary files. Only file previews (e.g., opening a PDF will let you preview it, but there is no built-in way to save to the phone). You can long press images to save them. Implementing true downloading in WKWebView has been quite the epic for our team and we've settled on a custom solution involving a custom protocol, outside the purview of react-native-webview in my opinion.

PR #1214, opened 2020-02-19
feat(iOS): iOS file downloads

trcoffman:
# Summary

This PR adds an optional iOS-only callback `onFileDownload` to react-native-webview to help support file downloads.

Our application uses `react-native-webview` and has many parts of web-app that serve file downloads.

`onFileDownload` is called with the URL that you can use to download the file. When RNCWebView detects that the HTTP response should result in a file download, `onFileDownload` is called. The client can then provide code to download the file.

RNCWebView determines that a file download should take place if either of the following is true:
1. The HTTP response contains a `Content-Disposition` header that is of type 'attachment'
2. The MIME type of the response cannot be rendered by the iOS WebView

The primary component of this PR is the change to `decidePolicyForNavigationResponse` in RNCWebView.m. In this method, the `WKNAvigationResponse` and `NSHTTPUrlResponse` are inspected to see if a file download should take place. If an `onFileDownload` callback is registered, the navigation is canceled and `onFileDownload` is called. If `onFileDownload` isn't registered, the old behavior is unchanged and the WebView will try and render the file.

This PR is irrelevant on Android because on Android file downloads are already fully implemented using the `DownloadManager`. Even with this PR, Android still provides a superior developer experience because all that's needed is the right permissions. For iOS the client of `react-native-webview` has to implement the code that actually downloads the file themselves.

Rsolves Issues #1187, #1126.

Just use the example on an iOS device and go to the 'Downloads' test. You should see this: 
Screen Shot 2020-02-18 at 4 40 13 PM

Click the link in the webview. You should see this:

Screen Shot 2020-02-18 at 4 39 16 PM

If you comment out the line in Downloads.tsx that sets up the onFileDownload prop, you should see the old behavior when you click the link, which is to try and render the zip file in the webview:

Screen Shot 2020-02-18 at 4 39 57 PM

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     | ✅ |
| Android |  ❌  |

## Checklist

- [X] I updated the TS type files

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -67,6 +67,8 @@ static NSDictionary* customCertificatesForHost;
     UIScrollViewDelegate,
 #endif // !TARGET_OS_OSX
     RCTAutoInsetsProtocol>
+
+@property (nonatomic, copy) RCTDirectEventBlock onFileDownload;
 @property (nonatomic, copy) RCTDirectEventBlock onLoadingStart;
 @property (nonatomic, copy) RCTDirectEventBlock onLoadingFinish;
 @property (nonatomic, copy) RCTDirectEventBlock onLoadingError;
@@ -973,24 +975,42 @@ static NSDictionary* customCertificatesForHost;
   decidePolicyForNavigationResponse:(WKNavigationResponse *)navigationResponse
                     decisionHandler:(void (^)(WKNavigationResponsePolicy))decisionHandler
 {
+  WKNavigationResponsePolicy policy = WKNavigationResponsePolicyAllow;
   if (_onHttpError && navigationResponse.forMainFrame) {
     if ([navigationResponse.response isKindOfClass:[NSHTTPURLResponse class]]) {
       NSHTTPURLResponse *response = (NSHTTPURLResponse *)navigationResponse.response;
       NSInteger statusCode = response.statusCode;
 
       if (statusCode >= 400) {
-        NSMutableDictionary<NSString *, id> *event = [self baseEvent];
-        [event addEntriesFromDictionary: @{
+        NSMutableDictionary<NSString *, id> *httpErrorEvent = [self baseEvent];
+        [httpErrorEvent addEntriesFromDictionary: @{
           @"url": response.URL.absoluteString,
           @"statusCode": @(statusCode)
         }];
 
-        _onHttpError(event);
+        _onHttpError(httpErrorEvent);
+      }
+
+      NSString *disposition = nil;
+      if (@available(iOS 13, *)) {
+        disposition = [response valueForHTTPHeaderField:@"Content-Disposition"];
+      }
+      BOOL isAttachment = disposition != nil && [disposition hasPrefix:@"attachment"];
+      if (isAttachment || !navigationResponse.canShowMIMEType) {
+        if (_onFileDownload) {
+          policy = WKNavigationResponsePolicyCancel;
+
+          NSMutableDictionary<NSString *, id> *downloadEvent = [self baseEvent];
+          [downloadEvent addEntriesFromDictionary: @{
+            @"downloadUrl": (response.URL).absoluteString,
+          }];
+          _onFileDownload(downloadEvent);
+        }
       }
     }
-  }  
+  }
 
-  decisionHandler(WKNavigationResponsePolicyAllow);
+  decisionHandler(policy);
 }
 
 /**
--- a/apple/RNCWebViewManager.m
+++ b/apple/RNCWebViewManager.m
@@ -34,6 +34,7 @@ RCT_EXPORT_MODULE()
 }
 
 RCT_EXPORT_VIEW_PROPERTY(source, NSDictionary)
+RCT_EXPORT_VIEW_PROPERTY(onFileDownload, RCTDirectEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onLoadingStart, RCTDirectEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onLoadingFinish, RCTDirectEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onLoadingError, RCTDirectEventBlock)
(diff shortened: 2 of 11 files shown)


safaiyeh:
Hi @trcoffman, thank you for the PR. I added this to my TODO, I'll review it tomorrow

trcoffman:
**Don't merge this yet!**

It turns out that this PR only works on iOS 13+

https://developer.apple.com/documentation/foundation/nshttpurlresponse/3240613-valueforhttpheaderfield?language=objc

trcoffman:
This PR has been updated so that the call to `valueForHTTPHeaderField` is gated behind an iOS version check to prevent crashes on iOS versions before 13. The documentation changes have been updated to reflect this.

On iOS versions before iOS 13, the only conditions that will trigger file download is if the MIME type cannot be rendered by the webview. We can't inspect the content-disposition header without iOS 13.

trcoffman:
@safaiyeh Could you please take a look at this PR if you get a chance? Thanks!

safaiyeh:
@trcoffman thanks for the PR. Could you update `Reference.md` as well?

So the example shows an alert that the download is detected, how can I check that the file actually got downloaded?

trcoffman:
> @trcoffman thanks for the PR. Could you update `Reference.md` as well?
> 
> So the example shows an alert that the download is detected, how can I check that the file actually got downloaded?

@safaiyeh, this PR does not include code to actually download the file. It just gives clients of `react-native-webview` a callback that gives them the opportunity to supply their own file download code to fetch the file at the given URL.

Actually downloading the file is outside of the scope of this PR. iOS doesn't have a universal file download experience like android does, so it shouldn't be up to `react-native-webview` to actually download the file. Clients need the flexibility to download the file with their own code.

The alert in the example app is just to demonstrate that the callback works. You get the URL to a file if a file download is detected.

trcoffman:
@safaiyeh I just rebased the branch on master, please take a look. This is a really important feature to have so that we can have feature parity between iOS and android.

vpankov:
Will it pass to callback full request with auth cookies or just url?

trcoffman:
> Will it pass to callback full request with auth cookies or just url?

Just the URL. Feel free to extend it to include auth cookies.

trcoffman:
@safaiyeh Could you please take a look at this PR?

apple/RNCWebView.m
```diff
@@ -973,24 +975,42 @@ - (void)                    webView:(WKWebView *)webView
   decidePolicyForNavigationResponse:(WKNavigationResponse *)navigationResponse
                     decisionHandler:(void (^)(WKNavigationResponsePolicy))decisionHandler
 {
+  WKNavigationResponsePolicy policy = WKNavigationResponsePolicyAllow;
   if (_onHttpError && navigationResponse.forMainFrame) {
```
tonywok, line 979: @trcoffman It seems like access to this functionality requires that `_onHttpError` is set. Is there a reason for that?

geezylucas:
Android? 😭

apple/RNCWebView.m
trcoffman: So, you are right that this should be improved for the sake of readability. However, from taking another look at the code, it looks like _onHttpError will always be set whether or not the user provides an implementation, there is a default implementation that does nothing whether or not the user supplies one.

So this doesn't actually cause any functionality problems in the codebase.

Originally this `decidePolicyForNavigationResponse` was only used for the sake of gathering error codes to send to `onHttpError`. However, I added another use-case to this method and I should have moved the check for _onHttpError to surround just the relevant part.

Another improvement that should be made is that the entire section where we check the content-disposition header should be inside a check to see if _onFileDownload exists. This way unnecessary code isn't running for clients that don't supply _onFileDownload. Unlike _onHttpError, _onFileDownload should actually be null if the user doesn't supply an implementation. This is such an easy optimization that I'm kicking myself for not writing it that way to begin with. It seems so obvious now that I give the code a fresh look.

I'm not really sure why default implementations like this in `WebView.ios.tsx` exist:
```
  onHttpError = (event: WebViewHttpErrorEvent) => {
    const { onHttpError } = this.props;
    if (onHttpError) {
      onHttpError(event);
    }
  }
```

All this does is ensure that _onHttpError will never be null in the native code even if the user doesn't supply a method. Additional code was written which just results in useless CPU cycles for users that don't supply an onHttpError handler. In library code for power constrained mobile devices, when we can optimize something simply by deleting code that seems like a win-win. Less code to maintain and it is more efficient.

donysukardi:
This solution won't work with blob URL or base64 data URI, will it?

trcoffman:
> Android? 😭

@geezylucas: Android already supports file downloads and has no need for this functionality. The app doesn't need to do anything except for declare the right permissions to be able to support file downloads.

> This solution won't work with blob URL or base64 data URI, will it?

Blob URLs certainly not. Data URIs, i don't think this solution works for those. I haven't tested them. 

If navigating to a data URI results in a call to `decidePolicyForNavigationResponse`, and if the value of `navigationResponse.canShowMIMEType` would be false for data

The only way that a data URI would work given the current implementation is if all of these are true:
1. navigating to a data URI results in a call to `decidePolicyForNavigationResponse`
2. This check evaluates to true: `[navigationResponse.response isKindOfClass:[NSHTTPURLResponse class]]` . I doubt this is the case because there's no HTTP involved for a data URI.
3. This mimetype check works: `!navigationResponse.canShowMIMEType`

donysukardi:
Thanks for the breakdown @trcoffman. I found a workaround for data URI, by adding `"data:*"` to the list of `originWhitelist` and handle it inside `onShouldStartLoadWithRequest` instead.

shimangdian:
@trcoffman  Hello, Would you want to add any props on Android? I want to control the download process myself, such as showing the download progress bar, and adjusting the save location and so on

shimangdian:
I tried to add some code on Android, but I was unable to execute 
```
 yarn start: android 
```
because there seems to be a problem with my system environment. Can you help me see if the code works?

Rajan-Savaliya-spt:
not work in android

Issue #295, opened 2019-01-31
Downloading a URI starting with blob:http://... on android crashes the app


chaudhryjunaid:
I have a webview on Android that attempts to download an embedded pdf file via a blob URL. However, the app crashes on such attempt.

jamonholmgren:
Hi @chaudhryjunaid , can you provide a minimal reproduceable repo so we can see this happening?

marexv:
I have the same issue. Need some help with that repo @jamonholmgren?

mobitar:
On iOS, navigating to a blob via `window.location.href = blobUrl` will show the file within the webview, but I haven't found a way to get the file to actually download and save to the user's photos or files. (Using the method to create a temporary `a` tag and and then `a.click()` has no effect.)

This answer has a workaround, although not too elegant:
https://stackoverflow.com/questions/45065085/how-to-read-a-blob-data-url-in-wkwebview

In case you want to test this, you could create a blob link in JS with this:

```javascript
 downloadData(useNavigation = false) {
    var link = document.createElement('a');
    link.setAttribute('download', "foo.json");
    var url = window.URL.createObjectURL(new Blob([JSON.stringify({foo: "bar"})], {type: 'text/json'}));
    link.href = url;
    if(useNavigation) {
      window.location.href = link.href;
    } else {
      document.body.appendChild(link);
      link.click();
      link.remove();
    }
  }
```

`useNavigation` is for React Native, since `link.click()` doesn't do anything.

nyalsadiq:
I'm having the same problem, any update on this?

memelie:
I am facing the same issue! @jamonholmgren any help?

JesperBry:
I'm facing the same issue, any update on this? 
I've noticed that files represented as a blob, don't get a path and the app crashes. However it works with image-files represented as blobs, but not with pdf's.

fvidalarcon:
me sumo al post, esoy enfrentando e mismo problema...

`java.lang.IllegalArgumentException: Can only download HTTP/HTTPS URIs: blob:`

`@Override
            public void onDownloadStart(String url, String userAgent, String contentDisposition, String mimetype, long contentLength) {
                final DownloadManager.Request request = new DownloadManager.Request(Uri.parse(url));
                request.allowScanningByMediaScanner();
                request.setMimeType(mimetype);`

aszenz:
@jamonholmgren 

Here's an example to reproduce it:

```
[...]

PR #1217, opened 2020-02-19
fix(android): Handle IllegalArgumentException when Webview starts to …

Aldeee:
…download something but http/https URIs

# Summary

One of the libraries which we use in production make manipulation like following:

new Worker(URL.createObjectURL(new Blob(['function() {}']))).

This manipulation leads webview to make attempt to download script from the following uri: `blob:https://app.example.com/16c473d5-ac10-4b79-aa1b-4cee436e0379`. This attempt makes webview to crash with exception:

`IllegalArgumentExceptionandroid.app.DownloadManager$Request Can only download HTTP/HTTPS URIs: blob:https://app.example.com/16c473d5-ac10-4b79-aa1b-4cee436e0379`.

Might fix the following issue #295.

You can reproduce the similar behavior by implementing following code

```

const App = () => {
  return (
    <View style={{ flex: 1 }}>
      <WebView
        source={{
        }}
        javaScriptEnabled
        allowFileAccess
        mixedContentMode={'always'}
        domStorageEnabled
        textZoom={150}
      />
    </View>
  )
}

export default App;
```

Implement app with the code specified. Click "Download JSON" button inside the webview. App crashes.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -76,6 +76,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
+import java.lang.IllegalArgumentException;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -196,36 +197,40 @@ protected WebView createViewInstance(ThemedReactContext reactContext) {
...
-        module.setDownloadRequest(request);
+          module.setDownloadRequest(request);
 
-        if (module.grantFileDownloaderPermissions()) {
-          module.downloadFile();
+          if (module.grantFileDownloaderPermissions()) {
+            module.downloadFile();
+          }
+        } catch (IllegalArgumentException e) {
+          System.out.println("IllegalArgumentException: " + e.toString());
```
trcoffman: This code fits in with the error-handling philosophy found in the rest of the method, which is to just ignore errors quietly and just log them. In that sense, this change seems fine.

Outside the scope of this PR: It seems to me like there should be some kind of callback added like `onDownloadFailedToStart`, which would alert the client of this library that a file download has failed to start.

Aldeee: @trcoffman thanks for you comment. Good point is to add `onDownloadFailedToStart` callback. But I've fixed concrete case here which affected us with crashes in production. Callback implementation should be done separately I suppose.

Biplab2014:
When will it be released?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
Biplab2014: Are you going to release it soon?

Aldeee: We have already publish it in our own fork https://www.npmjs.com/package/@pandadoc/react-native-webview.

cristianoccazinsp:
Looking forward this fix as well. Does this fix any blob:// URL download attempt?

Aldeee:
> Looking forward this fix as well. Does this fix any blob:// URL download attempt?

This fix just filters 'blob://' url from being crashed while attempting to download. It doesn't let you download blob.

cristianocca:
Please re-open and merge. This is still an awful issue on Android.

Issue #1140, opened 2020-01-14
Cannot play sound in silent mode


baquy96:
**Bug description:**
I am using WebView to load a webpage which has an embedded audio player. It works fine when the app is in ringer mode. But does not have any sound when App is in IOS silent mode.

**Expected behavior:** Sound working in IOS silent mode

**Screenshots/Videos:**

![image](https://user-images.githubusercontent.com/13012834/72322002-dc4b1c80-36d7-11ea-9677-3d423b087184.png)

**Environment:**
 - react-native version: 0.61.5
 - react-native-webview version: 8.0.1

afilp:
I also have a similar problem, it is actually this one which may be relevant:
https://github.com/react-native-community/react-native-webview/issues/542

Unfortunately, the Bot closed that as inactive, while it is an important bug (we cannot have HTML5 sounds within the webview). I hope there is someone that knows the internals to look at it, time permitting. Thanks!

jasonmcaffee:
This issue has not been fixed.  From what I can tell, this fix may have been included at some time, but is no longer there, and setting the ignoreSilentMode property (I forget the exact name, and it's not documented) does nothing.

Can this be re-opened, or should a new issue be created?

Geneonosis:
really could use a resolution to this as I am running version 11.0.0. I am using this in a project for different "courses" 2 of the courses work fine in silent mode (there is audio!) the others don't and I am failing to see the difference in material.

PR #1218, opened 2020-02-19
Fixes Issue #1140

redpandatronicsuk:
Fixes https://github.com/react-native-community/react-native-webview/issues/1140
Based on a solution found at: https://stackoverflow.com/questions/56460362/how-to-force-wkwebview-to-ignore-hardware-silent-switch-on-ios
I changed the code found in the linked source from Swift to Objective-C, as required by this project. WARNING: I haven't used Swift before and very limited experience with Objective-C.

- For me this seems to work, but it is not the cleanest solution in my opinion.
- It might also be possible to play generated sound (i.e. using oscillator) instead of hardcoding the silent base64 mp3 data.
- Maybe ignoring silence switch should only be done if a parameter is supplied

# Summary
Ignores the silence switch when playing audio from a `WebView`. Closes issue #1140 

## Test Plan
1. Initialise a `react-native` project, add `react-native-webview` as dependency.
2. In `App.js` add
```
<WebView injectedJavaScript={`try{var audioCtx = new (window.AudioContext || window.webkitAudioContext)();
  var oscillator = audioCtx.createOscillator();
  oscillator.type = 'sine';
  oscillator.frequency.setValueAtTime(420, audioCtx.currentTime);
  oscillator.start();
  var oscillator_volume = audioCtx.createGain();
  oscillator_volume.gain.linearRampToValueAtTime(1, 0);
  oscillator.connect(oscillator_volume);
  oscillator_volume.connect(audioCtx.destination);
  }catch(e){}
  true;`} />
```
3. Launch the App
4. Toggle the silent switch on the device or simulator
5. Notice that when using this PR the silence switch is ignored, i.e. the sound from the oscillator (or any other webaudio) will be audible even when the silent switch is set to _on_ (silence) position
6. (optional) Do the same 5 steps, but using the code from the master branch instead and notice that when setting the silence switch to _on_ all webaudio will be muted

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    NA     |

## Checklist

I haven't added anything to README.md, because it is just a bugfix.

I haven't added the change to CHANGELOG.md, but suppose a line could be added warning users of the new default behaviour, i.e. silence switch is ignored.


safaiyeh:
Thanks for the PR @redpandatronicsuk Can you add an example to the example app, so it is easier to test & future testing of the feature?

ios/RNCWebView.m
```diff
@@ -1031,13 +1041,42 @@ - (void)evaluateJS:(NSString *)js
...
+}
+
 /**
  * Called when the navigation is complete.
  * @see https://fburl.com/rtys6jlb
  */
 - (void)webView:(WKWebView *)webView
   didFinishNavigation:(WKNavigation *)navigation
 {
+    [self forceIgnoreSilentHardwareSwitch:true];
```
safaiyeh: This should not be on by default, I would prefer this to be a new prop that developers have the option to turn on.

redpandatronicsuk: I agree, that is what I meant with the 3rd bullet point in the description (Maybe ignoring silence switch should only be done if a parameter is supplied).

For consistency however the behaviour of video playback should also change. As far as I can tell video always ignores the silent switch (tested with an embedded YouTube video).

@safaiyeh  What is the best way to move forward? Should I make new PRs, one for the ignoreSilentSwitch parameter and one for the behaviour of video to change. Or do it all in this PR?

safaiyeh: I would say do it all I this PR so we do not ship the forced silence.

redpandatronicsuk:
@safaiyeh I added the parameter you requested and added documentation for it.

Video, I think, still behaves differently.

Another problem I noticed is that the silence switch is initially ignored, i.e. when you first open the app with the silent switch on, the sound plays. Then you have to toggle it on and off again for it to mute the sound. That was the previous behaviour too but probably needs fixing too. What do you think?

redpandatronicsuk:
@safaiyeh  I made a demo that better explains what I am describing above here: https://github.com/redpandatronicsuk/react-native-webview-hardware-silence-switch-demo-

It shows the inconsistencies with the hardware silence switch using web audio and video. Besides the bug that the hardware switch is initially ignored, I think the behaviour of audio is correct now. For consistency, I think video should behave the same. Just to be clear, this PR hasn't changed video behaviour in any way, the demo just showcases the inconsistencies.

redpandatronicsuk:
@safaiyeh  I can see the status is still as _changes requested_, but I already made the changes and committed them. How can I change the status back to _Review required_?

safaiyeh:
hi @redpandatronicsuk thanks for making the changes! I'll check this out soon.

redpandatronicsuk:
Thanks @safaiyeh, I'll wait for you to review the changes, just wasn't sure if I submitted them correctly as I can still see the _Changes requested_ message even after making the changes.

index.js
```diff
@@ -1,4 +1,4 @@
-import WebView from './lib/WebView';
+import WebView from './src/WebView';
```
safaiyeh: Can we remove this from this PR? It is an issue just should be in a separate PR.

safaiyeh: Can we remove this from this PR? It is an issue just should be in a separate PR.

safaiyeh:
One last piece of feedback @redpandatronicsuk, also we should update `Guide.md` with some more information about this,

index.js
redpandatronicsuk: OK, I just had to make this change so I could test my code, otherwise it wouldn't build.

redpandatronicsuk:
@safaiyeh thanks for your feedback. I've made your requested changes. Please feel free to edit the text I added to `Guide.md`. Maybe we should add a section _Inconsistencies_  to `Guide.md` that highlights inconsistencies across platforms, etc. and put it under that section?

ios/RNCWebView.m
```diff
@@ -1031,13 +1041,48 @@ - (void)evaluateJS:(NSString *)js
...
+}
+
+-(void)disableIgnoreSilentSwitch
+{
+    [self evaluateJS: @"document.getElementById('wkwebviewAudio').src=null;true" thenCall: nil];
+}
+
+-(void)appDidBecomeActive
+{
+    if (_ignoreSilentHardwareSwitch) {
```
enahum, line 1063: After upgrading to 9.2.0 Xcode is complaining that `_ignoreSilentHardwareSwitch` is not defined

redpandatronicsuk: Just added the changes to `RNCWebView.h`, must have forgotten to include that in the commit.

redpandatronicsuk: The change is here: https://github.com/redpandatronicsuk/react-native-webview/blob/0eac0c2d80d0aab18b1c9a14e43c9c6593e0aa72/ios/RNCWebView.h#L62
How can I include this in this PR now?

enahum: so I guess a new PR is incoming to fix it and bump the version number

safaiyeh: Yup, submit a new PR @redpandatronicsuk :)

safaiyeh: Thanks for the instant feedback @enahum

fungilation:
Can't build with upgrade to RN-webview 9.2. Previously fine at 9.1.4. My app's build Xcode errors:
image
image

PR #1230, opened 2020-02-25
Load Android Gradle Plugin conditionally

SaeedZhiany:
# Summary

- Load Android Gradle Plugin conditionally

This wraps the Android Gradle plugin dependency in the buildscripts section of android/build.gradle in a conditional:

```
if (project == rootProject) {
    // ... (dependency here)
}
```

The Android Gradle plugin is only required when opening the project stand-alone, not when it is included as a dependency. By doing this, the project opens correctly in Android Studio, and it can also be consumed as a native module dependency from an application project without affecting the app project (avoiding unnecessary downloads/conflicts/etc).

for more info, you can refer to [this thread](https://github.com/facebook/react-native/pull/25569) and especially [this comment.](https://github.com/facebook/react-native/pull/25569#issuecomment-532504277)

- Also moved getExtOrDefault to buildScript block to able to use everywhere in the file

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅     |

## Checklist


SaeedZhiany:
@jamonholmgren @Titozzz 

Please review and merge this

SaeedZhiany:
@jamonholmgren @Titozzz 

Please review and merge this

SaeedZhiany:
@jamonholmgren @Titozzz 

Please review and merge this

jamonholmgren:
Can someone help us test this branch? We want to make sure we don't break people's code.

SaeedZhiany:
@jamonholmgren 

This PR has no break change for users, I successfully applied the same config on many other libraries, like react-native-firebase.
You can check the link of the thread mentioned in first comment. I have references all my same PRs to that thread.

SaeedZhiany:
@jamonholmgren 

As I said in [this comment](https://github.com/react-native-community/react-native-webview/pull/1230#issuecomment-624821427), the changes are not actually a break-change and users' project does not affect, it was not necessary to upgrade library's major version, it could be published as a patch version.

however, you may want to update the changelog description and remove the break-change section.

Thanks for merging 👍

mikehardy:
I thought I commented on this elsewhere but @SaeedZhiany kotlin build stuff doesn't work with the conditional build style for some reason This is most likely the root cause of #1407 which is a build breaker for everyone as far as I can tell

jamonholmgren:
Apologies for the broken builds, everyone. Unfortunately our CI didn't catch this. Thanks to @Titozzz for the quick fix.

PR #1286, opened 2020-04-01
[reopen] injectedJavaScriptBeforeContentLoaded should run even if messaging is…

SRandazzo:
# Summary

I previously implemented injectedJavaScriptBeforeContentLoaded here:
https://github.com/react-native-community/react-native-webview/pull/1038

This change makes it so that this prop/API works even if the webview has messaging disabled.

I don't remember this being a conscious choice to disable injectedJavaScriptBeforeContentLoaded when messaging was not enabled since the two props do not depend on one another, so probably was an oversight on my end.

Just caught this because I am now running the webview without messaging enabled

## Test Plan

Run the same test described here:
https://github.com/react-native-community/react-native-webview/pull/1038

and do not provide any onMessage prop

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    N/A     |

## Checklist

- [N/A] I mentioned this change in `CHANGELOG.md`


SRandazzo:
re-opened but a few merge conflicts have arisen - will get this updated

SRandazzo:
thanks @jamonholmgren! Hope all is well

PR #1301, opened 2020-04-07
Ensure each mounted WebView binds their personal onMessage handler

ivari:
# Summary
Fixes
- https://github.com/react-native-community/react-native-webview/issues/1295
- https://github.com/react-native-community/react-native-webview/issues/1300

which were introduced in 9.1.0 after merging https://github.com/react-native-community/react-native-webview/pull/1203

This is done by binding a separate module to each WebView that is used.

Unfortunately I came up with no better way to do the uniqueness except by a random string. If anyone has any ideas, I'm all ears.
Other things I tried:
- ref, which is not easily converted to a unique string
- nativeTag, which is a private parameter on Component
- UUID, for generation of which we'd need to include a library

- Include two separate WebView components with a unique onMessage handler for each
- Send a message to React Native layer from within each WebView
- Observe correct onMessage handlers called by calling console.log

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    N/A     |
| Android |    ✅     |

## Checklist

- [ ✅ ] I have tested this on a device and a simulator
- [ N/A ] I added the documentation in `README.md`
- [ N/A ] I mentioned this change in `CHANGELOG.md`
- [ N/A ] I updated the typed files (TS and Flow)
- [ N/A ] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -398,6 +398,11 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   public void setMessagingEnabled(WebView view, boolean enabled) {
     ((RNCWebView) view).setMessagingEnabled(enabled);
   }
+
+  @ReactProp(name = "messagingModuleName")
+  public void setMessagingModuleName(WebView view, String moduleName) {
+    ((RNCWebView) view).setMessagingModuleName(moduleName);
+  }
    
   @ReactProp(name = "incognito")
   public void setIncognito(WebView view, boolean enabled) {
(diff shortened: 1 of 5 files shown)


ivari:
CC @safaiyeh

ivari:
Also seems to fix https://github.com/react-native-community/react-native-webview/issues/1295

safaiyeh:
Thanks for looking into this. I believe there is a UUID API part of Android core. https://developer.android.com/reference/java/util/UUID

I believe this would be the best approach.

ivari:
> Thanks for looking into this. I believe there is a UUID API part of Android core. https://developer.android.com/reference/java/util/UUID
> 
> I believe this would be the best approach.

I need to generate it in JS first, and then pass it to Android native code.

safaiyeh:
Got it. I would say adding a library would make me feel more confident about generating the uuid. What do you think?

ivari:
> Got it. I would say adding a library would make me feel more confident about generating the uuid. What do you think?

Sure. Just didn't want to make that decision singlehandedly off the bat. =) Will change the PR.

ivari:
> Got it. I would say adding a library would make me feel more confident about generating the uuid. What do you think?

Done. @safaiyeh

ammarahm-ed:
Facing the same issue. I hope this gets merged soon.

PR #1328, opened 2020-04-17
feat(macOS): Make podspec compatible with macOS

alloy:
# Summary

This updates the `podspec` to indicate it also works for macOS and updates the example to illustrate this.

In addition I’ve moved the common source into a `apple` dir, which for now is the pattern that we (MS) will be following for these cases. I’ve left the `ios` and `macos` dirs in tact for those that still want/need to link manually, but this might be something to cleanup in the future or merge all targets into a single Xcode project.

## Test Plan

Screenshot 2020-04-17 at 19 51 20


alloy:
cc @tom-un

package.json
```diff
@@ -65,7 +65,7 @@
     "metro-react-native-babel-preset": "0.54.1",
     "react": "16.9.0",
     "react-native": "0.61.5",
-    "react-native-macos": "0.60.0-microsoft.49",
+    "react-native-macos": "0.60.0-microsoft.73",
     "react-native-windows": "^0.61.0-beta.58",
     "rnpm-plugin-windows": "^0.5.1-0",
     "semantic-release": "15.13.24",
```
tom-un, line 71: The new `apple/` folder needs to be added to the `"files": [ ]` below.

PR #1334, opened 2020-04-20
fix(Android): uuid generation issue

### LINKED ISSUE #1312: Error with latest version on Android (Crypto package not exist for RN)
So I keep getting following error:

Screenshot 2020-04-10 at 08 20 32

Which can be easily solved by adding `react-native-get-random-values` package and importing it from `index.js` of my project.

Thah being said, I don't think it's a good practive to let users deal with this kind of thing. Are UUID and node's crypto packages required to run webview? 

Thanks.

artyorsh:
# Summary

It seems to be an issue in latest releases affecting Android.
Here I use ~the [solution described by @DomiR](https://github.com/react-native-community/react-native-webview/issues/1312#issuecomment-613359818)~ the solution described in conversation.

Closes #1312

I guess that `./yarn.lock` should be also regenerated, but let's leave it as it is and do it if the PR is accepted. 

## Test Plan

I used to test it in the project I currently working on.

```
"react-native": "0.62.2",
"react-native-webview": "^9.2.1"
```

Pixel 3, Android 10

```
➜ npx envinfo --preset react-native

  System:
    OS: macOS 10.15.3
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
  Binaries:
    Node: 13.2.0 - ~/.nvm/versions/node/v13.2.0/bin/node
    Yarn: 1.19.2 - ~/.nvm/versions/node/v13.2.0/bin/yarn
    npm: 6.13.7 - ~/.nvm/versions/node/v13.2.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.4, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK:
      API Levels: 23, 28, 29
      Build Tools: 28.0.3, 29.0.3
      System Images: android-23 | Google APIs Intel x86 Atom_64, android-29 | Google APIs Intel x86 Atom, android-29 | Google APIs Intel x86 Atom_64, android-29 | Google Play Intel x86 Atom
  IDEs:
    Xcode: 11.4.1/11E503a - /usr/bin/xcodebuild
  npmPackages:
    react: 16.11.0 => 16.11.0
    react-native: 0.62.2 => 0.62.2
```

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

- [] I added the documentation in `README.md`

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "@types/jest": "24.0.18",
     "@types/react": "16.8.8",
     "@types/react-native": "0.60.11",
-    "@types/uuid": "^7.0.2",
     "@typescript-eslint/eslint-plugin": "2.1.0",
     "@typescript-eslint/parser": "2.1.0",
     "babel-eslint": "10.0.3",
--- a/package.json
+++ b/package.json
@@ -34,8 +34,7 @@
   },
   "dependencies": {
     "escape-string-regexp": "2.0.0",
-    "invariant": "2.2.4",
-    "uuid": "^7.0.3"
+    "invariant": "2.2.4"
   },
   "devDependencies": {
     "@babel/core": "7.4.5",
--- a/src/WebView.android.tsx
+++ b/src/WebView.android.tsx
@@ -19,7 +19,6 @@ import {
   createOnShouldStartLoadWithRequest,
   defaultRenderError,
   defaultRenderLoading,
-  uuid,
 } from './WebViewShared';
 import {
   WebViewErrorEvent,
@@ -41,6 +40,11 @@ const RNCWebView = requireNativeComponent(
   'RNCWebView',
 ) as typeof NativeWebViewAndroid;
 const { resolveAssetSource } = Image;
+
+/**
+ * A simple counter to uniquely identify WebView instances. Do not use this for anything else.
+ */
+let uniqueRef = 0;
 
 /**
  * Renders a native WebView.
@@ -71,15 +75,14 @@ class WebView extends React.Component<AndroidWebViewProps, State> {
     lastErrorEvent: null,
   };
 
-  uniqueRef = uuid().replace(/-/g, '');
 
   webViewRef = React.createRef<NativeWebViewAndroid>();
 
-  componentDidMount = () => {
-    BatchedBridge.registerCallableModule(this.getMessagingModuleName(), this);
-  }
+  messagingModuleName = `WebViewMessageHandler${uniqueRef+=1}`;
 
-  getMessagingModuleName = () => `WebViewMessageHandler${this.uniqueRef}`;
+  componentDidMount = () => {
+    BatchedBridge.registerCallableModule(this.messagingModuleName, this);
+  };
 
   getCommands = () => UIManager.getViewManagerConfig('RNCWebView').Commands;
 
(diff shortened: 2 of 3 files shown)


DomiR:
I just want to add two more cents to this discussion:
1. I think this PR solves the necessity to add another native module (which has a blocking/synchronous call to native, which is not recommended by react-native and will prevent debugging in chrome, see discussion here: https://github.com/facebook/react-native/issues/26705 and code https://github.com/facebook/react-native/blob/master/Libraries/BatchedBridge/MessageQueue.js#L175)

2. I don't think uuid is necessary  though, we could simple use a counter (maybe in combination with Date.now()) as the only goal is to make sure that each webview has it's own callback handlers. This suggestions would a bit faster:
```js
let webviewCounter = 0;
function uuid() { // maybe rename this to generateWebviewId
   return `${Date.now()}${webviewCounter += 1}`
   // alternative without counter
  return `${Date.now()}${Math.floor(Math.random() * 10000000)}`
}
```

alloy:
Great stuff, @artyorsh, thanks for picking this up 👌

@DomiR I see, yeah then just a numerical counter should be enough, no?

DomiR:
@alloy Yeah, something like https://github.com/reach/reach-ui/blob/da461e473c713a5b70cdac726844829509ee0828/packages/auto-id/src/index.js#L3-L9 should work until sebmarkbage comes up with version of this baked into a react directly (see https://twitter.com/sebmarkbage/status/1109243423905595394)

I've checked https://github.com/react-native-community/react-native-webview/commit/04f9fb23ba2f67a5e7dcf056e317bcf51c89585f again and the numeric id will be stringified in https://github.com/react-native-community/react-native-webview/commit/04f9fb23ba2f67a5e7dcf056e317bcf51c89585f#diff-8493f7e094f45f5a6ceeb96a98fbe85dL78, so long as nobody uses the uniqueRef otherwise we should be good with numeric counter...

artyorsh:
@DomiR Thanks for such details 👍 Currently I'm not familiar with what you've described below that much to be honest 😄
However, this optimization sounds good to me and I'm ok with modifying PR if it's needed. Should we also see the others interest on it?

alloy:
> @DomiR Thanks for such details 👍 Currently I'm not familiar with what you've described below that much to be honest 😄

I think the `uuid` function can entirely be removed. Instead just add something like this to [the Android `WebView` class](https://github.com/react-native-community/react-native-webview/blob/e10eca81da52df67a24cc4f7ffeb16819ddcb5dd/src/WebView.android.tsx):

```ts
/**
  * A simple counter to uniquely identify WebView instances. Do not use this for anything else.
  */
let UniqueRef = 0;

class WebView extends React.Component<AndroidWebViewProps, State> {
  // ...

  readonly uniqueRef = UniqueRef += 1;

  // ...
}
```

> However, this optimization sounds good to me and I'm ok with modifying PR if it's needed. Should we also see the others interest on it?

No need to wait imo.

farwayer:
There is no need uuid at all. Counter will be enought. And I think you can remove `getMessagingModuleName()` function and just do:
```ts
class WebView extends React.Component<AndroidWebViewProps, State> {
  readonly messagingModuleName = `WebViewMessageHandler${uniqueRef += 1}`;
}
```

artyorsh:
@farwayer @alloy yup. That works, just tested it and updated.

alloy:
Thanks @artyorsh!

artyorsh:
🎉

Issue #1219, opened 2020-02-19
StatusBar turns white color when WebView loads pdf link file (iOS)


VitalikParshin:
**Bug description:**
StatusBar turns white color when WebView loads pdf link file (iOs)

**To Reproduce:**
uri_example = "https://static......Privacy_Policy.pdf"
Passing any source={{ uri }} prop to WebView, Status bar become from black text color to white 
<WebView source={{ uri }}  />

**Expected behavior:**
Passing any source={{ uri }} prop to WebView, Status bar shouldn't change color

**Screenshots/Videos:**
WebView with pdf remote link file is loaded
<img width="484" alt="Screenshot 2020-02-19 at 22 05 46" src="https://user-images.githubusercontent.com/41287813/74872001-18237280-5365-11ea-8d39-aa582f3bd658.png">

**Environment:**
 - OS: iOS
 - OS version: 13.3
 - react-native version: 0.61.5
 - react-native-webview version: 8.0.2

dillonhafer:
This is also happening for me when I load a PDF from a `file:///` url. A workaround I found is to set the status bar programmatically after the WebView loads, which flickers, but is better than nothing.

```
<WebView
  source={{ uri: 'file:///...' }}
  onLoadEnd={() => {
    [10, 50, 100, 500, 1000].forEach(timeout => {
      setTimeout(() => {
        StatusBar.setBarStyle("dark-content");
      }, timeout);
    });
  }}/>

**Comment by bluedaniel** [comment]
Still searching for solutions as this is unresolved

**Comment by VitalikParshin** [comment]
Issue still exists

**Comment by arjunkomath** [comment]
+1

**Comment by mattwoberts** [comment]
Related I think to https://github.com/react-native-community/react-native-webview/issues/1580

**Comment by mattwoberts** [comment]
Just to add to this, it's not limited to PDFs, I'm finding the same issue for all sites. 

The webview seems to be interfering with the statusbar color. I have a simple react app that's just rendering a webview, with a statusbar with dark content:

```jsx
<SafeAreaView style={{flex:1, backgroundColor:'white'}}>
      <StatusBar barStyle="dark-content"/>
      <WebView
        ref={webref}
        source={{ uri: url}}
        style={{ flex: 1 }}
        onNavigationStateChange={handleNavigate}
      />
    </SafeAreaView>
```

This works. If I change that to `light-content`, I see it light for a flash, then the webview loads it's content and the statusbar returns to being dark.

[...]

Issue #1059, opened 2019-12-04
the webview is changing StatusBar.StatusBarStyle


mydesweb:
react-native-webview should never change StatusBarStyle automatically . On iOS I set StatusBar.setBarStyle('dark-content') as default for the screen but after a while (tap some links inside the webview) the StatusBarStyle is changed automatically to 'light-content'

 In older versions of react-native-webview I don't remember to see this kind of issue

ruslanaplid:
I confirm
onMessage handler comming event.nativeEvent.data.type == 'win_height' and after StatusBarStyle changed automatically to 'light-content'

Issue #735, opened 2019-07-22
Statusbar turns white when the webview is loaded on iOS 12


partiellkorrekt:
**Bug description:**
PR #544 introduces a bug where the statusbar turns to UIStatusBarStyleLightContent (white text) when mounting a webview with a Youtube-video inside

**To Reproduce:**
- Turn the statusbar to default (dark text) using the react-native Statusbar-component
- Mount a webview with an embedded Youtube-video

**Expected behavior:**
Mounting a webview should not affect the statusbar

**Environment:**
 - OS: iOS
 - OS version: 12.2
 - react-native version: 0.60.4
 - react-native-webview version: 6.3.1

**What I can tell so far:**
- The following line (RNCWKWebView.m, line: 84) registers an event-handler that should only be called when entering/leaving fullscreen-mode:
```
[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(toggleFullScreenVideoStatusBars) name:@"_MRMediaRemotePlayerSupportedCommandsDidChangeNotification" object:nil];
````
- Using NSlog-statements I can see that this event-handler is also called when the webview has finished loading
- This event-handler assumes the webview has just entered fullscreen and turns the statusbar to UIStatusBarStyleLightContent

**Comment by partiellkorrekt** [comment]
As a Workaround adding `_isFullScreenVideoOpen = YES;` to `initWithFrame` in the same class fixes the behavior. But I think it would be better so see, why this event-handler gets called initially.

**Comment by jvandenaardweg** [comment]
I'm experiencing the same thing, but without any youtube video embedded.

I've tracked this down to the use of a chat service called Drift. I've used a website from Drift's customer page to demonstrate the problem

```javascript
// This works fine, statusbar stays intact
<WebView
  ref={this.webviewRef}
  source={{ uri: 'https://www.google.com/' }}
  startInLoadingState={true}
  sharedCookiesEnabled={true}
  thirdPartyCookiesEnabled={true}
  useWebKit
  incognito={false}
  decelerationRate="normal"
/>
```

```javascript
// This makes the statusbar text to become white
<WebView
  ref={this.webviewRef}
  source={{ uri: 'https://www.lionbridge.com/' }}
  startInLoadingState={true}
  sharedCookiesEnabled={true}
  thirdPartyCookiesEnabled={true}
  useWebKit
  incognito={false}
  decelerationRate="normal"
/>
```

No YouTube embeds involved...

[...]

PR #1336, opened 2020-04-21
add status bar prop

s-sushkevich:
# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -64,6 +64,7 @@ This document lays out the current public properties and methods for the React N
 - [`allowsLinkPreview`](Reference.md#allowsLinkPreview)
 - [`sharedCookiesEnabled`](Reference.md#sharedCookiesEnabled)
 - [`textZoom`](Reference.md#textZoom)
+- [`autoManageStatusBarEnabled`](Reference.md#autoManageStatusBarEnabled)
 
 ## Methods Index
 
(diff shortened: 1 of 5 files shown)


docs/Reference.md
```diff
@@ -1127,6 +1128,18 @@ Example:
 
 `<WebView textZoom={100} />`
 
+### `autoManageStatusBarEnabled`
+
+If the user watch fullscreen video in WebView then StatusBar will be light. The default value is `true`.
```
trcoffman, line 1133: Perhaps 'If the user watches fullscreen video in WebView then StatusBar color will be light.' might be better,

docs/Reference.md
```diff
@@ -1127,6 +1128,18 @@ Example:
...
+
+| Type   | Required | Platform |
+| ------ | -------- | -------- |
+| number | No       | iOS      |
```
trcoffman, line 1137: Seems like the type should be boolean not number right?

docs/Reference.md
```diff
@@ -1127,6 +1128,18 @@ Example:
 
 `<WebView textZoom={100} />`
 
+### `autoManageStatusBarEnabled`
```
trcoffman, line 1131: Perhaps the flag should be called something like autoStatusBarStyle instead, since it references the `setStatusBarStyle` method.

trcoffman:
Could you please update the description comment for the PR a bit?

jamonholmgren:
@s-sushkevich Could you resolve conflicts and also update based on @trcoffman's requests?

helloiamlukas:
Any updates on this? There are multiple open issues / requests regarding this problem. 

See here #1219, here #1059, here #914 and here #735

helloiamlukas:
@jamonholmgren @Titozzz any updates on this? 
as far as I can see, this is the same as #914 - unfortunately none of these are merged yet 😢

Bardiamist:
`autoManageStatusBarEnabled` should de disabled by default

PR #1371, opened 2020-05-08
fix(Android): Fix onLoadProgress callback when navigating to the previous page

hojason117:
# Summary

#### Step to reproduce:

1. Load any page
1. Click on a link to navigate to another page
2. Click the back button on an Android phone to navigate back

#### Observe:

`onLoadProgress` callback only gets called once with the progress around `0.1`

#### Desire:

`onLoadProgress` is called multiple times with the progress starting from a lower number all the way to `1`

### The cause:

`shouldOverrideUrlLoading` is not called for the previous page when user presses the back button on phone, therefore `activeUrl` is not pointing to the correct url. This causes `TopLoadingProgressEvent` not getting dispatched correctly.

### The fix:

 Set the correct `activeUrl` before executing the `goBack` command so that `onLoadProgress` could reflect the complete progress when user navigates to the previous page.

#### Before

Screen Shot 2020-05-07 at 8 53 20 PM

#### After

Screen Shot 2020-05-07 at 8 55 00 PM

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -29,6 +29,7 @@ import android.webkit.JavascriptInterface;
 import android.webkit.PermissionRequest;
 import android.webkit.URLUtil;
 import android.webkit.ValueCallback;
+import android.webkit.WebBackForwardList;
 import android.webkit.WebChromeClient;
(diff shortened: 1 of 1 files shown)


hojason117:
Closing this as I figured out a better way to tackle this bug.
https://github.com/react-native-community/react-native-webview/pull/1373

PR #643, opened 2019-06-15
[Android] Filter out extra onLoadProgress calls; add url to onLoadProgress

tyleralves:
# Summary
Fixes #642
This pr:
1. Adds url property to the `onLoadProgress` event
2. Filters out extra `onLoadProgress` calls 

## Screenshots:
**Before** (extra sequence of `onLoadProgress`):
Screen Shot 2019-06-14 at 5 50 39 PM

**After** (no more extra `onLoadProgress` sequence):
Screen Shot 2019-06-14 at 6 06 24 PM

## Implementation
`shouldOverrideUrlLoading` reliably fires with the proper url before any page loading occurs, so I use it to set an `activeUrl` variable. Then I filter out calls to our JS `onLoadProgress` event that aren't associated with the `activeUrl`.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅     |

## Checklist

- [X] I have tested this on a device and a simulator
- [ ] I added the documentation in `README.md`
- [ ] I mentioned this change in `CHANGELOG.md`
- [ ] I updated the typed files (TS and Flow)
- [ ] I added a sample use of the API in the example project (`example/App.js`)

PR #1373, opened 2020-05-09
Improve onLoadProgress consistency

hojason117:
# Summary

Currently `onLoadProgress` callback is very inconsistent, depend on the way the url is loaded, the callback might not be called with progress from a low number all the way to `1`.
Many of the inconsistencies come from #643 and the order of the webview callbacks(`onPageStarted`, `onPageFinished`, `shouldOverrideUrlLoading`, `onProgressChanged`) are called.

#### Normal load

Screen Shot 2020-05-08 at 9 42 25 PM

#### Navigate using `goBack`

Screen Shot 2020-05-08 at 9 44 40 PM

#### Pages that have Javascript changing the url

Screen Shot 2020-05-08 at 9 43 34 PM

## Original issue

Details of the original [issue](https://github.com/react-native-community/react-native-webview/pull/642) #643 was trying to solve:
We always return true to `shouldOverrideUrlLoading` so that we can asynchronously decide whether to load the url or not. But the first overridden synchronous load also creates additional call to `onLoadProgress`. We want to filter out those calls.

Screen Shot 2020-05-08 at 9 49 44 PM

## The new solution

This commit uses a different way to tackle with the original [issue](https://github.com/react-native-community/react-native-webview/pull/642) and gets rid of the inconsistencies.

#### Solution
Use a variable `isOverriddenLoad` to mark the first load(which is going to be overridden), and don't call `onLoadProgress` on the load. Only call `onLoadProgress` if we later in Javascript side decide to load this url. 

#### Outcome

`onLoadProgress` is called consistently from a low progress all the way to `1`, no matter how the url is loaded.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅     |

## Checklist


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -109,7 +109,7 @@
 @ReactModule(name = RNCWebViewManager.REACT_CLASS)
 public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
-  public static String activeUrl = null;
+  public static boolean isOverriddenLoad = false;
```
trcoffman, line 112: Since you are fixing this code, it would be nice if you could also fix the abuse of the static keyword here. Object instances should not use class variables to store state. This pattern is a bug if there are ever 2 webview instances in the same app.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -618,6 +618,7 @@ public void receiveCommand(WebView root, int commandId, @Nullable ReadableArray
         if (args == null) {
           throw new RuntimeException("Arguments for loading an url are null!");
         }
+        isOverriddenLoad = false;
```
trcoffman, line 621: Perhaps this should be called `isWaitingForCommandLoadUrl`. If you look at the lifecycle of the two places this is mutated, what the variable seems to really be doing is signifying that we are waiting for `onShouldStartLoadWithRequest` to complete and to give us the COMMAND_LOAD_URL so we can proceed with loading the url.

hojason117: Changed to `waitingForCommandLoadUrl`.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
hojason117: Nice catch! Created a static class in `RNCWebView` instead, to inject the filter logic into `RNCWebViewClient` and `RNCWebChromeClient`.

hojason117:
@kaiguo Can you please help me take a look at why Windows CI is failing for me? I'm not really familiar with Windows environment. Thanks 🙏

kaiguo:
> @kaiguo Kai Guo FTE Can you please help me take a look at why Windows CI is failing for me? I'm not really familiar with Windows environment. Thanks 🙏

`C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\VC\v160\Microsoft.CppBuild.targets(390,5): error MSB8020: The build tools for Visual Studio 2017 (Platform Toolset = 'v141') cannot be found. To build using the v141 build tools, please install Visual Studio 2017 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution". [D:\a\react-native-webview\react-native-webview\node_modules\react-native-windows\Folly\Folly.vcxproj]`

Looks like some build tools were not installed correctly on the build server, it probably has nothing to do with your changes. Try re-running it?

kaiguo:
@hojason117 hmm I saw it failed two times in a row with the same error, something probably wrong with the install script, I'll look into it. You can ignore the error for now.

hojason117:
Cool, thanks a lot! @kaiguo

kaiguo:
@hojason117 Windows CI should be back to green now : )

PR #1364, opened 2020-05-06
Support React Native 0.62

jussikinnula:
Support React Native 0.62 & deprecate react-native WebView (fixed #1280)

# Summary

I created this PR to support React Native 0.62, as it doesn't anymore have WebView - so this community-driven WebView is a real must.

## Test Plan

We have tested this WebView on our React Native application, and also the example application has been updated.

### Tested with iOS

Screenshot 2020-05-06 at 18 55 45

### Tested with Android

Screenshot 2020-05-08 at 11 00 17

### Tested with Expo SDK

https://github.com/jussikinnula/react-native-webview-expo-test (see also branches)

### What's required for testing (prerequisites)?

```
yarn
yarn start:ios # Test with iOS
yarn start:android # Test with Android
```

In external projects, one can use `npm i github:jussikinnula/react-native-webview#build` or `npm i github:jussikinnula/react-native-webview#build2` (the first one doesn't support Expo, second one has that fixed as well).

### What are the steps to reproduce (after prerequisites)?

N/A

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Expo iOS |    ✅     |
| Expo Android |    ✅     |
| Expo Web |    ✅     |

Note! Expo Web doesn't actually support WebView, this new implementation works similarly as it has worked earlier by displaying a box with red border.

## Checklist

- [x] I have tested this on a device and a simulator
- [ ] I added the documentation in `README.md`
- [ ] I mentioned this change in `CHANGELOG.md`
- [X] I updated the typed files (TS and Flow)
- [X] I added a sample use of the API in the example project (`example/App.js`)

PR #1401, opened 2020-05-25
Skip windows CI while investigating

kaiguo:
# Summary

The component installation script stopped working which lead to some "component not found" build error, skipping Windows build and test steps for now to unblock other check-ins while investigating.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -46,22 +46,22 @@ jobs:
         yarn build
         yarn tsc
  
-    - name: NuGet restore
-      run: nuget restore example\windows\WebViewWindows.sln
+    # - name: NuGet restore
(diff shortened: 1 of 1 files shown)


kaiguo:
@jamonholmgren the component installation script exited without any error, but seems like the install failed for some reason, I haven't figured out exactly why yet.

But RNW upgraded to v142 build tools (which comes with the build server by default) since 0.62, so once we merge in #1364, the CI loop should work again. And I'll just remove the extra installation step after merging.

jamonholmgren:
@kaiguo I'm thinking I'll just merge #1364 tonight, so this PR won't be necessary. Thanks for doing it, though!

PR #1404, opened 2020-05-26
Update windows-ci.yml

imbsky:
You should use a tag like v1 instead of semver. Please read this I wrote before for details. https://github.com/avsm/setup-ocaml#how-to-specify-the-version


imbsky:
FTR: https://github.com/NuGet/setup-nuget/issues/7

jamonholmgren:
@imbsky I'm not sure I understand. Can you explain further?

imbsky:
What didn't you understand? I can explain that.

imbsky:
The disadvantage of specifying in semver style is that you always have to care about the version. It makes more sense to use tag style like v1.

imbsky:
The benefit of upgrading to v2 is that we have internally changed to use zstd instead of gzip, which reduces the size of the cache and reduces the time required for compression and decompression. The smaller cache size reduces the need to worry about cache thrashing. (Well, maybe it's not really needed for this repository.)

PR #1447, opened 2020-06-09
Fix incognito invoking clear cookies when not enabled

jasonkellydk:
Summary

I discovered an issue when using the ``` incognito``` prop on android. Whenever the viewmanager updates the setters it will automatically clear all the data. But the expected behaviour should be when the flag is enabled

opening the example project and using the incognito prop

Set the incognito prop on Android

## Compatibility

| OS         | Implemented |
| --------| :------------: |
| iOS        |    ✅                |
| Android |    ✅                |

## Checklist

- [-] I added the documentation in `README.md`
- [-] I mentioned this change in `CHANGELOG.md`
- [-] I updated the typed files (TS and Flow)
- [-] I added a sample use of the API in the example project (`example/App.js`)


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -429,6 +429,11 @@ public void setMessagingModuleName(WebView view, String moduleName) {
 
   @ReactProp(name = "incognito")
   public void setIncognito(WebView view, boolean enabled) {
+    // Don't do anything when incognito is disabled
+    if (!enabled) {
+      return;
+    }
```
jamonholmgren, line 435: I think, for clarity, since we know that `enabled` is now `true` from this point onward, let's hardcode the `true | false` values of the following method calls rather than doing `!enabled`.

```javascript
view.getSettings().setAppCacheEnabled(false);

view.clearCache(true);

view.getSettings().setSavePassword(false);
view.getSettings().setSaveFormData(false);
```

jasonkellydk: Hey nice catch, I have updated the flags. Can you please verify they are set correct?

safaiyeh: @jasonkellydk the code sniper @jamonholmgren should be deleted as well.

jamonholmgren: Yup, looks good! Thanks @jasonkellydk !

jamonholmgren:
Appreciate it @jasonkellydk !

PR #668, opened 2019-06-23
Add explicit SSL error handling to RNCWebViewClient (Android)

eahrold:
# Summary

On android the JS `react-native-webview` WebView was not calling `onError` event or `renderError` when visiting a site with an invalid SSL certificate. Leaving the page blank.

This PR adds explicit handling to SSL errors by overriding 

```java
 public void onReceivedSslError(final WebView webView, final SslErrorHandler handler, final SslError error) {
```

In the [RNCWebViewManager.java -> RNCWebViewClient](https://github.com/react-native-community/react-native-webview/blob/master/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L591)

It uses descriptions from error code provided here [https://developer.android.com/reference/android/net/http/SslError.html](https://developer.android.com/reference/android/net/http/SslError.html)

This should have minimal impact, and should not affect any custom implementations. 

Possibly relates to https://github.com/react-native-community/react-native-webview/issues/259

### What's required for testing (prerequisites)?
- Setup a WebView
```javascript
import { WebView } from 'react-native-webview'
...
<WevView
  onError={({ nativeEvent }) => {
    alert(JSON.stringify(nativeEvent, null, 2))
  }}
/>
```
- On android navigate to site with invalid certificate. 

Screen Shot 2019-06-23 at 9 45 59 AM

## Compatibility

| OS         | Implemented |
| ------- | :------------: |
| Android |    ✅              |

## Checklist

- [x] I have tested this on a device and a simulator
- [ ] I added the documentation in `README.md`
- [ ] I mentioned this change in `CHANGELOG.md`
- [ ] I updated the typed files (TS and Flow)
- [ ] I added a sample use of the API in the example project (`example/App.js`)

PR #1450, opened 2020-06-10
Add SSL error handling for Android WebView

thephpjedi:
# Summary

Add SSL error handling, based on changes from PR #668 by @eahrold. Fixes security issues in apps which depend on `onError` being emitted for non-successful page loads, including SSL errors.

Navigations are cancelled and error is shown to user by default. No option is given to user to continue.

The only differences between PR #668 and this PR are:
* This PR excludes the unnecessary `SSL_MAX_ERROR` case, since it is never called per Android docs.
* This PR prefixes error descriptions with `SSL error:` to provide users and developers with more context, since some error messages do not provide sufficient context. Makes it more compatible with existing custom error screens too.

To quote the original PR:
> On android the JS `react-native-webview` WebView was not calling `onError` event or `renderError` when visiting a site with an invalid SSL certificate. Leaving the page blank.
> 
> This PR adds explicit handling to SSL errors by overriding
> 
> ```java
>  public void onReceivedSslError(final WebView webView, final SslErrorHandler handler, final SslError error) {
> ```
> 
> In the [RNCWebViewManager.java -> RNCWebViewClient](https://github.com/react-native-community/react-native-webview/blob/master/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L591)
> 
> It uses descriptions from error code provided here https://developer.android.com/reference/android/net/http/SslError.html
> 
> This should have minimal impact, and should not affect any custom implementations.

Set up a basic WebView. Optionally add `onError` handler.

```js

<WebView source={{ uri: "https://wrong.host.badssl.com" }} />
```

1. Navigate to a page which results in an SSL error, such as https://wrong.host.badssl.com

By default, the following screen is shown within the WebView:
ReactNativeWebView-SSL-error

With a custom `onError` handler, the event is sent to the handler for custom processing, if the app already shows its own custom error page. This behavior makes the change backwards compatible while still resolving the security issues.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    N/A     |
| Android |    ✅     |

## Checklist

- [`No docs needed`] I added the documentation in `README.md`
- [`There's no CHANGELOG.md? 😕`] I mentioned this change in `CHANGELOG.md`
- [`N/A`] I updated the typed files (TS and Flow)
- [`N/A`] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -10,6 +10,7 @@ import android.content.pm.PackageManager;
 import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.Manifest;
+import android.net.http.SslError;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Environment;
@@ -26,6 +27,7 @@ import android.webkit.CookieManager;
 import android.webkit.DownloadListener;
 import android.webkit.GeolocationPermissions;
 import android.webkit.JavascriptInterface;
+import android.webkit.SslErrorHandler;
 import android.webkit.PermissionRequest;
 import android.webkit.URLUtil;
 import android.webkit.ValueCallback;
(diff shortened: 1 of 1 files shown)


thephpjedi:
PR opened per https://github.com/react-native-community/react-native-webview/pull/668#issuecomment-636088631

@jamonholmgren This is the security fix PR

thephpjedi:
Thanks to both of you for review and merge.

thephpjedi:
Created PR #1466 to fix undesired behavior introduced in this PR.

Appreciate a review when possible.

PR #1454, opened 2020-06-10
Fixing commands with custom RNCWebView subclass

stigi:
# Summary

This adds the `viewName` prop to `nativeConfig` and uses it when looking up commands that are going to get performed on a custom web view.

This solves a bug that occurs with custom subclasses of `RNCWebView`. I didn't report the bug yet, but I added an example that demonstrates it to this PR.

## The issue

The short version is: The implementation of `getCommands` always checked the original `RNCWebView`, and therefore did never consider commands added to a custom subclass of `RNCWebViewManager`. Therefore we would use the wrong command indexes when executing a command. I.e. `injectJavascript` would usually be at index `1`. If a custom view manager subclass added a new command that alphabetically would come before `injectJavascript` it would push that command back by one. Without the fix we would still use the old index `1` even though points to another command now.

## Test Plan

I extended the iOS example with a case for a custom web view. There are two buttons in the example:

- "Inject Alert" that triggers an alert inside the webview via a call to `injectJavascript`
- "Show commands" that logs out the map of commands. It should contain `exampleCommand` that has been added to `CustomWebViewManager`.

To reproduce the original bug it's enough to simply not pass the `nativeConfig.viewName` prop in `Custom.tsx`. This will still render the webview, but `injectJavascript` won't work (side note: calling `webView.injectJavascript` will actually have the same effect as calling `webView.postMessage`).

Just the iOS example app.

- Checkout this PR
- `yarn install`
- `pod install --project-directory=example/ios`
- `open example/ios/example.xcworkspace` and run in Xcode (I wasn't personably able to run it from `yarn start:ios` like mentioned in Contribution.md)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| macOS     |    ✅     |
| windows     |    ✅     |

I only implemented the example for iOS though.

## Checklist

- [ ] I mentioned this change in `CHANGELOG.md` << **Where is this?**

I also updated `docs/Custom-iOS.md` but just the `nativeConfig` part. The whole file needs a revamp, but that's a know issue I guess.

### DIFF
--- a/docs/Custom-Android.md
+++ b/docs/Custom-Android.md
@@ -183,7 +183,10 @@ export default class CustomWebView extends Component {
 
   render() {
     return (
-      <WebView {...this.props} nativeConfig={{ component: RCTCustomWebView }} />
+      <WebView {...this.props} nativeConfig={{
+        component: RCTCustomWebView,
+        viewName: 'RCTCustomWebView'
+      }} />
     );
   }
 }
--- a/docs/Custom-iOS.md
+++ b/docs/Custom-iOS.md
@@ -195,6 +195,7 @@ export default class CustomWebView extends Component {
         nativeConfig={{
           component: RCTCustomWebView,
           viewManager: CustomWebViewManager,
+          viewName: 'RCTCustomWebView'
         }}
       />
     );
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -17,6 +17,7 @@ import Downloads from './examples/Downloads';
 import Uploads from './examples/Uploads';
 import Injection from './examples/Injection';
 import LocalPageLoad from './examples/LocalPageLoad';
+import Custom from './examples/Custom';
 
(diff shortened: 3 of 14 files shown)


jamonholmgren:
This looks good @stigi. Two questions:

1. Could you (or would you even want to) infer the custom component name from the `component` prop instead of having to specify another prop?
2. Custom WebView component seems like a _very_ good thing to put into the `Guide.md`. Would you be willing to author a guide like that, which talks about why you might want to do that, and gives an example? (this doesn't have to be in _this_ PR).

Thank you!

stigi:
Sorry @jamonholmgren. I got carried away by other work. I mainly opened this PR to upstream the changes from our project. I'll allocate some time to look into 1. and 2. this week. Let's not let this PR die in silence :)

PR #1466, opened 2020-06-16
fix(android): Update SSL error handling

alesandroortiz:
# Summary

Fixes issue introduced in PR #1450.

This PR updates the SSL error handling to call onReceivedError() only on top-level navigations.

This prevents iframes and other subresources from causing user-visible SSL error messages. The desired behavior is only to have top-level navigations show user-visible error messages. All other requests should be cancelled automatically with no user-visible error message.

Use any basic React Native WebView app (I have a PoC app I use to generally test RNWV behavior.)

#### iframe navigations
1. Navigate the top-level window to https://aogarantiza.com/react/webview-ssl-error-test.html
2. Tap the link in the iframe.

Observed behavior due to #1450: iframe navigation is cancelled but user-visible error is shown which covers the full page, making it seem like a top-level navigation issue.

Expected behavior, introduced with this PR: iframe navigation is cancelled and no user-visible error is shown.

#### Top-level navigations
1. Navigate the top-level window to https://aogarantiza.com/react/webview-ssl-error-test.html
2. Tap the link in the top document.

Observed behavior due to #1450: top-level window is cancelled and user-visible error is shown which covers the full page.

Expected behavior: Unchanged from #1450. (Desired behavior.)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    N/A     |
| Android |    ✅     |

## Checklist

- [`N/A`] I added the documentation in `README.md`
- [`N/A`] I mentioned this change in `CHANGELOG.md`
- [`N/A`] I updated the typed files (TS and Flow)
- [`N/A`] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -17,6 +17,7 @@ import android.os.Environment;
 import androidx.annotation.RequiresApi;
 import androidx.core.content.ContextCompat;
 import android.text.TextUtils;
+import android.util.Log;
 import android.view.Gravity;
 import android.view.View;
 import android.view.ViewGroup;
@@ -801,8 +802,9 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
     @Override
     public void onReceivedSslError(final WebView webView, final SslErrorHandler handler, final SslError error) {
+        // onReceivedSslError is called for most requests, per Android docs: https://developer.android.com/reference/android/webkit/WebViewClient#onReceivedSslError(android.webkit.WebView,%2520android.webkit.SslErrorHandler,%2520android.net.http.SslError)
         // WebView.getUrl() will return the top-level window URL.
-        // If a top-level navigation triggers this error handler, the top-level URL will be the failing URL (not the current URL).
+        // If a top-level navigation triggers this error handler, the top-level URL will be the failing URL (not the URL of the currently-rendered page).
         // This is desired behavior. We later use these values to determine whether the request is a top-level navigation or a subresource request.
         String topWindowUrl = webView.getUrl();
         String failingUrl = error.getUrl();
@@ -814,6 +816,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
         if (!topWindowUrl.equalsIgnoreCase(failingUrl)) {
           // If error is not due to top-level navigation, then do not call onReceivedError()
+          Log.w("RNCWebViewManager", "Resource blocked from loading due to SSL error. Blocked URL: "+failingUrl);
           return;
         }
 
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -801,10 +801,23 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
     @Override
     public void onReceivedSslError(final WebView webView, final SslErrorHandler handler, final SslError error) {
+        // WebView.getUrl() will return the top-level window URL.
+        // If a top-level navigation triggers this error handler, the top-level URL will be the failing URL (not the current URL).
+        // This is desired behavior. We later use these values to determine whether the request is a top-level navigation or a subresource request.
+        String topWindowUrl = webView.getUrl();
+        String failingUrl = error.getUrl();
+        
+        // Cancel request after obtaining top-level URL.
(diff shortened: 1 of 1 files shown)


thephpjedi:
Apologies for not catching this behavior in the prior PR. I figured `onSslError()` was called only for top-level navigations, but upon testing yesterday I discovered it was also called for subresources.

thephpjedi:
@jamonholmgren FYI, bug fix for PR #1450

thephpjedi:
@jamonholmgren Ping for this PR, blocking issue #1530.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -801,10 +801,23 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
 
     @Override
     public void onReceivedSslError(final WebView webView, final SslErrorHandler handler, final SslError error) {
+        // WebView.getUrl() will return the top-level window URL.
+        // If a top-level navigation triggers this error handler, the top-level URL will be the failing URL (not the current URL).
```
TheAlmightyBob, line 805: "current" is slightly confusing in this transitional context. I assume it means "last successful URL." If you initialize a WebView with a URL that has invalid SSL cert, what is the "current" URL? Just blank?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -801,10 +801,23 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
 
     @Override
     public void onReceivedSslError(final WebView webView, final SslErrorHandler handler, final SslError error) {
+        // WebView.getUrl() will return the top-level window URL.
```
TheAlmightyBob, line 804: Perhaps this comment could open with the simple observation that `onReceivedSslError` is called for subresources in addition to the top-level URL? That's key info that, like you, I had not expected (given that many other Android WebView events are not raised for subresources).

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -801,10 +801,23 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
...
+        String failingUrl = error.getUrl();
+        
+        // Cancel request after obtaining top-level URL.
+        // If request is cancelled before obtaining top-level URL, undesired behavior may occur.
+        // Undesired behavior: Return value of WebView.getUrl() may be the current URL instead of the failing URL.
         handler.cancel();
 
+        if (!topWindowUrl.equalsIgnoreCase(failingUrl)) {
+          // If error is not due to top-level navigation, then do not call onReceivedError()
+          return;
```
TheAlmightyBob, line 817: I assume the resulting behavior is just a blank space where the nested content would have been?

It would be nice to have _some_ indication of what happened. If not via new event, perhaps at least just log a message here so it can be picked up by logcat?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
thephpjedi: "current" is meant to say "currently rendered page", which could be a blank page if none is currently rendered. Will update shortly.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
thephpjedi: Will update shortly.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
thephpjedi: The behavior is controlled by the Android WebView. For visible resources (e.g. iframes), I believe it would be either blank or show an embedded SSL error page, but this may vary based on Android WebView version. For non-visible resources (e.g. scripts), nothing is visible.

I'll add quick log message.

TheAlmightyBob (approved):

Titozzz (approved):

alesandroortiz:
@Titozzz Thanks for review and merge! Apologies again for introducing this bug as part of the prior security fix.

clamiax:
I rather would like to have this as default behaviour or maybe as an option:

```
handler.proceed();
super.onReceivedSslError(webView, handler, error);
``

It's just my preference.

parasjain10:
image

React Native version: 0.76.1
Reaat native web view: 13.12.4

PR #1479, opened 2020-06-23
Synchronous onShouldStartLoadWithRequest for Android

dvicory:
# Summary

We (myself and @trcoffman) discovered a very interesting bug in our WebView app. Inside our app, there was a link that would HTTP redirect to a different origin, and then immediately manipulate history (to remove an authentication token from the URL and to navigate). This would cause the user to be stuck going back (you would go back and the history/location manipulation would happen again). The only way to get back was to tap the back button quickly multiple times in succession. Essentially, WebView added extra entries to the history stack that should not be there.

We tracked it down to the style of overriding loading. Essentially, every navigation on Android is suppressed in `shouldOverrideUrlLoading`, and an event is sent to JS to decide to navigate or not. If the navigation should happen, it causes a new page load. With this technique, the WebView seems to lose "context" about redirects and can improperly add to the history stack under the right conditions. Interestingly, this asynchronous approach also causes a lot of issues with debugging the WebView with the network inspector. Initiators are incorrect and the redirects aren't clearly exposed, since every load is a new page load.

The fix is to be able to synchronously decide in `shouldOverrideUrlLoading` whether to load or not. iOS does this approach by locking, sending an event, and waiting for the decision to come in asynchronously, and then unlocking. We tried a similar approach for Android but found it to deadlock (since it seems the WebView callbacks run on the same thread). The only way out was to use React Native's new JSI to be able to synchronously call into the JS runtime for the `onShouldStartLoadWithRequest` decision and have that result immediately for the WebView. I alluded to this [in earlier comments](https://github.com/react-native-community/react-native-webview/pull/991#issuecomment-580047127). We've known that doing this asynchronous style could be problematic, but never found a use case where it caused an issue until now. All said, this requires the addition of a native C++ module (JNI) to be built to be able to access the runtime using JSI.

The [Android docs warn against doing something very similar to what was being done](https://developer.android.com/reference/android/webkit/WebViewClient#shouldOverrideUrlLoading(android.webkit.WebView,%20android.webkit.WebResourceRequest)).

> Note: Do not call WebView#loadUrl(String) with the request's URL and then return true. This unnecessarily cancels the current load and starts a new load with the same URL. The correct way to continue loading a given URL is to simply return false, without calling WebView#loadUrl(String).

We found that with debugging, there is no JS runtime. This makes sense since it runs inside the Chrome process. As a fallback, the old asynchronous deny-then-load approach is kept when there is no JS runtime. This could allow for buggy behavior like we saw, but will allow users to debug their `onShouldStartLoadWithRequest`.

Exploiting this behavior is non-trivial. It requires a link that redirects to a different origin, that then immediately manipulates browser history (`window.location` plus `history`). I've attached a video of behavior before our fix. Notice that tapping back doesn't go back to the original screen with the menu of links. It requires multiple taps to get back.

ezgif-3-c932b0bd6e92

This is after our fix:

ezgif-3-27d2f00c7109

We've tested this fix quite a bit already in our app. I don't believe adding an example case makes sense.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

This requires react-native 0.60 or later for JSI to be in the expected location. We could not find any solid compatibility information within the repository, other than a peer dependency of `react-native >= 0.60`.

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -34,6 +34,15 @@ def getExtOrIntegerDefault(name) {
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
+task extractSo(type: Copy) {
+  from zipTree("${project.rootDir}/../node_modules/react-native/android/com/facebook/react/react-native/0.62.2/react-native-0.62.2.aar")
+  into "${buildDir}/_libraries/"
+  include "jni/**/libjscexecutor.so"
+
+}
+preBuild.dependsOn(extractSo)
+
+
 android {
   compileSdkVersion getExtOrIntegerDefault('compileSdkVersion')
   buildToolsVersion getExtOrDefault('buildToolsVersion')
@@ -45,6 +54,11 @@ android {
     ndk {
       moduleName "reactnativecommunity-webview"
       ldLibs "log"
+    }
+    externalNativeBuild {
+      ndkBuild {
+        arguments "BUILD_DIR=${buildDir}"
+      }
     }
   }
   buildTypes {
@@ -149,3 +163,4 @@ dependencies {
   implementation 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
 }
+
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -42,6 +42,10 @@ android {
     targetSdkVersion getExtOrIntegerDefault('targetSdkVersion')
     versionCode 1
     versionName "1.0"
+    ndk {
+      moduleName "reactnativecommunity-webview"
+      ldLibs "log"
+    }
   }
   buildTypes {
     release {
@@ -54,6 +58,16 @@ android {
   compileOptions {
     sourceCompatibility JavaVersion.VERSION_1_8
     targetCompatibility JavaVersion.VERSION_1_8
+  }
+  sourceSets {
+    main {
+      jni.srcDirs = ["src/main/cpp"]
+    }
+  }
+  externalNativeBuild {
+    ndkBuild {
+      path file('src/main/jni/Android.mk')
+    }
   }
 }
 
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ Pods/
 .gradle
 local.properties
 lib/android/src/main/gen
+.cxx/
 
 # node.js
 #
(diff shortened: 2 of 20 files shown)


dvicory:
We're investigating an issue with a JS-level thrown error in the `onShouldStartLoadWithRequest` causing a native app crash. That was not the intent based on the code (where it should be a C++ exception that then creates a Java exception), so please defer reviewing this until we've figured out what's going on.

dvicory:
We figured out that there was an ABI compatibility issue. `JSError` could not be caught in C++. In fact, any JS error would go into the default catch all block in C++, which we (safely) assume is non-recoverable.

While there is a new commit "fixing" this, it makes some critical assumptions. As a result, this is **not ready** to be reviewed or merged. However, apps that wish to have this functionality could use this PR as a template.

We require a patched `jsi.h`, to fix missing `JSI_EXPORT`. This [will be fixed in react-native 0.63](https://github.com/facebook/react-native/commit/84adc85523770ebfee749a020920e0b216cf69f8). We also have to link against `libjscexecutor` rather than build `jsi.cpp`. To do that, we have to extract the react-native AAR file, which assumes a certain version (0.62.2) and at a certain location. The following will need to be done to get this PR in a ready state for mass consumption:

- More sophisticated code to locate react-native node_module in gradle, in a version-agnostic way
- Opt-in feature/build flag
- Testing with older versions of react-native
- Create an `Android.mk` that declares `libjscexecutor` as a prebuilt instead of using `LOCAL_LDLIBS` to get rid of linker warnings
- Ensure compatibility with hermes as well as JSC

android/build.gradle
```diff
@@ -34,6 +34,15 @@ def getExtOrIntegerDefault(name) {
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
+task extractSo(type: Copy) {
+  from zipTree("${project.rootDir}/../node_modules/react-native/android/com/facebook/react/react-native/0.62.2/react-native-0.62.2.aar")
```
jamonholmgren, line 38: Does this line lock us to a specific version of React Native?

trcoffman: @jamonholmgren This is the line that lead to me closing the PR for now. This line does tie you to a specific version of react-native and a specific directory structure. It's appropriate for our app but not for upstream.

I'm working now on changes to build.gradle that will not tie you to any specific version of RN. We also need to test this with other versions of react native like 0.60 and 0.61.

jamonholmgren: Sounds good!

dvicory:
@jamonholmgren and lurkers, thanks for taking the time to review this! We have a completely different approach that avoids all this C++/JNI/JSI nastiness (and no more ABI compatibility issues). Please see #1590.

Issue #1090, opened 2019-12-19
Android hard crash with blob:// links


cristianoccazinsp:
**Bug description:**

Getting a hard crash on Android (see log below) when the webview tries to open a blob link. 

In addition, the `onShouldStartLoadWithRequest ` prop is not called which I'm also using to stop blob links from even running.

```
java.lang.IllegalArgumentException Can only download HTTP/HTTPS URIs: blob:https://app.zinspector.com/b20a08ad-41c7-4d93-bd14-2280b9d96d2c 
    DownloadManager.java:614 android.app.DownloadManager$Request.<init>
    RNCWebViewManager.java:199 com.reactnativecommunity.webview.RNCWebViewManager$2.onDownloadStart
    PG:537 sV.a
    PG:25 aHF.handleMessage
    Handler.java:106 android.os.Handler.dispatchMessage
    Looper.java:164 android.os.Looper.loop
    ActivityThread.java:7000 android.app.ActivityThread.main
    Method.java:-2 java.lang.reflect.Method.invoke
    RuntimeInit.java:441 com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run
    ZygoteInit.java:1408 com.android.internal.os.ZygoteInit.main
```

**To Reproduce:**

Test the webview with any page that as a blob download.

**Expected behavior:**

First, it shouldn't crash with a blob:// link. Secondly, at least `onShouldStartLoadWithRequest ` should be called so we can prevent the link from running.

**Screenshots/Videos:**

**Environment:**
 - OS: Android
 - OS version: 8.1, 9, 10
 - react-native version: 0.61.5
 - react-native-webview version: 7.6.0

manuelabarca:
The same issue but only in Android, when return blob link the webview crash and close this.

vpankov:
same issue

vpankov:
Not a bug fix but I hope that it will be helpful.

I used webView **postMessage** api, to send params from web application to native application and downloaded files using 

- **rn-fetch-blob** for Android, 
- **rn-fetch-blob**, **react-native-fs**, **react-native-share** for iOS.

**1. Web application side** (if you haven't access for web app code, you can try to use injectedJavaScript for this)
1.1 Check is `window.ReactNativeWebView` exists to verify that it's native app context and prevent crashing for regular web browsers.
1.2 Send data that you need to process api from native app, using `window.ReactNativeWebView.postMessage` 
1.3 "return null" to prevent blob link processing in webview
[...]

PR #1480, opened 2020-06-23
feat(android): WebView crash handling

cristianoccazinsp:
# Summary

On Android, the WebView process may get killed/crash. When this happens, if not handled appropriately, the whole app gets terminated in a quite brutal way (not even error reporting frameworks seemed to report this crash).

The fix: implement a handler for `onRenderProcessGone` (SDK >= 26) and return `true`.

The implementation also adds a new `onCrash` event so a webview crash can be handled on user land.

TODO: Review implementation for iOS, although it is most likely not needed.

## Test Plan
Tested on Motorola G5 and Google Pixel 2

Device / simulator

Test a webview that may run out of memory or render a extremely large image.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [X] I mentioned this change in `CHANGELOG.md`

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -927,6 +927,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
         }
 
         WritableMap event = createWebViewEvent(webView, webView.getUrl());
+        event.putBoolean("didCrash", detail.didCrash());
 
         dispatchEvent(
           webView,
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -71,7 +71,7 @@ import com.reactnativecommunity.webview.events.TopLoadingProgressEvent;
 import com.reactnativecommunity.webview.events.TopLoadingStartEvent;
 import com.reactnativecommunity.webview.events.TopMessageEvent;
 import com.reactnativecommunity.webview.events.TopShouldStartLoadWithRequestEvent;
-import com.reactnativecommunity.webview.events.TopCrashEvent;
+import com.reactnativecommunity.webview.events.TopRenderProcessGoneEvent;
 
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -578,7 +578,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
     export.put(TopShouldStartLoadWithRequestEvent.EVENT_NAME, MapBuilder.of("registrationName", "onShouldStartLoadWithRequest"));
     export.put(ScrollEventType.getJSEventName(ScrollEventType.SCROLL), MapBuilder.of("registrationName", "onScroll"));
     export.put(TopHttpErrorEvent.EVENT_NAME, MapBuilder.of("registrationName", "onHttpError"));
-    export.put(TopCrashEvent.EVENT_NAME, MapBuilder.of("registrationName", "onCrash"));
+    export.put(TopRenderProcessGoneEvent.EVENT_NAME, MapBuilder.of("registrationName", "onRenderProcessGone"));
     return export;
   }
 
@@ -905,12 +905,12 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
     @TargetApi(Build.VERSION_CODES.O)
     @Override
-    public boolean onRenderProcessGone(WebView view, RenderProcessGoneDetail detail) {
+    public boolean onRenderProcessGone(WebView webView, RenderProcessGoneDetail detail) {
         // WebViewClient.onRenderProcessGone was added in O.
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.O) {
             return false;
         }
-        super.onRenderProcessGone(view, detail);
+        super.onRenderProcessGone(webView, detail);
 
         if(detail.didCrash()){
           Log.e("RNCWebViewManager", "The WebView rendering process crashed.");
@@ -919,11 +919,18 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
           Log.w("RNCWebViewManager", "The WebView rendering process was killed by the system.");
         }
 
-        WritableMap event = Arguments.createMap();
-        event.putBoolean("didCrash", detail.didCrash());
+        // if webView is null, we cannot return any event
+        // since the view is already dead/disposed
+        // still prevent the app crash by returning true.
+        if(webView == null){
+          return true;
+        }
+
+        WritableMap event = createWebViewEvent(webView, webView.getUrl());
+
         dispatchEvent(
-          view,
-          new TopCrashEvent(view.getId(), event)
+          webView,
+          new TopRenderProcessGoneEvent(webView.getId(), event)
         );
 
         // returning false would crash the app.
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -17,6 +17,7 @@ import android.os.Environment;
 import androidx.annotation.RequiresApi;
 import androidx.core.content.ContextCompat;
 import android.text.TextUtils;
+import android.util.Log;
 import android.view.Gravity;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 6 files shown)


cristianoccazinsp:
Thoughts?

scriptspry:
+100. This is very necessary to handle webview crashes gracefully. The diff looked pretty clean to me (please keep in mind I haven't reviewed any pull-requests though).

Kindly review and merge this.

scriptspry:
It will be terrific if we can add support for iOS as well. Probably, by using the method

[`webViewWebContentProcessDidTerminate(_ webView: WKWebView)`](https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455639-webviewwebcontentprocessdidtermi) ?

cristianoccazinsp:
@scriptspry I believe the terminate callback on iOS is already implemented here https://github.com/react-native-community/react-native-webview/blob/master/apple/RNCWebView.m#L974

However, that fires every time the webview terminates, not just on crashes, and doesn't let you "prevent the app crash" like the android version. For this reason, I haven't considered iOS yet on this PR.  It seems like it needs some more research and testing so it is probably better for it to be on another PR.

jamonholmgren:
@cristianoccazinsp Would you mind resolving conflicts first?

cristianoccazinsp:
Resolved from the web since it was just a readme issue. Can you take a look if the readme update was appropriate? Looks like I may be missing a reference so the links work, but I couldn't figure out how's the readme getting updated.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -899,6 +903,33 @@ public void onReceivedHttpError(
...
+        super.onRenderProcessGone(view, detail);
+
+        if(detail.didCrash()){
+          Log.e("RNCWebViewManager", "The WebView rendering process crashed.");
+        }
+        else{
+          Log.w("RNCWebViewManager", "The WebView rendering process was killed by the system.");
+        }
+
+        WritableMap event = Arguments.createMap();
```
TheAlmightyBob, line 922: Why not use `createWebViewEvent(webView, webView.getUrl())`?

android/src/main/java/com/reactnativecommunity/webview/events/TopCrashEvent.kt
```diff
@@ -0,0 +1,25 @@
+package com.reactnativecommunity.webview.events
+
+import com.facebook.react.bridge.WritableMap
+import com.facebook.react.uimanager.events.Event
+import com.facebook.react.uimanager.events.RCTEventEmitter
+
+/**
+ * Event emitted when the WebView's process has crashed
+ */
+class TopCrashEvent(viewId: Int, private val mEventData: WritableMap) :
```
TheAlmightyBob, line 10: If this is going to be an Android-only event anyway, I would suggest actually using the more-specific name (`RenderProcessGone`) rather than `TopCrash`/`onCrash`. (it could be a bit confusing that the `onCrash` event data contains a boolean variable `didCrash` which may be _false_...)

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
cristianoccazinsp: You are right, I will look into it.

android/src/main/java/com/reactnativecommunity/webview/events/TopCrashEvent.kt
cristianoccazinsp: I think there's a subtle difference (which I wasn't sure how to handle). Both cases are crashes (I think), but one was a full crash, and another one just the OS killing the process.  It may be confusing when the webview process is actually done (e.g., the webview being closed), which this event does not fire.  I will take another look and review precisely when these 3 cases/events may happen.

cristianoccazinsp: What about just renaming `didCrash` to something else, or not including it at all. 

https://developer.android.com/reference/android/webkit/RenderProcessGoneDetail

From the docs, it really seems like the reason is always a crash, but it just indicates if it was observed or not. Thoughts?

cristianoccazinsp: Nevermind, it makes sense as you said to just rename it to use the same android native event. This way we will be covered for iOS in the future.

cristianoccazinsp:
Pushed commits with the above recommendations.

android/src/main/java/com/reactnativecommunity/webview/events/TopCrashEvent.kt
TheAlmightyBob: I think `didCrash` should still be included. I'm not entirely sure of its purpose either (in my experience it's always false), but it doesn't hurt and is one line for you to add while you're here (since you already have it in the docs and types)

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -919,11 +919,18 @@ public boolean onRenderProcessGone(WebView view, RenderProcessGoneDetail detail)
...
+        // since the view is already dead/disposed
+        // still prevent the app crash by returning true.
+        if(webView == null){
+          return true;
+        }
+
+        WritableMap event = createWebViewEvent(webView, webView.getUrl());
+
         dispatchEvent(
-          view,
```
TheAlmightyBob: Just in case my other comment got lost since it was labeled "outdated," please bring back `didCrash`

cristianoccazinsp: I'm sorry, I did see your comment and I thought I had kept the `didCrash` value, but I didn't.  Adding it back now.

cristianoccazinsp: Should be done now.

TheAlmightyBob: Thanks! Looks good

jasongaare:
LGTM! Any chance this would prevent the hard crash for a `blob://` URL described here: https://github.com/react-native-community/react-native-webview/issues/1090

cristianoccazinsp:
> LGTM! Any chance this would prevent the hard crash for a `blob://` URL described here: #1090

I don't think so. That crash is entirely on Java/native land and not related to the webview process dying.  Fix shouldn't be too hard though.

jasongaare:
> That crash is entirely on Java/native land and not related to the webview process dying. Fix shouldn't be too hard though.

Good observation, I've opened a new PR to address the crash: https://github.com/react-native-community/react-native-webview/pull/1546

cristianoccazinsp:
> > That crash is entirely on Java/native land and not related to the webview process dying. Fix shouldn't be too hard though.
> 
> Good observation, I've opened a new PR to address the crash: #1546

Awesome, looking forward to the update!

florian-milky:
Thanks for this, I can't wait for the release :)
It would be nice to also handle a related feature: https://developer.android.com/guide/webapps/managing-webview#renderer-importance

cristianoccazinsp:
> Thanks for this, I can't wait for the release :)
> It would be nice to also handle a related feature: https://developer.android.com/guide/webapps/managing-webview#renderer-importance

@florian-milky how would that help RN apps? Isn't the default priority already the highest?

florian-milky:
> @florian-milky how would that help RN apps? Isn't the default priority already the highest?

I'm sorry you're right, I thought the default priority was the middle one

cristianoccazinsp:
Any ideas when this can be merged? Would love to get it merged soon for our own product next release.

scriptspry:
Tagging and requesting @Titozzz to kindly review this.

Titozzz:
I'm trying to get back to Open Source. Thanks for contributing. Sorry about the delay
❤️

florian-milky:
@cristianoccazinsp I tried to use this but without any luck.
2 different emulators and a samsung s9, but onRenderProcessGone is not called and the whole app crashes still.

cristianoccazinsp:
Interesting, perhaps it is another type of crash? Do you have any sample code to forcibly trigger the crash? I only tried it on a very specific crash I was having, but perhaps the webview doesn't behave the same with every crash.

cristianoccazinsp:
@florian-milky I've just tried it with a fresh install from npm and it is working fine for me on a google pixel 2.  

My use case was a webview with an oddly large canvas that for some reason causes a crash on Android 10.  Keep in mind this change requires API 26 (Android 8) or above.

florian-milky:
I use a webworker in the webview to trigger the memory consumption. maybe that is why.
I need to investigate more.
Thanks for checking

thuongtv-vn:
Hi all,
My application still got crashing and the callback "onRenderProcessGone" never called. Do you have any idea?

cristianoccazinsp:
> Hi all,
> My application still got crashing and the callback "onRenderProcessGone" never called. Do you have any idea?

Perhaps the webview implementation of the device you're getting crashes on does not fire the native event. 
The SDK notes state that at least Android Oreo is required for this, but not all vendors may implement it.

Can you provide more info on the crashes?

daniel1up:
What about the crashes for data:base64 URL? Is there any handle for that?

Issue #1484, opened 2020-06-25
iOS bouncesZoom would be nice


bitfabrikken:
**Is your feature request related to a problem? If so, Please describe.**

Should be able to give it bouncesZoom=false so it doesn't bounce the zoom.

**Describe the solutions you came up with**

bouncesZoom=false on WebView props should disable zoom bounce.

**Additional context**

I'll submit a pull request soon

PR #1485, opened 2020-06-25
Add bouncesZoom prop for iOS + macOS

bitfabrikken:
# Summary

Following issue #1484 I've added the bouncesZoom prop for iOS and macOS.

## Test Plan

Using https://github.com/react-native-community/react-native-webview/blob/master/example/examples/Scrolling.tsx , change user-scalable=no to user-scalable=yes and add minimum-scale=0.8, maximum-scale=2.0
Set property bouncesZoom on the Webview to false.
Pinch to zoom in/out and observe the zooming causes a "bounce" effect/animation. 
Set the property bouncesZoom to true and observe it no longer bounces zoom.
                
See above.

Run the example above.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -120,6 +120,7 @@ static NSDictionary* customCertificatesForHost;
     super.backgroundColor = [RCTUIColor clearColor];
     #endif // !TARGET_OS_OSX
     _bounces = YES;
+    _bouncesZoom = YES;
     _scrollEnabled = YES;
     _showsHorizontalScrollIndicator = YES;
     _showsVerticalScrollIndicator = YES;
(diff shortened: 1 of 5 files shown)


bitfabrikken:
Not tested on macOS.

apple/RNCWebView.m
```diff
@@ -174,7 +175,7 @@ - (instancetype)initWithFrame:(CGRect)frame
                                                selector:@selector(hideFullScreenVideoStatusBars)
                                                    name:UIWindowDidBecomeHiddenNotification
                                                  object:nil];
-
+
```
TheAlmightyBob, line 178: Could you clean up all the extra whitespace changes in this PR?

Otherwise this looks good to me. (although I'm not sure why you'd want to turn this feature off ;) )

PR #1487, opened 2020-06-26
Fix Android onLoadingFinish url comparison to compare using lowercases to solve infinite loading

Foomethod:
# Summary

This pull request resolves the problem where in Android, the Webview will end up in a state of infinite loading when the navigation state changes to a similar URL but with some letters capitalised.

Example:
current: _www.google.com_
next url: _www.Google.com_

This changes compares the current and next url in lower cases, so that the state of the webview can be set back to "IDLE" and not forever stuck in "LOADING".

1. Have a redirection where the next url is similar but with capital letters.

1. Pull the changes & rebuild android.

## Compatibility

Issue only appears in **Android** due to a different way of checking.

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/src/WebView.android.tsx
+++ b/src/WebView.android.tsx
@@ -230,7 +230,6 @@ class WebView extends React.Component<AndroidWebViewProps, State> {
 
   onLoadingFinish = (event: WebViewNavigationEvent) => {
     const { onLoad, onLoadEnd } = this.props;
-    const { nativeEvent: { url } } = event;
     if (onLoad) {
       onLoad(event);
     }
--- a/src/WebView.android.tsx
+++ b/src/WebView.android.tsx
@@ -230,15 +230,18 @@ class WebView extends React.Component<AndroidWebViewProps, State> {
 
   onLoadingFinish = (event: WebViewNavigationEvent) => {
     const { onLoad, onLoadEnd } = this.props;
+    const { nativeEvent: { url } } = event;
     if (onLoad) {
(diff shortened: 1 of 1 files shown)


src/WebView.android.tsx
```diff
@@ -237,9 +237,9 @@ class WebView extends React.Component<AndroidWebViewProps, State> {
     if (onLoadEnd) {
       onLoadEnd(event);
     }
-    if (url === this.startUrl) {
+    if (url && this.startUrl && url.toLowerCase() === this.startUrl.toLowerCase()) {
```
TheAlmightyBob, line 240: I think this could also be written as
```typescript
if (url?.toLowerCase() === this.startUrl?.toLowerCase()) {
```
?

Also I'm curious if comparing URLs via `toLowerCase` is safe for URLs with special characters... (https://stackoverflow.com/a/2140644)

Foomethod: Understand that many URLs contain special code, maybe using 
`url.localeCompare(this.startUrl)`
would be a better option

TheAlmightyBob: Perhaps, although React Native has a rocky history with supporting `locale` functions like that on Android (e.g. https://github.com/facebook/react-native/issues/12597). I'm not sure what the current state of things is, whether it's safe to use in more recent RN versions.

Issue #906, opened 2019-09-26
Android: app crash RNCWebViewManager$RNCWebView, size 1080x176205 exceeds max size 8192


phuongwd:
**Bug description:**

crash log:

> JNI DETECTED ERROR IN APPLICATION: JNI CallObjectMethod called with pending exception java.lang.IllegalStateException: Unable to create layer for RNCWebViewManager$RNCWebView, size 1080x176205 exceeds max size 8192

<details>
  <summary>Click to open full crash log</summary>
  
 ```
2019-09-26 17:09:35.469 5701-5701/com.demo.app A/zygote64: java_vm_ext.cc:504] JNI DETECTED ERROR IN APPLICATION: JNI CallObjectMethod called with pending exception java.lang.IllegalStateException: Unable to create layer for RNCWebViewManager$RNCWebView, size 1080x176205 exceeds max size 8192
2019-09-26 17:09:35.470 5701-5701/com.demo.app A/zygote64: java_vm_ext.cc:504]   at void android.os.MessageQueue.nativePollOnce(long, int) (MessageQueue.java:-2)
2019-09-26 17:09:35.470 5701-5701/com.demo.app A/zygote64: java_vm_ext.cc:504]   at android.os.Message android.os.MessageQueue.next() (MessageQueue.java:325)
2019-09-26 17:09:35.470 5701-5701/com.demo.app A/zygote64: java_vm_ext.cc:504]   at void android.os.Looper.loop() (Looper.java:142)
2019-09-26 17:09:35.470 5701-5701/com.demo.app A/zygote64: java_vm_ext.cc:504]   at void android.app.ActivityThread.main(java.lang.String[]) (ActivityThread.java:6938)
2019-09-26 17:09:35.470 5701-5701/com.demo.app A/zygote64: java_vm_ext.cc:504]   at java.lang.Object java.lang.reflect.Method.invoke(java.lang.Object, java.lang.Object[]) (Method.java:-2)
2019-09-26 17:09:35.470 5701-5701/com.demo.app A/zygote64: java_vm_ext.cc:504]   at void com.android.internal.os.Zygote$MethodAndArgsCaller.run() (Zygote.java:327)
2019-09-26 17:09:35.470 5701-5701/com.demo.app A/zygote64: java_vm_ext.cc:504]   at void com.android.internal.os.ZygoteInit.main(java.lang.String[]) (ZygoteInit.java:1374)
[...]

PR #1491, opened 2020-06-30
Disable force overlapping rendering for Android 9 or higher

SS-In:
# Summary

Following issue #906 I've added the fix to avoid IllegalStateException android.os.MessageQueue exception.

On some devices e.g. Samsung Galaxy J4+, J6+ WebView crashes when hardware acceleration is enabled. It may happen only when WebView is wrapped by ScrollView and exceeds max height supported by GPU. Then exception IllegalStateException android.os.MessageQueue is thrown. Unfortunately, disabling hardware acceleration does not resolve the issue as expected (e.g. displaying videos does not work anymore). To resolve this issue I added a fix, which disables overlapping rendering. In result, WebView does not crash with ScrollView.

## Test Plan
Tested on Samsung Galaxy J6+.

## What's required for testing (prerequisites)?
It can be reproduced only on some devices (from my logs) :

- Samsung Galaxy J4+
- Samsung Galaxy J6+
- Samsung Galaxy J4 Plus 2018
- Samsung Galaxy Tab A 8.0 2017
- Samsung Galaxy Tab A 8.0 2017
- Nokia 2.1

## What are the steps to reproduce (after prerequisites)?

To reproduce this issue we need WebView with large content wrapped by ScrollView. Without this fix, app crashes with exception.

`  <ScrollView
                contentContainerStyle={{ height: 4000 }} // it causes WebView crash 
            >
                <WebView
                    style={{ height: 4000, opacity: 0.99, flex: 1 }}
                    androidForceHasOverlappingRendering={false}
                    source={{ html: '<p>Large html content</p>' }}
                />
            </ScrollView>
`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |      ❌     |
| Android |    ✅    |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -164,7 +164,13 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   }
 
   protected RNCWebView createRNCWebViewInstance(ThemedReactContext reactContext) {
-    return new RNCWebView(reactContext);
+    RNCWebView webView = new RNCWebView(reactContext);
+
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+      webView.forceHasOverlappingRendering(false);
+    }
+
+    return webView;
   }
 
   @Override


TheAlmightyBob:
The "steps to reproduce" code includes setting `androidForceHasOverlappingRendering` on the `WebView`... but that's not an existing property of `WebView` and this PR does not add such a property??

Issue #1505, opened 2020-07-09
[iOS] WebView automatically opens url that don't match originWhitelist in external browser


phuwin95:
**Bug description:**
In iOS, WebView with custom `originWhitelist` automatically opens custom scripts in Safari without any user interaction.

**To Reproduce:**
```javascript
<WebView
  source={{
     uri: 'https://www.bbc.com',
  }}
  originWhitelist={['https://*bbc.com']}
/>
```
**Expected behavior:**
WebView doesn't open custom scripts in Safari without any user interaction.

**Screenshots/Videos:**
On load, the webview opens a link that doesn't fit the whitelist automatically
https://imgur.com/a/Tvbua2b

Error in expo console:
![image](https://user-images.githubusercontent.com/60187861/87035722-69d47880-c1f2-11ea-97c4-407838c90ab3.png)

**Environment:**
 - OS: iOS
 - OS version: 13.5
 - react-native version: React Native 0.62.2 for Expo SDK38
 - react-native-webview version: ^9.4.0

phuwin95:
Related to this
https://github.com/react-native-community/react-native-webview/issues/785

ifero:
Hi, I'm facing the same issue, I've whitelisted only my company domain, but every internal iframe is redirected to Safari instead of being loaded as part of the webpage.

@phuwin95 did you solved the issue?

phuwin95:
> Hi, I'm facing the same issue, I've whitelisted only my company domain, but every internal iframe is redirected to Safari instead of being loaded as part of the webpage.
> 
> @phuwin95 did you solved the issue?

Yes I did. So you can't rely on the webview component to whitelist your domain since it's not working properly. 
I need to manually filter urls using `onShouldStartLoadWithRequest`. With `isTopFrame` attribute, you can filter out requests that are from iframes. `passesWhitelist` function is taken from `react-native-webview` source https://github.com/react-native-webview/react-native-webview/blob/master/src/WebViewShared.tsx#L20
```js
  const onShouldStartLoadWithRequest = ({url, canGoBack, isTopFrame}: shouldStartLoadWithRequestEvent): boolean => {
    /**
     * Special case handling for ios
     * isFirstLoad is needed b/c onShouldStartLoadWithRequest is called on first load
     * isTopFrame is needed to separate iframe requests from topframe requests
     */
    if (Platform.OS === 'ios') {
      const isFirstLoad = url === originalUrl && !canGoBack;
[...]

Issue #785, opened 2019-08-15
Origin whitelist fails on iOS for iframes


Niksac:
Hi,

i use react-native-webview 6.9.0 with react 59.5 to embed a page that contains an iframe with a standard google maps embed like this:

```<iframe src="https://www.google.com/maps/embed?pb=!1m18!1m12!1m3!1d1666.5345895801888!2d-111.841501241774!3d33.34314243590285!2m3!1f0!2f0!3f0!3m2!1i1024!2i768!4f13.1!3m3!1m2!1s0x872ba9c40ac5d6f9%3A0x83f80f281a875961!2sReact+Defense!5e0!3m2!1sen!2sde!4v1565892032668!5m2!1sen!2sde" width="600" height="450" frameborder="0" style="border:0" allowfullscreen></iframe>```

I have set `originWhitelist` to my website root.
With android, as expected you can navigate within my project and clicking external URLs will be handled by the OS.

With iOS (12.4) Safari gets opened immediately, loading the iframe src URL (the google embed page).

Is this a known problem for iframes?

mizangl:
hey @Niksac did you find any solution for this, bc I'm experimenting the same issue.

Ibrahimkara92:
I solved the problem
In the node modules file we open the apple file inside the react-native-webwiew file and open RNCWebView.m 
We need to make some additions here
if (_onShouldStartLoadWithRequest) {
    NSMutableDictionary<NSString *, id> *event = [self baseEvent];
    [event addEntriesFromDictionary: @{
      @"url": (request.URL).absoluteString,
      @"mainDocumentURL": (request.mainDocumentURL).absoluteString,
      @"navigationType": navigationTypes[@(navigationType)]
    }];
    if (![self.delegate webView:self
      shouldStartLoadForRequest:event
                   withCallback:_onShouldStartLoadWithRequest]) {
      decisionHandler(WKNavigationActionPolicyCancel);
      return;
    }
  }

such lines
if (_onShouldStartLoadWithRequest) {
    NSMutableDictionary<NSString *, id> *event = [self baseEvent];
    [event addEntriesFromDictionary: @{
      @"url": (request.URL).absoluteString,
      @"mainDocumentURL": (request.mainDocumentURL).absoluteString,
      @"navigationType": navigationTypes[@(navigationType)]
    }];
      if (!navigationAction.targetFrame.isMainFrame){
                    decisionHandler(WKNavigationActionPolicyAllow);
                          return;
                }
[...]

Issue #724, opened 2019-07-15
[ios] specifying originWhitelist to ['http://.*', 'https://.*'] jumps to browser


jayshah123:
**Bug description:**

When specifying prop in WebView:
```
          originWhitelist={['http://.*', 'https://.*']}
```
The webview directly opens the site in the browser in IOS.
Works fine in android.

**To Reproduce:**
Specify origin whitelist as above in your webview prop.

**Expected behavior:**
WebView should stay and open the web page.

**Screenshots/Videos:**
WIll add soon.

**Environment:**
 - OS: Mac OS High sierra
 - OS version:
 - react-native version: 0.59.8
 - react-native-webview version: 5.12.1

imsantoshyadav:
@jayshah123 have you found out any solution for this issue

jayshah123:
Nope

asrytis:
`originWhitelist={['http://*', 'https://*']}` 👈 this worked for me on both iOS & Android (RN 0.59.9). Note the absence of `.`

Ibrahimkara92:
I solved the problem
In the node modules file we open the apple file inside the react-native-webwiew file and open RNCWebView.m
We need to make some additions here
if (_onShouldStartLoadWithRequest) {
NSMutableDictionary<NSString *, id> *event = [self baseEvent];
[event addEntriesFromDictionary: @{
@"url": (request.URL).absoluteString,
@"mainDocumentURL": (request.mainDocumentURL).absoluteString,
@"navigationType": navigationTypes[@(navigationType)]
}];
if (![self.delegate webView:self
shouldStartLoadForRequest:event
withCallback:_onShouldStartLoadWithRequest]) {
decisionHandler(WKNavigationActionPolicyCancel);
return;
}
}

such lines
if (_onShouldStartLoadWithRequest) {
NSMutableDictionary<NSString *, id> *event = [self baseEvent];
[event addEntriesFromDictionary: @{
@"url": (request.URL).absoluteString,
@"mainDocumentURL": (request.mainDocumentURL).absoluteString,
@"navigationType": navigationTypes[@(navigationType)]
}];
if (!navigationAction.targetFrame.isMainFrame){
decisionHandler(WKNavigationActionPolicyAllow);
return;
}
if (![self.delegate webView:self
shouldStartLoadForRequest:event
withCallback:_onShouldStartLoadWithRequest]) {
decisionHandler(WKNavigationActionPolicyCancel);
return;
}
}
we have to change it this way

PR #1513, opened 2020-07-13
Fix(ios): iframe links got rejected 

phuwin95:
# Summary

Related to #1505, #785, and #724 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -937,6 +937,10 @@ static NSDictionary* customCertificatesForHost;
       @"mainDocumentURL": (request.mainDocumentURL).absoluteString,
       @"navigationType": navigationTypes[@(navigationType)]
     }];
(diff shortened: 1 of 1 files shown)


phuwin95:
The original implementation was suggested by @Ibrahimkara92.

TheAlmightyBob:
@phuwin95 #1537 added `isTopFrame` to `shouldStartLoadForRequest`, so I _think_ this change is no longer necessary and the desired behavior can be achieved in your app code?

JakeStoeffler:
Wait, why couldn't this be merged? I can't imagine a use case where someone **wants** an iframe whose url isn't in the whitelist to automatically open in Safari. It's a terrible user experience and developer experience to have to implement a workaround for every app.

Is there any way we could push this across the finish line? I'd be happy to help.

PR #1515, opened 2020-07-14
fix path for iOS

7laria:
# Summary
Trying to use this with iOS, i found that it doesn't work because of wrong path 
I changed it to ./src

## Test Plan
react-native run-ios

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist


TheAlmightyBob:
@7laria How are you trying to install `react-native-webview`? There is no `src` path when installing from `npm`.

Issue #1434, opened 2020-06-02
Redirects not working properly on Android


timboektoe:
**Bug description:**
On Android redirects are not handled correctly within react-native-webview. Redirects are not taking place while they are working correctly in normal browser and react-native-webview on ios.

The problem appears when executing the following steps on android react-native-webview:

1. navigate to https://mijn.uwv.nl/
2. click the leftmost button labeled 'DigiD Inloggen'
3. on android webview you will return to the same page in other browsers you are forwarded to the login page. (see also the github link underneath to reproduce the issue)

If I study the redirection flow in my desktop browser I see also that 'SET-COOKIE' is part of the http responses. Could it be that something goes wrong while setting the cookie?

**To Reproduce:**
Checkout:
https://github.com/timboektoe/react-native-webview-redirect-bug

1. npm install
2. expo start
3. run on ios --> you will be redirected correctly to login page after clicking leftmost 'DigiD Inloggen' button.
4 run on android you will be redirected back to the same page after clicking leftmost 'DigiD Inloggen' button.

**Expected behavior:**
On Android will be redirected to the login page like in all other browsers

**Screenshots/Videos:**
https://parser-prod.aqopi.com/extension/redirect-bug.mp4

**Environment:**
 - OS: Android
 - OS version: all Android versions
 - react-native version: all React native versions
 - react-native-webview version: 8.0 (probably earlier also, but discovered on 8.0)

alexhidalgo:
Curious why there aren't more eyeballs on this. Would be a really useful feature. I can confirm it works fine on iOS and fails on Android.

Balasnest:
I am also facing the exact issue in the only android. It was working with webView API of react native.

timboektoe:
> I am also facing the exact issue in the only android. It was working with webView API of react native.

Yes rumour was that this issue was fixed a while ago, but the redirect issue still exists in situations with multiple redirects. Do you experience this problem in a situation with multiple redirects particularly?

Balasnest:
Yes. Multiple redirects are happening.

nmilcoff:
Did anyone manage to get this working? I'm having this issue with multiple redirects as well. iOS works fine without any tricks

timboektoe:
There's a copy of this issue here: https://github.com/react-native-webview/react-native-webview/issues/1654

I only managed to get it working with tricks...... We would like to pay somebody to help us fix this. How can I find the right person for that?

PR #1200, opened 2020-02-11
WIP: fix(android): add Referer header for url redirect load

rocwind:
# Summary
For redirected url loading, current behehavior is we always return `true` to prevent it being loaded in `shouldOverrideUrlLoading`, with `TopShouldStartLoadWithRequestEvent` being sent to RN and then call `webview.loadUrl()` if such redirect loading was approved on RN.
However, the `Referer` header would be different between `webview.loadUrl` and return `false` in `shouldOverrideUrlLoading`. It might cause redirect issues ( #1073 ) if `Referer` matters at the server side.  
The code change inside this PR is to save a `refererUrl` when sending `TopShouldStartLoadWithRequestEvent` to RN and use it to construct a request with `Referer` header later.

### What are the steps to reproduce (after prerequisites)?
test loading `https://s.click.taobao.com/fj2tptv` (url from #1073 ) 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅     |

## Checklist

- [x] I have tested this on a device and a simulator
- [ ] I added the documentation in `README.md`
- [ ] I mentioned this change in `CHANGELOG.md`
- [ ] I updated the typed files (TS and Flow)
- [ ] I added a sample use of the API in the example project (`example/App.js`)

Issue #1073, opened 2019-12-11
URL is always redirected in Android


bashen1:
**Bug description:**
In Android, https://s.click.taobao.com/fj2tptv will be redirected until it is blocked. However, the browser or other browsers with it jump normally. IOS normal jump

**To Reproduce:**

**Expected behavior:**

**Screenshots/Videos:**

**Environment:**
 - OS: Android
 - OS version: 9
 - react-native version: 0.61.5
 - react-native-webview version: 7.6.0

bashen1:
https://github.com/react-native-community/react-native-webview/pull/991

PR #1525, opened 2020-07-18
fix: add `Referer` header for Android `loadUrl()` command

rocwind:
- add "Referer" header in Android `loadUrl()` command to fix issues that caused by its absense
- expose `referer` field to `onShouldStartLoadWithRequest` event to let client include it into custom headers when necessary

# Summary
This is an update of previous PR #1200 

For redirected url loading, current behehavior is we always return true to prevent it being loaded in shouldOverrideUrlLoading, with TopShouldStartLoadWithRequestEvent being sent to RN and then call webview.loadUrl() if such redirect loading was approved on RN.
However, the Referer header would be different between webview.loadUrl and return false in shouldOverrideUrlLoading. It might cause redirect issues ( #1073, #1434 ) 

The proposed solution here is to:
1. (Android) keep a referer url when navigation happens and use it as referer header as default behavior.
2. (Both Android & iOS) include the `referer` along with other parameters in `onShouldStartLoadWithRequest` to let js land have a chance to deal with combining referer and other custom headers

Regarding the multiple consecutive redirects case described in #1434 (details: https://github.com/react-native-community/react-native-webview/pull/1200#issuecomment-650983886). It may still have issue on Android 6.0 and below with this fix. Since `boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request)` only works on Android N and above, and we cannot tell if current url loading request is server redirect or not without the `WebResourceRequest` - cannot decide whether the referer header need to be updated or not.

## Test Plan
1. test loading https://s.click.taobao.com/fj2tptv (url from #1073 )
2. test loading https://mijn.uwv.nl/ and click the leftmost button labeled 'DigiD Inloggen' ( #1434) on Android 7+

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -640,7 +640,18 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
           throw new RuntimeException("Arguments for loading an url are null!");
         }
         ((RNCWebView) root).progressChangedFilter.setWaitingForCommandLoadUrl(false);
-        root.loadUrl(args.getString(0));
+        String url = args.getString(0);
+        HashMap<String, String> headerMap = new HashMap<>();
+        // try to add referer header when navigate to a new url
+        String refererUrl = ((RNCWebView) root).progressChangedFilter.getRefererUrl();
+        if (refererUrl != null &&
+          // "no-referrer-when-downgrade" policy is the default behavior
+          // https://www.w3.org/TR/referrer-policy/#referrer-policy-no-referrer-when-downgrade
+          (!refererUrl.startsWith("https://") || url.startsWith("https://"))
+        ) {
+          headerMap.put("referer", refererUrl);
+        }
+        root.loadUrl(url, headerMap);
         break;
       case COMMAND_FOCUS:
         root.requestFocus();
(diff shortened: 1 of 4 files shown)


jamonholmgren:
This looks good. Can someone test it to make sure it works?

felipeaugusto1:
I just tested on an Android device and confirm that it works

Transfusion:
Tested in the Pixel 2 API 29 simulator, my use case is a website that only allows downloading of certain files if the referrer is on the same domain, and returns a 403 page otherwise. With the changes above the file download was triggered by the DownloadManager, which is an improvement over the behavior in master (the WebView redirects to the 403 page directly), however, because the DownloadManager didn't pass the Referer on, the resulting file's contents were the HTML plaintext of the 403 page.

I suggest adding 
```java
        String refererUrl = ((RNCWebView) webView).progressChangedFilter.getRefererUrl();
        if (refererUrl != null &&
          // "no-referrer-when-downgrade" policy is the default behavior
          // https://www.w3.org/TR/referrer-policy/#referrer-policy-no-referrer-when-downgrade
          (!refererUrl.startsWith("https://") || url.startsWith("https://"))
        ) {
          request.addRequestHeader("referer", refererUrl);
        }
```
right below https://github.com/react-native-community/react-native-webview/blob/261ba6c60d1a3d4101c48c198e59744ae7a29814/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L223 
to ensure that the Referer is passed over to the `DownloadManager` too.

rocwind:
Thanks for the testing & great suggestions.

I saw #1590 has been merged and released in v10.8.3: by handling the `shouldOverrideUrlLoading` in synchronous way and rely on the system (webview) default behavior, the redirect issue this PR was targeting for should be solved in #1590 too. So I am closing this PR as it seems no longer needed.

timboektoe:
We also tested this PR for mijn.uwv.nl and problem is solved now. Great work!

Issue #1532, opened 2020-07-22
Cookie not being sent with resource requests


1rahul1:
We have a use case in which when we loading a page in the webview, it itself does many resource requests and these resource requests require the cookie to be sent with them. Currently, react-native-webview can only intercept the URLs on new page load but not for the resource requests sent by the page itself.

**Proposed solution**
We can use a map to keep the record of the "authority name" of the URL and the cookie for it. We get this data from the header in the **source** prop. We can then override the **shouldInterceptRequest** callback in the RNCWebViewClient to intercept and send the cookie according to the "authority name". This will send the cookie with the resource requests which are sent by the page which we loaded. 
I can raise a PR for this.

TommyLeong:
Same here having issue in getting cookie pass into request header for subsequent pages load.

PR #1534, opened 2020-07-23
[Android] Sends cookies with resource requests

1rahul1:
To attach cookie with all resource requests for the page loaded
# Summary
* This PR solves this feature request [#1532 ](https://github.com/react-native-community/react-native-webview/issues/1532)

* How did you implement the solution?
I implemented this solution by adding a map for storing cookies for authority name and overriding the shouldInterceptRequest callback in RNCWebViewClient.

* What areas of the library does it impact?
RNCWebViewManager, RNCWebViewClient

## Compatibility

| OS      | Implemented |
| Android |✅|

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -110,6 +110,8 @@ import javax.annotation.Nullable;
  */
 @ReactModule(name = RNCWebViewManager.REACT_CLASS)
 public class RNCWebViewManager extends SimpleViewManager<WebView> {
+  
+  private static final Map<String, String> COOKIE_MAP = new HashMap<>();
 
   public static final int COMMAND_GO_BACK = 1;
   public static final int COMMAND_GO_FORWARD = 2;
@@ -797,6 +799,20 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
     public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request) {
       final String url = request.getUrl().toString();
       return this.shouldOverrideUrlLoading(view, url);
+    }
+
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+    @Override
+    public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceRequest request) {
+      String cookie = request.getRequestHeaders() == null ? null : request.getRequestHeaders().get("cookie");
+      if (null == cookie || cookie.length() == 0) {
+        if (COOKIE_MAP.containsKey(request.getUrl().getAuthority())) {
+          CookieManager.getInstance().setCookie(request.getUrl().toString(), COOKIE_MAP.get(request.getUrl().getAuthority()));
+        }
+      } else {
+        COOKIE_MAP.put(request.getUrl().getAuthority(), cookie);
+      }
+      return super.shouldInterceptRequest(view, request);
     }
 
     @Override


jamonholmgren:
Thanks for submitting this, @1rahul1 ! Does this only affect Lollipop?

TommyLeong:
@jamonholmgren  Unfortunately I wasn't having issue with Android, earlier my issue of passing cookie with resource request was with iOS. Somehow the current version of Android works fine for me for this part.

gkode:
@jamonholmgren @jamonholmgren @TommyLeong  I've faced this issue and customize it locally.  I have tested this code and it works fine. Can we merge it?

gkode:
@Titozzz

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -799,6 +801,20 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
       return this.shouldOverrideUrlLoading(view, url);
     }
 
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+    @Override
+    public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceRequest request) {
+      String cookie = request.getRequestHeaders() == null ? null : request.getRequestHeaders().get("cookie");
```
dwxw, line 807: Hi, just wondering if "cookie" should be case insensitive?  It didn't work when I declared my headers as `headers={{ Cookie: "key=value" }}`.  Appears to be case insensitive on iOS.

1rahul1: Hi @dwxw, Did you check this in Android? Did you check `headers={{ cookie: "key=value" }}`?

dwxw: Yes I tried your change and compared it to the existing iOS behaviour.  The headers prop works with `Cookie` or `cookie` in iOS, but with this change only `cookie` works on Android.  [AFAIK](https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive/5259004#5259004) the http header should be case-insensitive so you should handle both cases.

jamonholmgren: @1rahul1 Do you agree with @dwxw? If you want to make this case insensitive, I'll merge & release after. Or if not, I can merge and release now.

jamonholmgren:
We do intend to merge & release this. Just one outstanding question.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
1rahul1: @dwxw, thanks for checking that. I checked it on Android 9.0 but wasn't able to reproduce the issue which you pointed out. For me, it worked for both cases. Can you give more details on which version of Android used and how to reproduce it?

1rahul1: @jamonholmgren I checked this and it worked for me for both the cases. In the headers from [ `request.getRequestHeaders()`](https://github.com/1rahul1/react-native-webview/blob/9970504f2a6601531cc8557bd82f4342f9c1b3f2/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L831), the **keys** of the headers are in **lowercase** for the headers passed in the source prop, **irrespective** of the case of the key passed. [Here](https://github.com/1rahul1/react-native-webview/blob/9970504f2a6601531cc8557bd82f4342f9c1b3f2/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L831), I have used "cookie" which is in lowercase. So, I think the case-insensitive case is already handled.

dwxw: I've had another look and I still have this problem where only "cookie" works for me. The server does not receive the cookie unless I use "cookie", but after digging into this code it must be something else.  I've looked at `request.getRequestHeaders()` myself and it always returns a lowercase "cookie" no matter the case I use in the `headers` prop.  So not sure what my problem is but I think @1rahul1's code looks good.  Sorry for any confusion.

gkode:
@jamonholmgren Can you please merge it

gkode:
@jamonholmgren  Can you please specify what's your "outstanding question" on this PR?

Zaidos9:
Don't have access to resolve conflict. Created [PR 1700 [Android] Sends cookies with resource requests](https://github.com/react-native-webview/react-native-webview/pull/1700) to replace this PR.

PR #1535, opened 2020-07-24
Prevent video playing on background (Android)

ofito:
# Summary
We are having problems submitting an app to the Play Store this is due to one of our webviews with YouTube videos, when the app is in background those videos keep playing and interferes with [Play Store policies](https://play.google.com/about/privacy-security-deception/device-network-abuse/). 

This fix is based on the feature propose by the user @cereme in https://github.com/react-native-community/react-native-webview/issues/610. A perfect feature for this should be add a prop to activate the background playing when is needed.

Android device/simulator
Navigates to a YouTube video and send the app to a background state. Before the fix the audio still playing, after the fix the video is on pause.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -1122,12 +1122,12 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
     @Override
     public void onHostResume() {
-      // do nothing
+      super.onResume();
     }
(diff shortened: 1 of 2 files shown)


TheAlmightyBob:
Could you add a property to toggle this behavior? (in case there is some negative side-effect... I don't know the history but feel like empty overrides of `onHostPause`/`onHostResume` may have been put there for a reason...)

PR #1536, opened 2020-07-25
Convert ContentInsetAdjustmentBehavior from an enum back to a string union type

TheAlmightyBob:
# Summary

`ContentInsetAdjustmentBehavior` was originally added as a string union type (https://github.com/react-native-community/react-native-webview/pull/730). The change to support MacOS (https://github.com/react-native-community/react-native-webview/pull/1164) replaced it with an enum (https://github.com/react-native-community/react-native-webview/pull/1164/files#diff-bad447a30759e328812e9bae6798621cR269), which was a breaking change, especially as the enum was not exported from the main library. Exporting the enum would probably also help, but since this was the only enum and the rest of `react-native-webview` uses string unions instead, it seems best to revert it to a string union.

## Test Plan

The `ContentInsetAdjustmentBehavior` prop can now once again be simply set to a string value like "always." Before this change the property required using a value of the `ContentInsetAdjustmentBehavior` enum, which was difficult to set since that enum was not exported by `react-native-webview`.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅  |
| Android |    ✅     |

## Checklist

- [ ] I mentioned this change in `CHANGELOG.md` (uh, **this file doesn't exist?**)

### DIFF
--- a/src/WebViewTypes.ts
+++ b/src/WebViewTypes.ts
@@ -285,12 +285,7 @@ export interface AndroidNativeWebViewProps extends CommonNativeWebViewProps {
   readonly urlPrefixesForDefaultIntent?: string[];
 }
 
-export enum ContentInsetAdjustmentBehavior {
-  automatic = 'automatic',
-  scrollableAxes = 'scrollableAxes',
-  never = 'never',
-  always = 'always'
-};
+export declare type ContentInsetAdjustmentBehavior = 'automatic' | 'scrollableAxes' | 'never' | 'always';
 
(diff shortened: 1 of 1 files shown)


TheAlmightyBob:
@jamonholmgren I'm not sure what this project's bar is for breaking changes. Technically I would consider it an un-breaking of a previous breaking change, and it's a pretty small break. But if anybody _was_ using the enum, then they will need to update code (even if just a tiny bit), so I could see it being labeled a breaking change.

PR #1537, opened 2020-07-25
feat(events): Add isTopFrame to shouldStartLoadForRequest

TheAlmightyBob:
# Summary

`onLoadingStart` is not raised for inner frames, but `onShouldStartLoadWithRequest` still is on iOS. This keeps that behavior but adds `isTopFrame` to `onShouldStartLoadWithRequest` so that apps can perform their own filtering if desired. Android does not have that behavior but `isTopFrame` is still present and hardcoded to `true` for API consistency.

## Test Plan

Ran `yarn ci`. Tested with an app loading pages with inner frames.

I'm a bit fuzzy on the type naming patterns (e.g. `WebViewNavigation` vs `WebViewNavigationEvent`) so open to feedback there.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [X] I updated the typed files (TS and Flow) (**"Flow"?? I don't see any Flow types in this project**)

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/events/TopShouldStartLoadWithRequestEvent.kt
+++ b/android/src/main/java/com/reactnativecommunity/webview/events/TopShouldStartLoadWithRequestEvent.kt
@@ -14,6 +14,8 @@ class TopShouldStartLoadWithRequestEvent(viewId: Int, private val mData: Writabl
 
   init {
     mData.putString("navigationType", "other")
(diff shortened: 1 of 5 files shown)


phuwin95:
Would also be good to export `shouldStartLoadWithRequestEvent` in `index.d.ts` file so it can be used without extending `WebViewNavigation`

PR #1546, opened 2020-07-30
Add try-catch for DownloadManager.Request

jasongaare:
Prevents a crash when attempting to download a non HTTP/HTTPS link (like a url beginning with `blob://`)

# Summary

**Changed:** To prevent a crash when attempting to download a file that is from a URI other than HTTP/HTTPS, wrap the download functionality with a try-catch in `RNCWebViewManager.java` so it will fail slightly, rather than crashing the entire application.

_Notes:_
- This is a duplicate of PR https://github.com/react-native-community/react-native-webview/pull/1217, which was unfortunately closed due to inactivity. 
- This PR also is a solution for https://github.com/react-native-community/react-native-webview/issues/1090, "solution" meaning this will prevent the app from crashing. 
- This PR does NOT resolve any issues related to downloading a url that begins with `blob://`, **it merely prevent the crash**

(Excellent example lifted from previous PR)
```jsx

const App = () => {
  return (
    <View style={{ flex: 1 }}>
      <WebView
        source={{
        }}
        javaScriptEnabled
        allowFileAccess
        mixedContentMode={'always'}
        domStorageEnabled
        textZoom={150}
      />
    </View>
  )
}

export default App;
```

Using the above `uri` source, attempting to click the link "Download JSON" currently causes a crash on Android because of the error described in #1090:

```
java.lang.IllegalArgumentException Can only download HTTP/HTTPS URIs
```

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist


cristianoccazinsp:
Any thoughts on also allowing `onShouldStartLoadWithRequest` to fire / be used? This way, we could possibly warn/prevent the crash before it even happens. 

As far as I know, iOS will fire `onShouldStartLoadWithRequest` before trying to download a blob, whereas Android will just crash/ignore the prop completely.

jasongaare:
> Any thoughts on also allowing `onShouldStartLoadWithRequest` to fire / be used? This way, we could possibly warn/prevent the crash before it even happens.

💯 I can see if I can hunt that down too -- do you have any insights?

cristianoccazinsp:
> > Any thoughts on also allowing `onShouldStartLoadWithRequest` to fire / be used? This way, we could possibly warn/prevent the crash before it even happens.
> 
> 💯 I can see if I can hunt that down too -- do you have any insights?

I haven't really looked into it, but if I have to guess, I would say that the whole download manager / download process ignores `onShouldStartLoadWithRequest`.

jasongaare:
Perhaps a few leads here: https://github.com/react-native-community/react-native-webview/issues/381

Specifically I think the `shouldInterceptRequest` method could be utilized

jasongaare:
@Titozzz @jamonholmgren, kindly requesting if this could be reviewed. A simple fix to prevent an unpleasant crash.

TheAlmightyBob:
Re: @jamonholmgren's question, it looks to me like the `DownloadManager.Request` constructor is what raises the exception.

Also, perhaps it would be worth raising an error so it could be handled in React Native code, rather than just using `println`?

benediktviebahn:
What is the state of this PR? Would be nice if this could be merged and released soon. Is there a quick workaround I can use in the meantime to prevent this crash?

jasongaare:
I'm sorry I haven't had time to go in and work on this, locally I created a `.patch` file that runs in my post install script to apply the above changes until a fix is merged. Again, haven't had time to isolate exactly where the try-catch could be whittled down, but the patch is working for us for now

cristianoccazinsp:
Can this still be merged?

cristianoccazinsp:
@jasongaare it would be great to update the branch and eventually merge this!

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -202,36 +203,40 @@ public void onDownloadStart(String url, String userAgent, String contentDisposit
...
         RNCWebViewModule module = getModule(reactContext);
 
-        DownloadManager.Request request = new DownloadManager.Request(Uri.parse(url));
-
-        String fileName = URLUtil.guessFileName(url, contentDisposition, mimetype);
-        String downloadMessage = "Downloading " + fileName;
-
-        //Attempt to add cookie, if it exists
-        URL urlObj = null;
         try {
```
erAdnan, line 215: please merge this PR or else get some fixes for blob crash..

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -202,36 +203,40 @@ public void onDownloadStart(String url, String userAgent, String contentDisposit
...
-          urlObj = new URL(url);
-          String baseUrl = urlObj.getProtocol() + "://" + urlObj.getHost();
-          String cookie = CookieManager.getInstance().getCookie(baseUrl);
-          request.addRequestHeader("Cookie", cookie);
-        } catch (MalformedURLException e) {
-          System.out.println("Error getting cookie for DownloadManager: " + e.toString());
-          e.printStackTrace();
-        }
+          DownloadManager.Request request = new DownloadManager.Request(Uri.parse(url));
```
erAdnan, line 216: here if we will add new DownloadManager.Request(Uri.parse(url.replace("blob:", "")), then blob crash can be fixed or else if any other solution is there that will be fine

erAdnan (changes requested):

cristianoccazinsp:
Bump. This still needs attention

Nikooos:
I'd love to bump this one too as it's causing crashes for me too.

avinashsahoo:
**Android:**

I've just fixed the issue, by putting a if condition before DownloadManager.Request in RNCWebViewManager.java. If the Url contains "blob", then it'll simply ignore rest of the logic. At least it will prevent crashing the app on android.

    if(!url.contains("blob"))

**iOS**

It's anyway handling by onShouldStartLoadWithRequest.

    onShouldStartLoadWithRequest={(req) => { 
      if (req.url.startsWith("blob")) { 
        alert("Files can't be downloaded.");
        return false; 
      } 
        return true; 
      }
    }

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
avinashsahoo: By doing this, it will still download the junk file. It'll not prevent downloading the file, but yes it'll prevent crashing

Issue #1445, opened 2020-06-09
Semver violation for peer dependency changes


elashpear:
**Bug description:**
[Peer dependencies are generally considered part of a package's public API](https://github.com/semver/semver/issues/502). When a change is made (such as https://github.com/react-native-community/react-native-webview/commit/228f10d9176098dd75a5e44739c55b3200727ec4#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R32 ) that removes previously valid peerdep versions from the accepted range, it should be handled as a breaking change and trigger a major version bump. 

This will make it easier to track down versions that introduced a peerdep change as well as flag consumers that they will have to update their code to be compatible with the new library version. Thanks.

**To Reproduce:**
N/A

**Expected behavior:**
When removing valid ranges from `peerDependencies`, please bump the major version.

Also, as a side note, your latest version specifies peerdeps on
```
    "react": "16.11.0",
    "react-native": ">=0.60 <0.63"
```
However, react-native 0.60.x specifies an exact peerdep of `react@16.8.6`, and 0.61.x requires `react@16.9.0`, so this constraint is not satisfiable with any version of react-native less recent than 0.62.0.
**Screenshots/Videos:**
N/A

**Environment:**
N/A

chrisbobbe:
Similarly, https://github.com/react-native-community/react-native-webview/commit/bf1d64571#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R29 (released in v8.0.5, a patch version) effectively brought the minimum supported React Native version to v0.61 (see https://github.com/facebook/react-native/commit/0ccedf396#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R84).

chrisbobbe:
@github-actions bot, please don't close this yet—there's no sign a maintainer is aware of the issue.

elashpear:
Boo :(

safaiyeh:
Made a PR here

PR #1583, opened 2020-08-19
fix(deps): Update package.json

safaiyeh:
# Summary

WebView is locking the version from the latest React Native version. https://github.com/react-native-community/react-native-webview/issues/1581 & https://github.com/react-native-community/react-native-webview/issues/1445, we should not have an upper version lock so developers can report issues with the latest RN version

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [] I added the documentation in `README.md`

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   },
   "peerDependencies": {
     "react": "16.11.0",
-    "react-native": ">=0.60 <0.63"
+    "react-native": ">=0.60"
   },
   "dependencies": {
     "escape-string-regexp": "2.0.0",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "type": "Component"
   },
   "peerDependencies": {
-    "react-native": ">=0.60"
+    "react-native": ">=0.60 <0.64"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


chrisbobbe:
Coming from #1445, one thing that would be a big help is a change of versioning strategy for the future. As @elashpear said:

> When removing valid ranges from `peerDependencies`, please bump the major version.

Note that this is also relevant for changes to `react` under `peerDependencies`.

Also, we'd like it to be known that React Native versions before 0.62.0 are not supported, despite the fact that ">=0.60" is part of the range for `react-native`. This is because of the `react` peer dependency having "16.11.0" as its entire range (this happened randomly in a minor release, 10.1.0, without being announced as a breaking change). Details are at #1445.

safaiyeh:
Hmm removing the react peer dependency would be better. I believe the real goal is to make sure the module only supports >= react-native 0.60

chrisbobbe:
> Hmm removing the react peer dependency would be better.

I think that sounds good, as long as the module still works with `react` versions outside the range of "16.11.0".

PR #1588, opened 2020-08-23
feat(android): Add androidLayerType as prop

oliviacaraiman:
# Summary
This PR is to add a new React prop, `androidLayerType`, that allows [setting the layer type](https://developer.android.com/reference/android/view/View#setLayerType(int,%20android.graphics.Paint)) for Android using props. It also deprecates `androidHardwareAccelerationDisabled` prop, as the use of both of them at the same time might cause undefined behaviour.

**Motivation:**
The general motivation is to be able to control the layer type through props.
It would also solve the problem of rendering big pieces of HTML in a WebView when the hardware acceleration is disabled.
When we use HA, all drawing operations that are performed on a View's canvas use the GPU, so the layer is set to LAYER_TYPE_HARDWARE. Disabling HA means setting the view to LAYER_TYPE_SOFTWARE, which uses a drawing cache with a specific amount of available memory. If we want to render big pieces of HTML, they might not fit in memory and so they won't be displayed. The problem can be solved by using LAYER_TYPE_NONE, meaning the view is not backed-up by any buffers.

## Test Plan
Tested in my app by linking my fork to it. Worked as expected.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -299,6 +299,21 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
     }
   }
 
+  @ReactProp(name = "androidLayerType")
+  public void setLayerType(WebView view, String layerTypeString) {
+    int layerType = View.LAYER_TYPE_NONE;
+    switch (layerTypeString) {
+        case "hardware":
+          layerType = View.LAYER_TYPE_HARDWARE;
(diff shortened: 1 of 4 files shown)


oliviacaraiman:
Hey @Titozzz, any chance I could have a review for this any time soon? :) 
I think the tests should be rerun, but I can't do that. The failed tests don't seem related to my changes

Titozzz (approved):

PR #1590, opened 2020-08-24
feat(android): Synchronous onShouldStartLoadWithRequest

dvicory:
# Summary

*This is a re-imagining of #1479. #1479 had some ABI issues that could only be solved by backporting react-native patches and using NDK r21, which we believe actually exploited a compiler bug. This PR does not have any such issues as it does not use JSI/JNI to accomplish its goals, but only pure Java. This will be much easier to maintain for the future.*

We (myself and @trcoffman) discovered a very interesting bug in our WebView app. Inside our app, there was a link that would HTTP redirect to a different origin, and then immediately manipulate history (to remove an authentication token from the URL and to navigate). This would cause the user to be stuck going back (you would go back and the history/location manipulation would happen again). The only way to get back was to tap the back button quickly multiple times in succession. Essentially, WebView added extra entries to the history stack that should not be there.

We tracked it down to the style of overriding loading. Essentially, every navigation on Android is suppressed in `shouldOverrideUrlLoading`, and an event is sent to JS to decide to navigate or not. If the navigation should happen, it causes a new page load. With this technique, the WebView seems to lose "context" about redirects and can improperly add to the history stack under the right conditions. Interestingly, this asynchronous approach also causes a lot of issues with debugging the WebView with the network inspector. Initiators are incorrect and the redirects aren't clearly exposed, since every load is a new page load.

The fix is to be able to synchronously decide in `shouldOverrideUrlLoading` whether to load or not. iOS does this approach by locking, sending an event, and waiting for the decision to come in asynchronously, and then unlocking. We were able to do a similar approach for Android with very certain APIs. Similar to `onMessage`, we send a direct message to JS, lock the main UI thread, and synchronously call back into the Java from JS with the decision. Any other bridge APIs we tried to use caused a deadlock. We've made this so that we're very defensive and robust about which thread `shouldOverrideUrlLoading` is on, as it doesn't seem there is any guarantee made in the documentation.

The [Android docs warn against doing something very similar to what was being done](https://developer.android.com/reference/android/webkit/WebViewClient#shouldOverrideUrlLoading(android.webkit.WebView,%20android.webkit.WebResourceRequest)).

> Note: Do not call WebView#loadUrl(String) with the request's URL and then return true. This unnecessarily cancels the current load and starts a new load with the same URL. The correct way to continue loading a given URL is to simply return false, without calling WebView#loadUrl(String).

We found that with debugging, it's not possible to use synchronous methods. So first we check that we can get a JS runtime before trying the synchronous approach (which seems to be the best proxy that this technique is possible). As a fallback, the old asynchronous deny-then-load approach is kept when there is no JS runtime. This could allow for buggy behavior like we saw, but will allow users to debug their `onShouldStartLoadWithRequest`.

Exploiting this behavior is non-trivial. It requires a link that redirects to a different origin, that then immediately manipulates browser history (`window.location` plus `history`). I've attached a video of behavior before our fix. Notice that tapping back doesn't go back to the original screen with the menu of links. It requires multiple taps to get back.

ezgif-3-c932b0bd6e92

This is after our fix:

ezgif-3-27d2f00c7109

We've tested this fix quite a bit already in our app. I don't believe adding an example case makes sense.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

We don't depend on any internals of react-native other than what #1203 already used.

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -4,7 +4,6 @@ import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
 import android.app.DownloadManager;
 import android.content.Context;
-import android.content.Intent;
 import android.content.pm.ActivityInfo;
 import android.content.pm.PackageManager;
 import android.graphics.Bitmap;
@@ -14,8 +13,7 @@ import android.net.http.SslError;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Environment;
-import androidx.annotation.RequiresApi;
-import androidx.core.content.ContextCompat;
+import android.os.SystemClock;
 import android.text.TextUtils;
 import android.util.Log;
 import android.view.Gravity;
@@ -41,6 +39,12 @@ import android.webkit.WebView;
 import android.webkit.WebViewClient;
 import android.widget.FrameLayout;
 
+import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
+import androidx.core.content.ContextCompat;
+import androidx.core.util.Pair;
+
+import com.facebook.common.logging.FLog;
 import com.facebook.react.views.scroll.ScrollEvent;
 import com.facebook.react.views.scroll.ScrollEventType;
 import com.facebook.react.views.scroll.OnScrollDispatchHelper;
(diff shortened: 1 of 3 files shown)


Titozzz:
This is so cool. Great work guys.

Titozzz (approved):

PR #1616, opened 2020-09-06
WebViewAssetLoader support

duxiaofeng-github:
# Summary

i need to access my local asset, a html file, over https. because some of my js code require [Secure_Contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts), so i can't access that html by **file:///** protocol.

android x provide a new lib **WebViewAssetLoader** to achieve this goal. [more detail](https://developer.android.com/reference/androidx/webkit/WebViewAssetLoader).

there is a [example](https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/webkit/integration-tests/testapp/src/main/java/com/example/androidx/webkit/AssetLoaderAjaxActivity.java) code on android source code to show how we can use WebViewAssetLoader.

---

according to androidx.webkit's [doc](https://developer.android.com/reference/androidx/webkit/package-summary), the **minimum sdk version** to use this library is 14. react-native-webview is 16, so the requirement is fit.

finally we need to add an dependency for build.gradle

```
dependencies {
    implementation 'androidx.webkit:webkit:1.2.0'
}
```

## Test Plan

according to this [doc](https://developer.android.com/reference/androidx/webkit/WebViewAssetLoader), after apply WebViewAssetLoader, we can access assets by **https://appassets.androidplatform.net/assets/some-assets** now.

## Compatibility

there is nothing need to do with other platforms

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌    |
| Android |    ✅    |

## Checklist

### DIFF
--- a/docs/Guide.md
+++ b/docs/Guide.md
@@ -97,15 +97,8 @@ For example:
 
 ```tsx
 <WebView
-  ref={webViewRef}
-  originWhitelist={['*']}
   source={{
     uri: 'https://appassets.androidplatform.net/assets/index.html',
-  }}
-  onLoadEnd={sendRequests}
-  onMessage={(event) => {
-    console.log('resolve', event);
-    resolve(event.nativeEvent.data);
   }}
 />
 ```
--- a/docs/Guide.md
+++ b/docs/Guide.md
@@ -89,6 +89,27 @@ class MyWeb extends Component {
 }
 ```
 
+**Access local assets by WebViewAssetLoader on android x**
+
+Android x provide a new way to access assets, by using WebViewAssetLoader. now we can access assets via uri **https://appassets.androidplatform.net/assets/some-assets**
+
+For example:
+
+```tsx
+<WebView
+  ref={webViewRef}
+  originWhitelist={['*']}
+  source={{
+    uri: 'https://appassets.androidplatform.net/assets/index.html',
+  }}
+  onLoadEnd={sendRequests}
+  onMessage={(event) => {
+    console.log('resolve', event);
+    resolve(event.nativeEvent.data);
+  }}
+/>
+```
+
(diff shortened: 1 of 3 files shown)


docs/Guide.md
```diff
@@ -89,6 +89,20 @@ class MyWeb extends Component {
 }
 ```
 
+**Access local assets by WebViewAssetLoader on android x**
+
+Android x provide a new way to access assets, by using WebViewAssetLoader. now we can access assets via uri **https://appassets.androidplatform.net/assets/some-assets**
```
TheAlmightyBob, line 94: nit: I think it should be "Android X"?

TheAlmightyBob (changes requested):
This looks good to have support for, but could you make it configurable? In apps where the devs don't have full control of the content being loaded in the WebView, I could imagine it being undesirable that a page could link to `https://appassets.androidplatform.net` to make the WebView unexpectedly display random app assets?
(it would probably be _ideal_ to expose full configuration of domain URL, supported paths, etc, but for the purpose of this PR I think just having the feature disabled by default and a property to enable would be good)

Does iOS not have this problem with Secure Contexts and "file://", or are you just not using iOS?

docs/Guide.md
duxiaofeng-github: 😂

duxiaofeng-github:
ok, i'll make it configurable.

> Does iOS not have this problem with Secure Contexts and "file://", or are you just not using iOS

Secure context is supported by "file://" protocol of iOS by default. so we don't need to do any magic for it.

mlecoq:
Hi, I would like to user WebAssetLoader, could this be merged ?

PR #1636, opened 2020-09-14
fix(macOS): Don't include iOS pull-to-refresh control

bralax:
# Summary
In the past couple weeks, a new feature was added to allow for pull to refresh support. This was implemented using a UIRefreshControl which breaks the support for react-native-macos as macOS does not have access to that library. I have moved functions relating to pull to refresh to be inside an if the target is not macOS so that it can compile on macOS.

## Test Plan

Nothing has changed to the library except it now can compile again in a react-native-macos project. I created a project, added the dependency, fixed these things and it worked again. No code was added, just platform conditionals were added.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| macOS     |    ✅     |

## Checklist

### DIFF
--- a/apple/RNCWebView.h
+++ b/apple/RNCWebView.h
@@ -63,7 +63,9 @@
 @property (nonatomic, assign) BOOL ignoreSilentHardwareSwitch;
 @property (nonatomic, copy) NSString * _Nullable allowingReadAccessToURL;
 @property (nonatomic, assign) BOOL pullToRefreshEnabled;
+#if !TARGET_OS_OSX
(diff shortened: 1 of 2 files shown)


TheAlmightyBob (approved):

alloy (approved):
I resolved the merge conflict and now waiting for CI to go green.

alloy:
This is actually green, but it looks like due to this repo moving to a different GitHub org CircleCI isn't reporting back the status. See https://app.circleci.com/pipelines/github/react-native-webview/react-native-webview?branch=pull%2F1636

PR #1679, opened 2020-10-07
SSL pinning using TrustKit SDK added

RajeshCentrica:
SSL pinning using TrustKit SDK added


RajeshCentrica:
Not required this PR

Issue #1650, opened 2020-09-22
Add property to prevent keyboard from showing [$100]


mikob:
I need a boolean prop (perhaps called autoShowKeyboard) that controls whether the keyboard appears when a user focuses an input within the webview. Sometimes I need to disable the keyboard from showing automatically, other times I need to let it show. I don't have control over the content in the webview - it can be any website.

Note: using injectJS to, for example, put a "readonly" prop on the inputs won't work because the DOM content can change dynamically at any time - eg. if the website in the webview uses React, Vue or another SPA framework.

mikob:
I have put $100 on bounty source for this issue. https://www.bountysource.com/issues/93085569-add-property-to-prevent-keyboard-from-showing Please let me know if this is not allowed and I will remove this comment!

sanathusk:
Should the behavior be input specific or consistent across all the input within a webview? Is it ok to have a behavior where we can either disable all the inputs or enable all of them but not selectively?

rares-pop-sage:
I have the same problem. Any solution anyone?

PR #1696, opened 2020-10-13
[Add] Prop autoShowKeyboard to prevent keyboard from showing

remarket-vn:
Issue related #1650


jamonholmgren (commented):
Looks great. Can you update the documentation too?

jamonholmgren:
Additionally, can you _only_ add the listener if the property is enabled, and remove the listener if it's not needed? The listener should really only run if `autoShowKeyboard` is set to `false`.

aland96:
I'm interested in this feature, it would be handy in cases where a custom keyboard is already used within the webview.

R-iskey:
Why it's closed, what's the problem with that?

PR #1704, opened 2020-10-19
update injectedJavascript ios docs

Crisfole:
This is a gotcha that's gotten me more than once. `onMessage` is a required parameter _if_ you're injecting javascript.


jamonholmgren:
Good call! Thanks @Crisfole !

Issue #1705, opened 2020-10-19
Add support for preventDefault when using onError method


nochkin:
I would like to propose a small feature request.

I would like to handle errors in my code. I tried to use the 'onError' method and tried to do syntheticEvent. preventDefault() to do that, but it still continued and I did not get the behaviour I wanted.

I propose to add `isDefaultPrevented()` in the onError method to support it.
The change does not affect the existing code unless someone does `preventDefault()`.

nochkin:
Here is my proposed PR to make it happen:
https://github.com/react-native-webview/react-native-webview/pull/1706

nochkin:
I provided a PR, but it's waiting for someone to approve it.

nochkin:
The PR to resolve it was submitted already.
#1706

nochkin:
Still waiting on merging the PR.

nochkin:
The PR is opened and approved by one of the maintainers. Waiting to be merged.

Titozzz:
I think it was solved by https://github.com/react-native-webview/react-native-webview/commit/279d6ae98c449b3c2d9ab2cc567f99bcc6e298b6.

But, to be fair, you fought admirably well against that Github bot, sorry 🤗

PR #1706, opened 2020-10-19
Add support for preventDefault when using onError method

nochkin:
This is my proposed PR to support `preventDefault()` when using `onError` method as described in this [feature request](https://github.com/react-native-webview/react-native-webview/issues/1705).


docs/Reference.md
```diff
@@ -265,6 +265,8 @@ Example:
 />
 ```
 
+The `syntheticEvent` can be stopped doing its default action by calling `syntheticEvent.preventDefault()`.
```
jamonholmgren, line 268: I'd probably put this at the end of this section (just before the ### `onLoad` line).

jamonholmgren (changes requested):
Looks straightforward enough! Just one minor tweak to the docs and should be ready to go.

docs/Reference.md
nochkin: Updated it according to your suggestion.

nochkin:
I updated the documentation only and I see some tests are failing (one of them is failing due to timeout downloading packages).
Is there a way to re-run all tests without making any changes to PR?

nochkin:
@jamonholmgren is there something else I need to do to make it merged? I updated the doc as suggested.

matgardon:
+1 would be interested in this also, current behavior doesn't match my needs.

nochkin:
@matgardon,
Yes, the PR was kinda approved at first and then I was asked to make a minor doc update which I did, and then it hangs somewhere in the space. I hope someone on a way to Mars will see it and pick it up. What are the chances?

nochkin:
@jamonholmgren please merge.

Titozzz (approved):

Titozzz:
I was on my way to mars...

This has been fixed by https://github.com/react-native-webview/react-native-webview/commit/279d6ae98c449b3c2d9ab2cc567f99bcc6e298b6.

Sorry for not coming back to you earlier. Thanks for fighting hard !

nochkin:
@Titozzz 
Thanks for the follow up, but I see the fix you mentioned just removes some warnings in the console.
My proposed change actually adds a new behavior where you could handle errors in your own code instead of letting the RN-Webview to shows its own error for a user.

Titozzz:
You are right. I thought the logs annoyed you, not the state. What's your usecase ? I'm worried this change might be an issue since it affects the internal state of the WebView

If we are being honest I'd rather remove all this state management as it doesn't make sense to deal with it inside the WebView but that's quite a breaking change

nochkin:
The default behavior stays the same. If you want to handle your own errors, then you do `preventDefault`.
Unless someone did `preventDefault` for no reason (it won't do anything in the current version), it would not affect the way it's being used now.

You do 'preventDefault` in the `onError` callback only if you want this new behavior. Otherwise, it's default.

A particular use case is when you want to handle errors in your own app instead of letting RN-Webview spit out a generic error to a user. For example, you want to have a fallback procedure or come up with a message suggesting on how this can be resolved on the user's side.

Again, the existing apps would not see it unless there is `preventDefault` which triggers the new behavior.

Hopefully, that will make sense and not confuse you. I know Mars is pretty far :-) Let me know.

Titozzz:
No no I understood you're changes but they are a small patch to a bigger issue. Why does WebView even handles it in a first place 😂. I'll merge when I get home

PR #1709, opened 2020-10-20
ssl error fixed on android

HamzaNadeemIqbal:


jamonholmgren:
@HamzaNadeemIqbal Can you explain what you're doing here? Looks like you're just removing the error handling code?

HamzaNadeemIqbal:
This code was not letting my webview display url without ssl validation as i did not have access to the website itself, so i landed upon a suggestion to remove the handler which worked for me @jamonholmgren

ahelord:
Any update?

PR #1722, opened 2020-10-27
Add Highlight color change and Edit Menu overwrite

brian-perlego:
Hello,

For a project I needed to add highlight actions to the WebView, with those modifications it's possible to change the colour of the selection on iOS and on both iOS and Android the edition menu when you long press on text is hidden.

This is working fine on my project, but I would like to add those to the origin repo, it will be easier to maintain and maybe it will be useful for someone.

If I could get some pointer on what needs to be added for the PR to be approve that would be great,

Thank you!


TheAlmightyBob:
The 32 commits is a bit suspicious... I'm guessing you need to rebase?

src/WebView.android.tsx
```diff
@@ -164,6 +164,18 @@ class WebView extends React.Component<AndroidWebViewProps, State> {
     );
   };
 
+  /**
+   * Change the color of the selection tool, use only by Perlego for the highlight
```
TheAlmightyBob, line 168: "use only by Perlego"??

index.d.ts
```diff
@@ -32,6 +32,11 @@ declare class WebView<P = {}> extends Component<WebViewProps & P> {
      */
     extraNativeComponentConfig: () => any;
 
+    /**
+     * Change the color of the selection tool
+     */
+    setSelectionColor: (hexColor: string) => void;
```
TheAlmightyBob, line 38: I'm not seeing corresponding configuration for whether or not to hide the menu?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -661,6 +667,8 @@ public void receiveCommand(WebView root, int commandId, @Nullable ReadableArray
           throw new RuntimeException(e);
         }
         break;
+      case COMMAND_SET_SELECTION_COLOR:
+        break;
```
TheAlmightyBob, line 671: This seems incomplete?

TheAlmightyBob (changes requested):

PR #1728, opened 2020-10-29
fix(iOS): restore cookie handling for iOS <= 10

jackmu95:
This PR fixes the shared cookie handling for iOS <= 10 when the request is `nil`.
This would otherwise end in a `EXC_BAD_ACCESS` crash.


Titozzz (approved):

jackmu95:
@Titozzz Any update on this?

Titozzz (approved):

PR #1735, opened 2020-11-04
[Android] Asks for camera permission when trying to access camera

CassioMG:
## Issue Solved
Currently, the Android camera permission has to be requested **before** loading the `<WebView>`, which usually frighten users since they didn't have to tap any camera button to see the native request permission dialog. This PR solves this UX issue by asking for camera permission when user taps the camera button, like it already does on iOS.

## Testing
@fnando has created this testing page that just renders a "Choose file" button which should display the camera button. This is the expected behaviour for the camera button:
- Once the camera button is tapped, the camera permission request dialog should pop-up.
- Once permission is accepted, it should open the camera and normally capture the image.
- If permission is denied, it should display the following toast: `Cannot open camera as permission was denied. Please provide permission to access camera, in order to capture a photo.`
- If camera button is tapped after denying permission, it should display the permission request dialog once again.
```js
<WebView source={{ uri: "https://codepen.io/fnando/full/XWKNjWE" }} />
```

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
@@ -1,6 +1,7 @@
 package com.reactnativecommunity.webview;
 
 import android.Manifest;
+import android.annotation.TargetApi;
 import android.app.Activity;
 import android.app.DownloadManager;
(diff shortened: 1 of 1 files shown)


CassioMG:
it should user v7.6.0 as base branch

Issue #1719, opened 2020-10-26
OriginWhitelist isn't working for deeplinks on Android


gabrielmoncea:
**Bug description:**

Adding the url scheme of the app in `originWhitelist` didn't prevent the WebView to throw the loading page error.
Having a deeplink inside a webview will always throw the following error on Android.

```
Encountered an error loading page {"canGoBack": true, "canGoForward": false, "code": -10, "description": "net::ERR_UNKNOWN_URL_SCHEME", "loading": false, "target": 163, "title": "Webpage not available", "url": "urbanair://whereTo"}
```

The workaround for this was to call `Linking.openURL()` and then `return false` inside `onShouldStartLoadWithRequest` when the `request.url` starts with the url scheme. The error will still show in the webview, but shortly after it opens the deeplink.

**To Reproduce:**

Add the apps deeplink inside the webview and then press it.

**Expected behavior:**

The webview should open the app directly through the deeplink.

**Screenshots/Videos:**

![IMG_4E6127203179-1](https://user-images.githubusercontent.com/39256258/97160599-4293be00-1785-11eb-98a0-2cfd6b029be5.jpeg)

**Environment:**
 - OS: MacOS
 - OS version: `10.15.7 (19H2)` 
 - react-native version: `0.63.3`
 - react-native-webview version: `10.9.3`

nascode:
i seconded this. i tried to open whatsapp deeplink url.

klandell:
Same issue here.

amagnolo:
this might be related to #1741

gabrielmoncea:
Don't close

a613:
how can all these fundamental critical bugs go ignored so long...

sasindroid:
Did anyone get a resolution to this please?

amagnolo:
It seems that the project maintainers don't realize how impacting this bug is, in terms of people affected and issue severity.

j-piasecki:
If I understood the issue correctly, this is intended behaviour. According to the [documentation](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#originwhitelist) if the origin is added to the whitelist, it is handled by the `Webview`, otherwise it is handled by the OS. Can you try removing the origin from whitelist to see if it solves the problem?
[...]

Issue #1135, opened 2020-01-11
 Doesn't support some URL schemes


Visakeswaran:
**Bug description:**
I am using react-native-community/react-native-webview, to render a web view in my application. My URL scheme looks something like intent://cdn.glitch.com/.... (which is basically a WebAR, that basically opens a 3D model in the space around with help of camera). It throws an error saying ERROR_UNKNOWN_URL_SCHEME.

**To Reproduce:**
Add source uri as **'intent://user?screen_name=paul_kinlan#Intent;scheme=twitter;S.browser_fallback_url=https%3A%2F%2Ftwitter.com%2Fpaul_kinlan;end'**

**Expected behavior:**
Should load the url content

**Screenshots/Videos:**
[https://ibb.co/zNb9H7P](url)

**Environment:**
 - OS: Mac OS
 - OS version: Mojave
 - react-native version: 0.61.5
 - react-native-webview version: 8.0.1

thekingrenz23:
do you solved this problem?

j-piasecki:
https://github.com/facebook/react-native/pull/31836

PR #1760, opened 2020-11-26
Bugfix for Unknown URL Scheme in Android

lrusso:
**Bugfix for "Unknown URL Scheme" error in Android.**

In the demo code below, I'm using the **nativejstn** scheme as an example, it could be anything. The custom URL scheme must be added into the **originWhitelist** array in order to work.

````javascript
const _onShouldStartLoadWithRequest = (request) => {
    if (request.url.indexOf('http') === 0) {
        return true
        } else if (request.url.indexOf('nativejstn') === 0) {
        console.log(request.url)
        }
    return false
}

return (
    <View style={{width:'100%',height:'100%'}}>
        <WebView
            source={{ uri: 'https://www.example.com' }}
            onShouldStartLoadWithRequest={_onShouldStartLoadWithRequest.bind(this)}
            originWhitelist={[
                    'https://*',
                    'http://*',
                    'nativejstn://*',
                    'about:*',
                ]}
            />
        </View>
    )
````

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -142,7 +142,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   // Use `webView.loadUrl("about:blank")` to reliably reset the view
   // state and release page resources (including any running JavaScript).
   protected static final String BLANK_URL = "about:blank";
-  protected static final int SHOULD_OVERRIDE_URL_LOADING_TIMEOUT = 250;
+  protected static final int SHOULD_OVERRIDE_URL_LOADING_TIMEOUT = 1500;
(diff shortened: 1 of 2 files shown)


TheAlmightyBob:
@lrusso It's not clear to me what bug this is supposed to fix or how it fixes it? Is there an open issue that discusses the problem?

lrusso:
I found out that increasing that delay makes the URL schemes to work in Android. Otherwise, you get the ERR_UNKNOWN_URL_SCHEME message in the WebView.

https://github.com/react-native-webview/react-native-webview/issues/1719
https://github.com/react-native-webview/react-native-webview/issues/1135

whitefang57:
@lrusso does increasing that timeout fix the issue in most cases? I'm running into a similar issue and would love a potential fix

lrusso:
Yes, it works with all Android devices. Increase the timeout if you see that in some device is not working. It is the only way to fix this issue.

PR #1768, opened 2020-11-30
Upgrade react-native-windows to 0.63.11

jamonholmgren:
This should fix the failing Windows CI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "react": "16.11.0",
     "react-native": "0.62.2",
     "react-native-macos": "0.60.0-microsoft.73",
-    "react-native-windows": "^0.63.11",
+    "react-native-windows": "0.62.17",
     "selenium-appium": "0.0.15",
     "selenium-webdriver": "4.0.0-alpha.7",
     "semantic-release": "15.13.24",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -65,12 +66,11 @@
...
-    "react-native-windows": "^0.62.5",
-    "semantic-release": "15.13.24",
-    "typescript": "3.8.3",
-    "appium": "1.17.0",
+    "react-native-windows": "^0.63.11",
     "selenium-appium": "0.0.15",
-    "selenium-webdriver": "4.0.0-alpha.7"
+    "selenium-webdriver": "4.0.0-alpha.7",
+    "semantic-release": "15.13.24",
+    "typescript": "3.8.3"
```
jamonholmgren, line 73: appium, selenium, semantic-release, and typescript were just moved, not changed -- automatically I guess? Alphabetically now.

Only thing that changes is react-native-windows version from 0.62.5 to 0.63.11.

NickGerleman:
There might be some work in going from 0.62 to 0.63. Consider just updating to the latest patch version of 0.62 if you want to fix CI issues building. 

cc @jonthysell in case the auto-linking errors for webview are interesting.

jamonholmgren:
Downgraded to react-native-windows@0.62.17. We should figure out why 0.63.x doesn't work; don't want to fall too far behind on that.

zwang:
Hi guys, any plan about when to merge this and upgrade to 0.63.x? 

Thanks for the nice work.

PR #1776, opened 2020-12-08
Xcode 12.2 compat

ghiculescu:
Similar to https://github.com/react-native-webview/react-native-webview/pull/1643, but on Xcode 12.2 I needed to change to `React/Core` as per [this advice](https://github.com/facebook/react-native/issues/30018#issuecomment-716041423) to get the build working.

image


ghiculescu:
It looks like the Windows build is broken entirely (separate from this).

cristianoccazinsp:
I don't think this is right. Everywhere else (other libraries), the right value is `React-Core`.  Why would it be `React/Core` for this one?

ghiculescu:
Ahhhh, nevermind. Sorry. I think we are on an old react version and that's what's causing this.

PR #1817, opened 2021-01-08
fix: add onBlobFileDownload to prevent crash when open link start with blob in android

twisger:
Similar with #1546 and #1760, But adding a new api(onBlobFileDownload) to pass through url of blob link for resolving it outside.

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -17,8 +17,6 @@ jobs:
 
     - name: Setup MSBuild
       uses: microsoft/setup-msbuild@v1.0.2
-      with:
-        vs-version: 16.5
 
     - name: Check node modules cache
       uses: actions/cache@v1


stefanchiriac:
this is exactly what I need for handling the blob files downloads. I hope it gets merged soon.

worstpractice:
@twisger Great work!

@jamonholmgren  Looks like CI for Windows failed on the mention of `__builtin_clz`, which is a GNU builtin (GCC extension) that is undefined on Microsoft compilers.

GCC [defines this builtin](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) as:

> Built-in Function: _int_ **__builtin_clz** _(unsigned int x)_
> 
> Returns the number of leading 0-bits in _x_, starting at the most significant bit position. If _x_ is 0, the result is undefined.

Titozzz:
Being the devil's advocate here but isn't it possible to match iOS behavior maybe ? Would be best for the users if they don't have to implement differently for each platform

twisger:
> Being the devil's advocate here but isn't it possible to match iOS behavior maybe ? Would be best for the users if they don't have to implement differently for each platform

I tried to make it trigger `onShouldStartLoadWithRequest` in android and success, but it will cause a cycle call. `onShouldStartLoadWithRequest` will trigger download event again.Maybe adding some tags to request can resolve this issue, but I prefer adding a new API.

dvicory:
I'm not a fan of having to expand the API surface either, especially since ultimately we want to trigger a download just like any other file (and this would still leave a lot for library users to implement - after all, what do you do with a blob URL?).

What do you think about adding a JS interface to the page that can be exploited to handle the blob URL and funnel the data back to native and handle the download seamlessly as though it were any other? I would be willing to give a stab at the implementation.

twisger:
@dvicory In my case, the blob link is a pdf file, I just want to show it in the app but not download it, So I just get the blob link and fetch it in js,then load it by pdf viewer. And it is not difficult if I want to download directly,just need build a download request manually.
Actually, if `onShouldStartLoadWithRequest ` work well with blob link, I don't needto add a new API. but I don't find a good way to make `onShouldStartLoadWithRequest` to  intercept blob link.

twisger:
@dvicory You are right, I also have to fetch the blob in JS of webview.It seems impossible to fetch it in native code because of impletention of blob link.

About making blob link can be downloaded correctly, I think it also a acceptable fix way. And in most case it maybe a better choice.
It will be helpfull if you can send a PR about it.

Issue #124, opened 2018-11-01
Did not receive response to shouldStartLoad in time, defaulting to YES


dlferro:
Can someone give me a bit more details on why I would get the above warning?

Titozzz:
I've never seen that before, can you give us a way to repro this ?

dlferro:
It appears to exist in the core Webview as well.

https://github.com/react-native-community/react-native-webview/blob/7f3534463291893154bd41dc560c092a7b9bf746/ios/RNCWKWebViewManager.m#L157

My component presenting it is just
```
import React, { PureComponent } from 'react';
import { WebView } from 'react-native-webview';

class CloudPageScreen extends PureComponent {
  render() {
    const { navigation } = this.props;
    const uri = navigation.getParam('uri');

    return <WebView source={{ uri }} />;
  }
}

export default CloudPageScreen;
```

Titozzz:
We have a current PR https://github.com/react-native-community/react-native-webview/pull/107 trying to implement this on android and fix iOS, hopefuly it will help

zmnv:
Is it a some critical thing or I can skip it for your updates?

MickaelNeves:
Same here, can I ignore those warnings? I try to found more info about them but no success.

RN: 0.57.8
RN Webview: 5.2.1

![captura de ecra 2019-02-20 as 15 51 26](https://user-images.githubusercontent.com/20476271/53104719-cd8bd200-3527-11e9-9e35-d23d50741323.png)

zmnv:
I can ignore it if there are 3-4 warnings like that.
But... more code = more warning messages..

Titozzz:
If you have chrome debugger on it's 'normal' cause it's too slow

bosshog:
even with chrome debugger off I see a lot of these warnings (on iOS)

IlinIgor:
Are there any fixes of that?

Johncy1997:
I am getting the same warning and can't able to view the web page.Is there any fixes?

scarlac:
Error is caused by app being slow. It attempts to execute the callback defined for `onShouldStartLoadWithRequest`, an event handler that should return a boolean (aka `YES`/`NO`) indicating if loading of a given URL should actually happen, but if that function does not return within 250ms, it'll still load the page. Even if you define `onShouldStartLoadWithRequest={() => true}` it will still give this warning.

[...]

PR #1840, opened 2021-01-25
fix: Did not receive response to shouldStartLoad in time, defaulting to YES

hqdai:
I've created a prop **blockMainThreadEnabled** for iOS only to disable the action blocking main thread for 250ms by default. User should be able to enable/disable this action to prevent the warning message.

This issue has reported at: https://github.com/react-native-webview/react-native-webview/issues/124

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -1091,6 +1091,17 @@ Example:
 
 `<WebView textZoom={100} />`
 
+---
+### `blockMainThreadEnabled`
+
+The default value is true. If the value of this property is false, iOS main thread will not be blocked for a maximum of 250ms until the JS thread returns. 
+
+| Type    | Required | Platform |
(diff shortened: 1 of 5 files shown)


Titozzz:
What would be the goal of such a prop ? Have you had any issues in production with this ?

hqdai:
> What would be the goal of such a prop ? Have you had any issues in production with this ?

oh yes, as I mention in PR description, the issue has reported at: https://github.com/react-native-webview/react-native-webview/issues/124
This PR will by pass the block main thread on iOS.

keegandonley:
@Titozzz This would be super nice for development as the warning popping up constantly gets in the way, even though I'm not using a custom shouldStartLoad so don't care about the warning

keegandonley:
Still interested in something to fix this issue!

Issue #1869, opened 2021-02-09
onShouldStartLoadWithRequest Not Working on Android


topherPedersen:
**Bug description:**
On Android, the onShouldStartLoadWithRequest method is not working. For example, we load a single page application from a third party website into our WebView, while navigating the single page application the onShouldStartLoadWithRequest method never executes. However, loading the same URL into a native Android WebView does cause the onShouldStartLoadWithRequest method to execute.

**To Reproduce**
Load a single page application into react-native-webview for Android and see if the onShouldStartLoadWithRequest method ever executes.

**Expected behavior:**
The onShouldStartLoadWithRequest method should be called just like it would if one were using a native android.webkit.WebView and android.webkit.WebViewClient. For example, I created a native Android WebView to experiment working with this particular single page application, and the onShouldStartLoadWithRequest method is called as expected. But this behavior is not matched when using react-native-webview. (Sorry I am not able to provide the exact link to the single page application as it's a proprietary, closed-source, and not publicly available)

**Environment:**
- OS: Android 9
- react-native version: 0.63.4
- react-native-webview version: 11.2.1 / 11.0.2

mzain512:
I am facing the same issue. I have loading my website and `onShouldStartLoadWithRequest ` never calls on first launch and whenever I click on links in webpage. using v11.0.2

rpander93:
Looks like #1855 that I reported some days ago

davidohlin:
@mzain512 I missed it too, but the [documentation](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#onshouldstartloadwithrequest) for `onShouldStartLoadWithRequest` say:

> On Android, is not called on the first load.

>

mzain512:
@davidohlin ok we can ignore the first launch. What about after that when we click on any of the links in webpage? On iOS I have handled external links through this callback but on android its not called and my implementation is failing.

yuvapoojary:
Yes it is not working in single page web application.
And onNavigationStateChange is also not working .

accprudolfs:
The same issue onShouldStartLoadWithRequest is not called at all on Android

creativemind1:
Only few requests are loading and not all with `onShouldStartLoadWithRequest` (Android)
`"react-native-webview": "^8.1.1",`

martinhellwig:
[...]

Issue #1785, opened 2020-12-11
onNavigationStateChange(event): event.navigationType is undefined on Android


bastienJanon:
**Bug description:**
onShouldStartLoadWithRequest(event => myFn(event))
event.navigationType is undefined on android.

**To Reproduce:**
const html =   `<html><head><meta name="viewport" content="width=device-width, initial-scale=1.0"></head><body><a href="https://github.com/react-native-webview/react-native-webview">open link</a></body></html>`;

<WebView
      originWhitelist={['*']}
      source={{ html }}
      scrollEnabled={false}
      onShouldStartLoadWithRequest={myFn}
      onNavigationStateChange={myFn2}
      style={{ 300 }}
    />
On Android 10 api 30 (and 9 api 28) same result: callback from onShouldStartLoadWithRequest event.navigationType == undefined but on IOS all seems to be ok.
related issue: https://github.com/react-native-webview/react-native-webview/issues/1314

**Expected behavior:**
event.navigationType should be 'other' or 'click'

**Environment:**
 - OS: macOs Catalina
 - OS version: 10.15.7
 - react-native version: 0.60.5
 - react-native-webview version: latest release so 11.0.2

bastienJanon:
this doesn't happened on rn-webview 8.0.4 so i revert rn-webview version to 8.0.4

Issue #1314, opened 2020-04-10
'navigationType' property is missing from the event passed to 'onNavigationStateChange' on Android


iancucatalin1993:
**Bug description:**
The WebView navigation listener 'onNavigationStateChange' only has the following properties on Android
```
{
  canGoForward: false
  canGoBack: false
  title: ""
  loading: false
  url: "about:blank"
  target: 3723
}
```

According to the following library code fragments ```navigationType``` shoud be present and have one of the defined values:
```
export interface AndroidWebViewProps extends WebViewSharedProps {
    onNavigationStateChange?: (event: WebViewNavigation) => void;
.....
}
export interface WebViewNavigation extends WebViewNativeEvent {
    navigationType: 'click' | 'formsubmit' | 'backforward' | 'reload' | 'formresubmit' | 'other';
    mainDocumentURL?: string;
}
```

**As it is, there is no proper way that I know of to determine whether the user clicked on something to initiate a navigation, or the redirect was automatic as it often happens with social media embeds.**

I don't have an iOS build, but according to other iOS developers it seems to be working there.

**To Reproduce:**
Assign `onNavigationStateChange` listener to a WebView on Android. The property navigationType is missing entirely
```
<WebView
        onNavigationStateChange={this.onNavigationStateChange}
        {...this.props}
 />

onNavigationStateChange = (navState) =>
{
   //  typeof navState.navigationType === 'undefined'
}
```

**Expected behavior:**

```
onNavigationStateChange = (navState) =>
{
    if (navState.navigationType === 'click') {
        // User clicked something
    }
}
```

**Environment:**
 - OS: Android
 - OS version: any
 - react-native version: <= 0.62.2
 - react-native-webview version: <= 9.1.3

DmitryFt:
Also having this issue, RN 0.63.3, RNWebview 10.9.2

bastienJanon:
i'm facing the same issue. When the debugger is active event.navigationType return 'click' as expected but when debugger is off event.navigationType is undefined.

JohnTsaiAndroid:
Any solutions?

VanLaerAshley:
any solutions?

VanLaerAshley:
This should be open

peterdee:
Having the same issue with Android. iOS works fine.

I am using React Native with Expo (Expo SDK 40).

erkie:
It's iOS only.  https://github.com/react-native-webview/react-native-webview/pull/1880

[...]

PR #1881, opened 2021-02-16
Implement a way to see if `shouldStartLoadWithRequest` was initiated by user gesture on Android

erkie:
`navigationType` is an iOS attribute that can't be replicated on Android. One idea was to just use `if hasGesture then navigationType = click`, but as a user of this library I would use it assuming it would work the same on both Android and iOS.  `navigationType` has a lot of different values for different scenarios on iOS that Android doesnt provide. That's why I propose documenting it as such, and using the Android-ism `hasGesture` for this case.

Fixes https://github.com/react-native-webview/react-native-webview/issues/1869
Fixes https://github.com/react-native-webview/react-native-webview/issues/1785

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -847,6 +847,10 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
     @Override
     public boolean shouldOverrideUrlLoading(WebView view, String url) {
+      return shouldOverrideUrlLoading(view, url, null);
+    }
+
+    public boolean shouldOverrideUrlLoading(WebView view, String url, @Nullable WebResourceRequest request) {
       final RNCWebView rncWebView = (RNCWebView) view;
       final boolean isJsDebugging = ((ReactContext) view.getContext()).getJavaScriptContextHolder().get() == 0;
 
@@ -857,6 +861,10 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
         final WritableMap event = createWebViewEvent(view, url);
         event.putInt("lockIdentifier", lockIdentifier);
+        if (request != null) {
+          event.putBoolean("hasGesture", request.hasGesture());
+          event.putBoolean("isTopFrame", request.isForMainFrame());
+        }
         rncWebView.sendDirectMessage("onShouldStartLoadWithRequest", event);
 
         try {
@@ -898,7 +906,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
     @Override
     public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request) {
       final String url = request.getUrl().toString();
-      return this.shouldOverrideUrlLoading(view, url);
+      return this.shouldOverrideUrlLoading(view, url, request);
     }
 
     @Override
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -731,6 +731,7 @@ lockIdentifier
 mainDocumentURL (iOS only)
 navigationType (iOS only)
 isTopFrame (iOS only)
+hasGesture (Android only, API 21+)
 ```
 
 ---
--- a/src/WebViewTypes.ts
+++ b/src/WebViewTypes.ts
@@ -103,7 +103,7 @@ export interface WebViewNativeProgressEvent extends WebViewNativeEvent {
 }
 
 export interface WebViewNavigation extends WebViewNativeEvent {
-  navigationType:
+  navigationType?:
     | 'click'
     | 'formsubmit'
     | 'backforward'
@@ -115,6 +115,7 @@ export interface WebViewNavigation extends WebViewNativeEvent {
 
 export interface ShouldStartLoadRequest extends WebViewNavigation {
   isTopFrame: boolean;
+  hasGesture?: boolean;
 }
 
 export interface FileDownload {


TheAlmightyBob:
I looked into doing something like this about 2 years ago, but at the time my testing found that `hasGesture` was almost always `true` (i.e. for cases where iOS would _not_ have a `navigationType` of `click`). So I definitely agree with not trying to merge this with `navigationType`.

I'm curious if you've actually found `hasGesture` to be informative...?

How is this related to #1869? I don't see any mention of `navigationType`/gestures being a concern there.

I don't think this addresses #1785 either, it looks like that's a regression related to the new synchronous `onShouldStartLoadWithRequest` introduced in #1590. The event it fires no longer uses the existing `TopShouldStartLoadWithRequestEvent`, and is thus missing parameters that were previously included. The previous event is still used as a fallback though, which I'd guess is why #1314 noted that
> When the debugger is active event.navigationType return 'click' as expected but when debugger is off event.navigationType is undefined

I'm not opposed to adding `hasGesture` if you find it useful, but the two code paths should ideally be made consistent (I'm not sure if there's a technical reason that `TopShouldStartLoadWithRequestEvent` couldn't be used in the synchronous path??), which I'd _expect_ also means bringing back `navigationType` but setting it to `other` as was the previous/documented behavior.

erkie:
> I'm curious if you've actually found hasGesture to be informative...?

Yes, we use it to detect clicks on external links. It works great and it's been in production for a while. I remember reading in the documentation about it that it has historically been a bit buggy, but from my understanding it works well now.

Regarding `Fixes #1785`, I honestly can remember why this I wrote that it fixes #1785...

TheAlmightyBob:
I didn't notice your edit until now... GitHub doesn't notify of that.

> And just to clarify, to get this merged, what changes should I be making? Thanks!

I'm not an "owner" so can't promise merging, but I do think that the two places that raise `onShouldStartLoadWithRequest` in `shouldOverrideUrlLoading` should be consistent.

There's that big `if` condition at the top: https://github.com/react-native-webview/react-native-webview/blob/1b766c48fb3ba13aab18d175c8efafa77000dafd/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L857

If that's `true`, then it uses `sendDirectMessage` to raise `onShouldStartLoadWithRequest`, which is where you added parameters.

But if it's false, then it uses the "old way" of dispatching `TopShouldStartLoadWithRequestEvent`, which defines its own parameters (and these are the parameters defined by the Typescript types and documentation).

Right now, the `sendDirectMessage` approach is missing parameters from `TopShouldStartLoadWithRequestEvent`, and `TopShouldStartLoadWithRequestEvent` is missing your new `hasGesture` parameter. So `onShouldStartLoadWithRequest` will have different parameters depending on that `if` statement.

_Ideally_ I would expect `sendDirectMessage` to itself use `TopShouldStartLoadWithRequestEvent`.... but if there's some technical reason these are incompatible, then the parameters could be manually synchronized (perhaps with a comment in code pointing out that they should be kept in sync).

This would have the bonus effect that you could legitimately say that this fixes #1785 ;)

---

I think that this would include having `sendDirectMessage` mimic the old behavior from `TopShouldStartLoadWithRequestEvent` of setting `navigationType` to `other` rather than making it optional. Though since the breaking change of removing it sort of already happened, you _could_ instead commit to that and remove it from `TopShouldStartLoadWithRequestEvent` also, and update the docs/types/comments accordingly.

e.g. https://github.com/react-native-webview/react-native-webview/blob/d6672c87eb61827c9b0215733a4766c14f68d01a/src/WebViewTypes.ts#L1089

erkie:
This is still a valid PR imo, I just don't have the time and resources to get in fixed for merge. :(

nishantsambyal123:
shouldStartLoadWithRequest start working after  setSupportMultipleWindows={false} on Android

`<WebView
          style={styles.webView}
          source={{uri: uri}}
          onNavigationStateChange={this.handleNavigationStateChange.bind(this)}
          javaScriptEnabled={true}
          ref={webView => (this.webView = webView)}
          //ref={this.myRef}
          sharedCookiesEnabled={true}
          pullToRefreshEnabled
          bounces={false}
          cacheEnabled={true}
          allowFileAccessFromFileURLs={true}
          allowUniversalAccessFromFileURLs={true}
          setSupportMultipleWindows={false}
          onMessage={event => this.onMessage(event)}
          onLoad={this.biometricTest}
          // onLoadStart={this.onLoadStart}
          onShouldStartLoadWithRequest={request => {
            console.log('onShouldStartLoadWithRequest called');
            if (request.url.includes('links')) {
              this.setState({currentUrl: request.url}, () => {
                console.log('onShouldStartLoadWithRequest Called', request);
                this.onLoadStart();
              });
              return false;
            }
            return true;
          }}
        />`

dwardin:
I have got to say - excellent work @erkie! Saved me a lot of headaches with this. Hopefully this makes it into the core repo - happy to help with this if you'd like

erkie:
@dwardin glad it saved you time! I'd still love to get it merged, we still use this in production. So if you want to take over the branch and resolve the comments by @TheAlmightyBob I haven't even had time to try to understand what they request I'm afraid.

dwardin:
More than happy to! Will have a look at it in the next week or two. I've taken your changes and applied them to the latest build via `patch-package` and it all works beautifully.

I believe @TheAlmightyBob's comments were pointing out some additional blast radius which should be considered for completeness and consistency with regards to the WebView library as a whole. Basically a few other places in the library that need these parameters to be applied consistently to what your changes already do in one area of the library.

erkie:
Sounds great :D if you create a new PR please link it here so I can follow the progress. Thanks for the help!

TheAlmightyBob:
(I'm also still around if you have questions.... sorry that my comments were hard to understand 😞  )

erkie:
> (I'm also still around if you have questions.... sorry that my comments were hard to understand 😞  )

Sorry, I didn't mean to sound critical! I was in the middle of the deadline to get our app out when I made this PR, so was a bit overwelmed by tasks left and right. My inbox was swamped and just got pushed down until stalebot came in. 😅

erkie:
@shukerullah  I just checked the deps on our react-native app and not seeing us using this fork anymore. I think it got overwritten in a React upgrade... We're still trying to use `hasGesture` so I guess that part of our app us just now buggy.

PR #1916, opened 2021-03-15
fix(iOS): webview memory leak

sunzhongliangde:
# fix(iOS): webview memory leak

 

When close the webview, dealloc function does not execute, there have a retain cycle in RNCWebView.m file

NO
Hit berakpoint in dealloc function，after the webview close，this function is not excute
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


Titozzz:
Thank you very much

PR #1925, opened 2021-03-19
- Adding some links for api references

hkar19:
I am adding some iOS references on how to use some props.


TheAlmightyBob:
Most of this looks okay, however for `keyboardDisplayRequiresUserAction` you're linking to the `UIWebView` doc, which could be confusing since `react-native-webview` doesn't use UIWebView..... I get that the docs explain the concept, but it might be better to add any missing info to the `react-native-webview` doc directly, or at least explain that the linked doc just describes the behavior that `react-native-webview` is trying to mimic.

Issue #1938, opened 2021-03-23
dispatchEvent method of RNCWebviewManager in Android from static to non static


gkode:
**Is your feature request related to a problem? If so, Please describe.**

I'm working on Custom RNCWebView (extending RNCWebViewManager), When I init Custom RNCWebView, i don't have react context available, so I use activity context. This scenario I am able to handle in my own code. 
But  dispatchEvent is created as static method, it is causing an issue, as I can't override this method in my custom RNCWebview. 
I see there is no harm in creating dispatchEvent method as non static and can be added to RNCWebView object itself. I tested this, it's working fine.

PR #1939, opened 2021-03-23
Converting dispatchEvent static method to non static method

gkode:
Create Issue#https://github.com/react-native-webview/react-native-webview/issues/1938


gkode:
@jamonholmgren Can you please review. its just refactoring. I was stuck in one of the scenario where i was not able to override this method while working on some customrncwebviewmanager. it doesn't impact the functionality of webview and provides a way to override dispatchevent method if required.

Issue #920, opened 2019-09-30
Apple pay in webview


MaksymLebedevs:
Is it possible to use Apple Pay on pages opened with react-native-webview? As I understand to use it on pages they should be opened in SFSafariViewController, while react-native-webview uses WKWebView. Is it possible to switch to SFSafariViewController?

meshi:
i'm following this issue

BilalArxd:
I am having same issue using stripe with Apple Pay and Google Pay. Any solution or workaround would be appreciated. Thanks

ianneub:
I updated our app to react-native-webview 7.0.5 from react-native's 0.42 version. We found that Apple Pay now works in our webview app! Later on I added an `onMessage` handler to the WebView and now Apple Pay does not work.

I assume this is another issue, but does anyone else using the `onMessage` handler? Maybe that is why you don't get Apple Pay?

esauro:
I'm having the same issue. I updated to react-native-webview 7.4.3 from react-native 0.60. It didn't work. We have an onMessage handler, I removed the onMessage and it's still not working.

I will try the same with a SFViewController, but that won't provide the UX I want to provide

shusain93:
> In iOS 13, webpages loaded in WKWebView can now accept Apple Pay. In order to protect the security of Apple Pay transactions in WKWebView, Apple Pay cannot be used alongside of script injection APIs such as WKUserScript or evaluateJavaScript(_:completionHandler:).
> 
> If these APIs are invoked before a webpage uses Apple Pay, Apple Pay will be disabled. If a webpage uses Apple Pay before evaluateJavaScript(_:completionHandler:) is invoked, the completion handler will be called with a non-nil NSError. These restrictions are reset every time the top frame is navigated.

https://webkit.org/blog/9674/new-webkit-features-in-safari-13/

hmolotsi:
I'm having this same issue and I'm using react-native-webview 8.0.6. Has anyone gotten this to work? What WebView parameters are required?

pxr13:
I got it to work by putting all the Apple Pay processing code on a page in my web app and using that as the `uri` in the `source` prop of the `WebView` component.

You'll need to put the button and Apple Pay session code in the web app for this to work and don't use the `injectedJavascript` prop in the `WebView` component because that will disable Apple Pay (as the post by @shusain93 shows).

temitope:
[...]

PR #1946, opened 2021-03-31
Apple Pay support in iOS

netmaxt3r:
Add new prop `enableApplePay` for ios which disables script injection and allow Websites to invoke Apple Pay api.
This comes with a cost features like script injection,html5 History , shared Cookies, atEndScript , atStartScript will not work
[Apple Documentation](https://developer.apple.com/documentation/safari-release-notes/safari-13-release-notes#Payment-Request-API). We can still use postMessage provided  `window.postMessage` wont work `window.webkit.messageHandlers.ReactNativeWebView.postMessage` will work

possible fix for #920

### DIFF
--- a/apple/RNCWebView.h
+++ b/apple/RNCWebView.h
@@ -68,6 +68,7 @@
 @property (nonatomic, assign) BOOL ignoreSilentHardwareSwitch;
 @property (nonatomic, copy) NSString * _Nullable allowingReadAccessToURL;
 @property (nonatomic, assign) BOOL pullToRefreshEnabled;
+@property (nonatomic, assign) BOOL applePayCompactMode;
 #if !TARGET_OS_OSX
 @property (nonatomic, weak) UIRefreshControl * _Nullable refreshControl;
 #endif
--- a/apple/RNCWebView.h
+++ b/apple/RNCWebView.h
@@ -68,7 +68,7 @@
 @property (nonatomic, assign) BOOL ignoreSilentHardwareSwitch;
 @property (nonatomic, copy) NSString * _Nullable allowingReadAccessToURL;
 @property (nonatomic, assign) BOOL pullToRefreshEnabled;
-@property (nonatomic, assign) BOOL applePayCompactMode;
+@property (nonatomic, assign) BOOL enableApplePay;
 #if !TARGET_OS_OSX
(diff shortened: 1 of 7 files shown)


TheAlmightyBob:
Minor detail, but "compact" makes me think of visual layout rather than featureset. Perhaps something like `applePayRestrictedMode` would be more clear?

netmaxt3r:
@TheAlmightyBob   would `applePayRestrictedMode`  hints "apple Pay" is restricted, i will change this prop name to simple `enableApplePay`

TheAlmightyBob:
> `applePayRestrictedMode` hints "apple Pay" is restricted

Yeah, that's fair. Naming is hard.

Could you update https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md?

TheAlmightyBob (approved):

chuanxie:
Hi @netmaxt3r when i run the following example code, i am still get `your browser does not support apple pay on the web`, do you have any ideas? or something need to be addressed? thanks
   
```
<WebView
   enableApplePay={true}
   source={{uri: "https://applepaydemo.apple.com/"}} 
    automaticallyAdjustContentInsets={false}>
```
I am using the expo 41 and latest react native webview on 11.6.1

chuanxie:
Screenshot 2021-05-24 at 14 54 29
so apple pay works on pure react native but not expo @netmaxt3r

netmaxt3r:
@chuanxie  I do not have hands on with expo. All this patch does is remove script injection in webview, are you injecting any script in your webview?. let me take look at expo 41 let you know

netmaxt3r:
@chuanxie since this change has native code changes and expo 41 has bundled [react-native-webview 11.2.3](https://github.com/expo/expo/blob/master/CHANGELOG.md#4100--2021-03-10) this wont be possible in a "managed" workflow. However in "bare" workflow use react-native-webview@11.6.0 up ( default is 11.2.3 ) it is working.

chuanxie:
thanks @netmaxt3r  . no,  i didnt inject any scripts. just the braintree payment iframe from my web page.
hmm.. i forced the expo to use the react-native-webview@11.6.1 and its not complaining about it

Titozzz:
It's not complaining but it wont work either because the native code just isn't here!

Issue #588, opened 2019-05-22
Event Listener does not trigger any events


mjfiandaca:
**Bug description:**
Events are never triggered when attaching an event listener to a webview via <script>, or injecting javascript.

**To Reproduce:**
Associate a script to a webview.
This can be done by: a) adding the <script> tag to the html being output.
b) using injectjavascript.
c) using injectedjavascript (I am not using this as it is not reliable to know where it will be hit in the render cycle, potentially being added after the event I am looking for has already fired).

In my example code, I have attempted to include all iterations of getting this to work, commented out.  Ultimately, I want to write a message to the window when the webview has completed loading so that it can pass the window height to React Native.  I have a workaround in that I am using onDoneLoading to inject the javascript to obtain this information; however, I want to know why the event listener doesn't work.  I attempted to replace the 'DOMContentLoaded' with 'click' or 'scroll' to see if it was just that event that wasn't firing, but nothing seems to work.

Additionally, when I do receive the dimensions of the webview, it seems to be off using the document.documentElement.clientHeight values. I am unsure if there is additional data I need to obtain to reflect the appropriate height for my enclosing webview.

Source code here: [https://github.com/mjfiandaca/Playground](url)

**Expected behavior:**
The webview should fire events if the event listener has been added to the script.

**Environment:**
 - OS: Android
 - OS version:
 - react-native version: 0.57.8
 - react-native-webview version: 5.10.0

mjfiandaca:
For some reason, the link does not go to my github even though pasting it into the browser works fine, so here is the code: 
```import React, { Component } from 'react';
import { View, StyleSheet } from 'react-native';
import { WebView as RNWebView } from 'react-native-webview';
import { WebViewSourceHtml } from 'react-native-webview/lib/WebViewTypes';

type Props = {
  onHeightChange?: (height: string) => void;
  height?: number;
  // TODO: Source is currently hard-coded for proof of concept.
  // Eventually, it will be the Html source as coming from the API.
  source?: WebViewSourceHtml;
  // TODO: until the API is in place, we might want to pass in the message text and create
  // an 'html' file.
  text?: string;
};

class WebView extends Component<Props> {
  // The following private ref & 'run' are used if you want to injectjavascript.
  private webref: any = '';
  render() {
    const run = `
    const data = 'hello there!';
    window.ReactNativeWebView.postMessage(data);
      true;
    `;

    // Using this for injectedjavascript will return the window size - but it is
[...]

PR #1953, opened 2021-04-08
Add Autolinking Support for Windows

chiaramooney:
**_Why_**
Module needs a special configuration to properly perform autolinking for Windows because of the two project module structure.

**_What_**
Altered `react-native.config.js` to set up proper configuration for Windows autolinking. Updated app certificate as original certificate was expired. Removed blacklist mentions as they cause the build of the example app to fail. 

Tested that new example app configuration works for Windows and android. However, I don't have the development environment to test iOS or macOS.


asklar (commented):
@acoates-ms to take a look too. Starting to take a look myself as well.

example/react-native.config.js
```diff
@@ -0,0 +1,48 @@
...
+const path = require('path');
+
+const macSwitch = '--use-react-native-macos';
+const windowsSwitch = '--use-react-native-windows';
+
+if (process.argv.includes(macSwitch)) {
+  process.argv = process.argv.filter(arg => arg !== macSwitch);
+  process.argv.push('--config=../metro.config.macos.js');
+  module.exports = {
+    reactNativePath: '../node_modules/react-native-macos',
```
asklar, line 27: I believe require.resolve would do the right thing here but we weren't able to get it working properly. I'm not too worried about this one being a little more hardcoded since it is an example app only but it would be good to figure out the right pattern that doesn't hardcode these.

metro.config.windows.js
```diff
@@ -6,13 +6,12 @@
  */
 const fs = require('fs');
 const path = require('path');
-const blacklist = require('metro-config/src/defaults/blacklist');
 
 const rnPath = fs.realpathSync(
```
asklar, line 10: shouldn't this file go into the example folder?

package.json
```diff
@@ -1,21 +1,15 @@
 {
   "name": "react-native-webview",
   "description": "React Native WebView component for iOS, Android, macOS, and Windows",
-  "main": "index.js",
-  "typings": "index.d.ts",
   "author": "Jamon Holmgren <jamon@infinite.red>",
   "contributors": [
     "Thibault Malbranche <malbranche.thibault@gmail.com>"
   ],
   "license": "MIT",
```
asklar, line 8: since we are changing the repo's folder structure, you will likely want to take a pass at updating the repo's CI to make sure the right commands run in the right folders

src/package.json
```diff
@@ -0,0 +1,77 @@
...
+      "start:ios": "react-native run-ios",
+      "start:macos": "node node_modules/react-native-macos/local-cli/cli.js start --use-react-native-macos",
+      "start:windows": "react-native start --use-react-native-windows",
+      "ci": "CI=true && yarn lint",
+      "ci:publish": "yarn semantic-release",
+      "lint": "yarn tsc --noEmit && yarn eslint ./src --ext .ts,.tsx",
+      "build": "yarn tsc",
+      "prepare": "yarn build",
+      "appium": "appium",
+      "test:windows": "yarn jest --setupFiles=./jest-setups/jest.setup.windows.js"
```
asklar, line 26: a lot of these are probably for the example app only (the starts, and I think the appium and tests)

yarn.lock
```diff
@@ -9161,33 +9162,32 @@ metro@0.58.0, metro@^0.58.0:
...
     source-map "^0.5.6"
-    strip-ansi "^4.0.0"
     temp "0.8.3"
     throat "^4.1.0"
     wordwrap "^1.0.0"
     write-file-atomic "^1.2.0"
     ws "^1.1.5"
     xpipe "^1.0.5"
-    yargs "^14.2.0"
+    yargs "^9.0.0"
```
asklar, line 9190: seems weird that a number of these got downgraded?

src/package.json
```diff
@@ -0,0 +1,77 @@
...
+      "eslint-plugin-react-native": "3.7.0",
+      "jest": "24.9.0",
+      "metro": "^0.56.4",
+      "metro-react-native-babel-preset": "^0.59.0",
+      "react": "16.11.0",
+      "react-native": "0.62.2",
+      "react-native-macos": "0.60.0-microsoft.73",
+      "react-native-windows": "0.62.17",
+      "selenium-appium": "0.0.15",
+      "selenium-webdriver": "4.0.0-alpha.7",
```
asklar, line 72: if the tests are just using the example app then maybe it'll be worth moving these to the example

example/android/app/src/main/java/com/example/MainApplication.java
```diff
@@ -29,7 +29,6 @@ public boolean getUseDeveloperSupport() {
           List<ReactPackage> packages = new PackageList(this).getPackages();
           // Packages that cannot be autolinked yet can be added manually here, for example:
           // packages.add(new MyReactNativePackage());
-          packages.add(new RNCWebViewPackage());
```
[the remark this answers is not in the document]
chiaramooney: Yes! If the line is left in there is an error message that states that the package was added twice. Once by autolinking and once by this manual entry.

metro.config.windows.js
chiaramooney: There was a note on the RNW website that mentioned that metro.config.windows.js should be in the root of the repo. "Add metro.config.windows for Windows bundling (example). Make sure the config file is at the root of the repo (see Metro bug #588)." I can test to see if the code still works when the files are moved?

yarn.lock
chiaramooney: Hmm weird the only dependency I touched was slightly bumping up the metro version.

metro.config.js
```diff
@@ -3,10 +3,7 @@
  * integration tests during local development or on CI services.
  */
 
-const blacklist = require('metro-config/src/defaults/blacklist');
```
acoates-ms, line 6: Why are you removing all these?  -- I suspect the macos builds may still need them.

chiaramooney: Sync'd offline.

asklar (approved):

asklar:
LGTM, thanks @chiaramooney  - the only thing to do now is when we merge we should edit the commit message to match what semantic-release expects (see https://github.com/semantic-release/semantic-release)
Maybe something like `fix(autolink-windows): enable autolinking on Windows`
does that commit msg look right @jamonholmgren / @Titozzz ?

Titozzz:
fix(): for patch
feat(): for minor

Anything with BREAKING CHANGE: for major

PR #1994, opened 2021-05-20
Changing ThemedReactContext to ReactContext

gkode:
Requirement :

I create custom component by extending RNCWebViewManager where I pass reactApplicationContext instead of ThemedReactContext. Since, Constructor and other methods doesn't allow me to do it, I have to override everything and change RNCWebview library itself.


gkode:
@Titozzz  @jamonholmgren Could you please review it

gkode:
> Do you know why ThemedReactContext was used ?

I don't see any explicit use of it in RNCWebView? Could you please help if there is any or it may break something? RNCWebview will be still working as expected as ReactContext is parent of ThemedReactContext

Titozzz:
I guess we need to update that then: https://github.com/react-native-webview/react-native-webview/blob/master/docs/Custom-Android.md .
Can you do it?

gkode:
> I guess we need to update that then: https://github.com/react-native-webview/react-native-webview/blob/master/docs/Custom-Android.md .
> Can you do it?

Done. Updated documentation

gkode:
> I guess we need to update that then: https://github.com/react-native-webview/react-native-webview/blob/master/docs/Custom-Android.md .
> Can you do it?

gkode:
> > I guess we need to update that then: https://github.com/react-native-webview/react-native-webview/blob/master/docs/Custom-Android.md .
> > Can you do it?
> 
> Done. Updated documentation @Titozzz

@Titozzz  Can you please approve and merge

Titozzz (approved):

ifsnow:
@Titozzz @gkode Have you tried compiling with this fix? I faced on these errors.

```
RNCWebViewManager.java:119: error: RNCWebViewManager is not abstract and does not override abstract method createViewInstance(ThemedReactContext) in ViewManager
RNCWebViewManager.java:172: error: method does not override or implement a method from a supertype
```

Since [ViewManager](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewManager.java) is defined to use `ThemedReactContext`, It seems that this change doesn't work.

Titozzz:
@ifsnow I reverted the changes
@gkode It seems you did not test your changes properly, please submit another PR

PR #2016, opened 2021-05-25
fix #1353 

### LINKED ISSUE #1353: No visible @interface for 'NSHTTPURLResponse' declares the selector 'valueForHTTPHeaderField:'
**Bug description:**
`No visible @interface for 'NSHTTPURLResponse' declares the selector 'valueForHTTPHeaderField:'` issue while run `react-native run-ios` or in `xCode`

**To Reproduce:**
Once i remove `react-native-webview` app works without issues.

**Expected behavior:**
open dynamic link when button click.

**Screenshots/Videos:**

image

**Environment:**
 - OS: macOS
 - OS version: 10.13.6
 - react-native version: "0.62.2
 - react-native-webview version: "^9.4.0"

thanhly:
making sure that older xcode would compile by using pre-prcessor checks as @available(iOS 13, *) is not sufficient at compile time.

This fix also is consistent with the rest of the code where pre-processor checks are built in.

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -1036,7 +1036,7 @@ static NSDictionary* customCertificatesForHost;
       }
 
       NSString *disposition = nil;
-#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000 /* iOS 13 */ 
       if (@available(iOS 13, *)) {
         disposition = [response valueForHTTPHeaderField:@"Content-Disposition"];
       }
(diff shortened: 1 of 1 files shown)


apple/RNCWebView.m
```diff
@@ -1036,9 +1036,11 @@ - (void)                    webView:(WKWebView *)webView
       }
 
       NSString *disposition = nil;
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
```
TheAlmightyBob, line 1039: small detail, but it would be nice to add a comment to clarify this version check as is done elsewhere in `react-native-webview`... it may not always be obvious to folks what `130000` means

e.g.
https://github.com/react-native-webview/react-native-webview/blob/30685edda045f7ba5d1400b415278e1109436dcd/apple/RNCWebViewManager.m#L18

https://github.com/react-native-webview/react-native-webview/blob/30685edda045f7ba5d1400b415278e1109436dcd/apple/RNCWebViewManager.m#L81

TheAlmightyBob (approved):

apple/RNCWebView.m
thanhly: You're right, this is also consistent with the rest. I've added one commit to include the comment.

TheAlmightyBob (approved):

PR #2025, opened 2021-05-28
Avoid null pointer exception in RNCWebViewManager

agoyal2801:
There is a rare case when getCurrentActivity() is returning null and causing NullPointerException which leads to crash. Added check to avoid NullPointerException in RNCWebViewManager and added warning log in else block.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -707,88 +707,90 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
   protected void setupWebChromeClient(ReactContext reactContext, WebView webView) {
     if (mAllowsFullscreenVideo) {
-      int initialRequestedOrientation = reactContext.getCurrentActivity().getRequestedOrientation();
-      mWebChromeClient = new RNCWebChromeClient(reactContext, webView) {
-        @Override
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -707,88 +707,93 @@ public static RNCWebViewModule getModule(ReactContext reactContext) {
...
+            mReactContext.getCurrentActivity().setRequestedOrientation(initialRequestedOrientation);
+
+            mReactContext.removeLifecycleEventListener(this);
+          }
+        };
+        webView.setWebChromeClient(mWebChromeClient);
+      }
+      else {
+        Log.w("RNCWebViewManager", "Unable to get current activity thus could not set web chrome client with full screen video.");
+      }
```
TheAlmightyBob, line 796: This means the web chrome client won't be set at all, right? Is that a problem? Should there be some kind of fallback logic, e.g. to use the below branch as if fullscreen video wasn't enabled?

(I'm not sure, just asking questions here)

TheAlmightyBob: Why do you think that play-services-tasks is related to react-native-webview and that this change would fix it?

abhishek-netizen: My bad, its a mistake, play console warning nothing to do with webview

PR #2043, opened 2021-06-08
remove warning logs for common/handled error events.

cristianoccazinsp:
Remove logs for errors that are typically handled by the user (e.g., page load errors). These cause unnecessary noise and are not useful.


cristianoccazinsp:
Could this be merged? Anything wrong with the PR? The spam from these warnings is really annoying.

@Titozzz

Titozzz:
Could we instead log if onError isn't defined ?

cristianoccazinsp:
@Titozzz makes sense. I've done that change now.

PR #2046, opened 2021-06-09
Fix / in react-native.config.js for non windows platforms.

namrog84:
Fix an issue for non windows, such as on mac for ios or android builds.

When doing a `pod install` on mac, the error below occurs.
When building for android, the same error occurs.

This is because the way react-native-cli still reads/parses the react-native.config.js file EVEN if it's not actually going to do anything with it. In other words, it doesn't do a lazy read.

I've let the other appropriate teams know for longer term. But in interim, this will fix it for non windows, and it doesn't break windows since windows is perfectly happy with / or \ or \\\

The underlying error.

```
error ENOENT: no such file or directory, open '/Users/namrog84/MyRepo/src/apps/MyApp/node_modules/react-native-webview/windows/ReactNativeWebView\ReactNativeWebView.vcxproj'. Run CLI with --verbose flag for more details.
[!] Invalid `Podfile` file: 767: unexpected token at 'Error: ENOENT: no such file or directory, open '/Users/namrog84/MyRepo/src/apps/MyApp/node_modules/react-native-webview/windows/ReactNativeWebView\ReactNativeWebView.vcxproj'
```

### DIFF
--- a/react-native.config.js
+++ b/react-native.config.js
@@ -54,11 +54,11 @@ module.exports.dependency = {
       solutionFile: 'ReactNativeWebView.sln',
       projects: [
         {
-          projectFile: 'ReactNativeWebView\\ReactNativeWebView.vcxproj',
+          projectFile: 'ReactNativeWebView/ReactNativeWebView.vcxproj',
           directDependency: true,
         },
         {
-          projectFile: 'WebViewBridgeComponent\\WebViewBridgeComponent.vcxproj',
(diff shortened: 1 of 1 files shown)


juan-alternova:
We need a quick fix, please!

namrog84:
@juan-alternova since these were blocking us badly, I was able to apply this same change/fix via https://www.npmjs.com/package/patch-package until these can get added.   
However I hope to get this PR in as soon approvals!

asklar (approved):

asklar:
thanks @namrog84 !

PR #2052, opened 2021-06-15
Update explanation to open remote debug window

asukiaaa:
It took time for me to open remote debug window on chromium.
How about to change like this with a reference link?


TheAlmightyBob:
The previous doc used to be accurate, but it seems Chrome moved things around a few months ago. Thanks!

TheAlmightyBob (approved):

Titozzz (approved):

Titozzz:
Thanks!

PR #2054, opened 2021-06-16
Remove upper bound on React Native peerdep

### LINKED ISSUE #1935: React native 0.64.0 support
Issue was brought up a couple months ago here: [1886](https://github.com/react-native-webview/react-native-webview/issues/1886)

Wanted to resurface this since this is causing build issues when I attempt to upgrade.

```
error "react-native-webview#react-native@>=0.60 <0.64" doesn't satisfy found match of "react-native@0.64.0"
```

swrobel:
Fixes #1935

I've tested and it works just fine on 0.64.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "type": "Component"
   },
   "peerDependencies": {
-    "react-native": ">=0.60 <=0.64"
+    "react-native": ">=0.60"
   },
   "dependencies": {
     "escape-string-regexp": "2.0.0",


chrisbobbe:
Any `react-native-webview` releases that don't have an upper bound for the React Native peer dep would be claiming that they work as-is with any future version of React Native. I don't think that's realistic.

Titozzz (approved):

Titozzz:
Since react-navigation uses *, I think it's fine to remove the upper bound too

chrisbobbe:
> Since react-navigation uses *, I think it's fine to remove the upper bound too

react-navigation predictably creates bugs like https://github.com/react-navigation/react-navigation/issues/9803, where:

- The peer deps told a user that the version of `react-navigation` they were on would work with RN v0.65-rc.3.
- That version didn't actually work with v0.65-rc.3.
- The user was left to discover on their own that it didn't work.

I understand that many app projects don't really look for a meaningful signal in peer-dep warnings, beyond a vague feeling that something's slightly wrong. But some do. 🙂

So this is a trade-off, not a pure improvement. It's possibly a trade-off worth making; this is just one perspective. 🙂

Issue #1568, opened 2020-08-11
[Bug Android] - WebView inside ScrollView scroll in iOS but no in Android 


mrigo:
**Bug description:**
I've created a WebView inside a ScrollView.
I know is not a great UX choice for usability, but I need it to show privacy text that is veeeeery long with links.
In iOS, I can scroll the ScrollView AND the WebView, interacting with the links inside the WebView.
In Android, I can scroll ONLY the ScrollView, but I can still interact with the links inside the WebView.

**To Reproduce:**
- Install the library
- Create a WebView inside a ScrollView
- Run Android
- Try to scroll the WebView

**Expected behavior:**
I can scroll the content inside the WebView, like in iOS.

**Environment:**
 - OS: Android
 - OS version: every Android version
 - react-native version: 0.60.5
 - react-native-webview version: 10.4.0

lukeramsden:
Getting this exact same issue, RN 0.62 and react-native-webview 10.7.0. Is there any workaround? I tried the auto-height webview method (https://gist.github.com/jkga/4ee1c366d46a3eba9045f88c58d3bc37) but that has some insane rendering issues on my device and I have no idea why.

jsamr:
@lukeramsden Might want to try this: https://formidable-webview.github.io/webshell/docs/autoheight/

aditya1711:
This issue still persist with latest web-view and pan-gesture-handler. Any solutions of workaround?

krunalinfynno:
getting same issue

PR #2056, opened 2021-06-17
feat: add prop to allow nested scroll on android

rlemasquerier:
Currently, on Android, it is not possible to scroll inside a webview when the webview is itself placed inside a scrollview.

This behaviour is documented in this issue : https://github.com/react-native-webview/react-native-webview/issues/1568

On iOS, this behaviour is already handled.

The purpose of this PR is to add an android prop that enable the "nested" scroll inside the webview. 

There is a counterpart though : when the prop is set to `true`, the user has to scroll outside the webview to effectively scroll the scrollview, since the native event is now handled by the webview.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -19,6 +19,7 @@ import android.os.SystemClock;
 import android.text.TextUtils;
 import android.util.Log;
 import android.view.Gravity;
+import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.ViewGroup.LayoutParams;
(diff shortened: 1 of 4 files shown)


TheAlmightyBob:
Could you update the docs as part of this? https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md

(it might be nice to include the "counterpart" that you warned about in the MR description also)

TheAlmightyBob (approved):

mtnptrsn:
Do you know if this is related to [this](https://github.com/react-native-webview/react-native-webview/issues/2057) issue? Would this solve this problem?

rlemasquerier:
@mtnptrsn it seems related.
I'm not sure if it will fix this behaviour, but I would definitely recommend you to give it a try by applying the patch on your example repo to confirm if it does

helloWorld2020102:
Hi,
This fix complies with our requirements. Very thankful for raising the PR.
Any idea when it may get approved?
@rlemasquerier, @jamonholmgren, @Titozzz

cjhines:
This is also a blocking issue for us. Please consider reviewing @Titozzz @jamonholmgren

jgalianoz:
Please release this 🙏🏼

Melinag1992:
issue for me as well , please take a look  @jamonholmgren @Titozzz

Titozzz (approved):

PR #2063, opened 2021-06-24
Support dark mode on Android

### LINKED ISSUE #1677: feat(android): Dark mode
Android doesn't support dark-theme media query

Bridge the interface in Android to set the dark mode
```
if(WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK)) {
    WebSettingsCompat.setForceDark(myWebView.getSettings(), WebSettingsCompat.FORCE_DARK_ON);
}
```

CouldyZ:
On iOS the WebView automatically uses the dark-theme if the system appearance is set to dark. On Android this is not happening.

This PR adds support for dark-theme, and fixes #1677 (this issue was closed because stale).

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -131,9 +131,11 @@ repositories {
 }
 
 def kotlin_version = getExtOrDefault('kotlinVersion')
+def webkit_version = getExtOrDefault('webkitVersion')
 
 dependencies {
   //noinspection GradleDynamicVersion
   implementation 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+  implementation "androidx.webkit:webkit:$webkit_version"
 }
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -1,4 +1,5 @@
 ReactNativeWebView_kotlinVersion=1.3.50
+ReactNativeWebView_webkitVersion=1.4.0
 ReactNativeWebView_compileSdkVersion=29
 ReactNativeWebView_buildToolsVersion=29.0.3
 ReactNativeWebView_targetSdkVersion=28
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -45,6 +45,8 @@ import androidx.annotation.Nullable;
 import androidx.annotation.RequiresApi;
 import androidx.core.content.ContextCompat;
 import androidx.core.util.Pair;
+import androidx.webkit.WebSettingsCompat;
+import androidx.webkit.WebViewFeature;
 
 import com.facebook.common.logging.FLog;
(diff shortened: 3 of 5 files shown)


Andrew-Chen-Wang:
A little off topic, but on an iOS simulator with dark mode toggled on, the web view for an injected/string HTML source does not use the iOS dark mode theme. I suggest a broader option that defaults to true:

We can take advantage of an HTML meta tag: `<meta name="color-scheme" content="dark light">`. Not sure how we'd inject this tag though since I'm not familiar with this library.

<details><summary>Posted this on the wrong thread</summary>

This is the first place I saw "Dark mode" for the web view, so just wanted to share:

For anyone who uses `source: {html: '' }` or is willing to just download the html and use `source: {html: '' }`, then you can add in a string along with your source html:

```html
${darkMode ? '<meta name="color-scheme" content="dark light">' : ''}
```

This will use the browser's automatic dark mode color scheme

Combine that string with the html you're about to inject. The meta tag should be inside a head tag (though IIRC it can also be in a body tag, though it's been awhile since I tried).

There is also a CSS property using `:root { color-scheme: dark light }`. It does the same thing. I believe all browsers only open Safari and Chrome for iOS and Android respectively so I think there should be an option in the WebView component to just add this. We can add a boolean prop that injects this CSS property for you.

Ref: https://web.dev/color-scheme/

</details>

Titozzz:
@Andrew-Chen-Wang I'll merge this as this enables it on android. For your use case either open another PR or inject manually inside the HTML ?

Andrew-Chen-Wang:
@Titozzz not sure where to even put the code; but yea, I've done it myself.

davidli3100:
Would doing the following be enough to enable dark mode on android? I'm not too sure what "calling a static method" refers to in `reference.md`

```
<Webview forceDarkOn />
```

CouldyZ:
> Would doing the following be enough to enable dark mode on android? I'm not too sure what "calling a static method" refers to in `reference.md`
> 
> ```
> <Webview forceDarkOn />
> ```

@davidli3100 
1. Make sure that ```forceDarkOn`` is what you want, please check first:[WebSettingsCompat.DARK_STRATEGY_PREFER_WEB_THEME_OVER_USER_AGENT_DARKENING](https://developer.android.com/reference/androidx/webkit/WebSettingsCompat#DARK_STRATEGY_PREFER_WEB_THEME_OVER_USER_AGENT_DARKENING)
2. usage:
> ```
> import { useColorScheme  } from 'react-native'
> ...
> const colorScheme = useColorScheme()
> <WebView forceDarkOn={colorScheme === 'dark'} />
> ```

VanuPhantom:
Sorry for the necro, but I ran into this issue myself yesterday and couldn't find a clear solution anywhere. I ended up [posting my solution to StackOverflow](https://stackoverflow.com/a/78304676/7470402). It boils down to this;

To fully support dark mode, you need to change the `AppTheme`'s parent from `Light` to `DayNight` in `main/res/values/styles.xml`.

Issue #2078, opened 2021-07-04
WebView crashes app on Android when clicking a link that attempts to download an image.


Dor256:
**Bug description:**
I have a `WebView` that at some point presents a link that attempts to download an image.
On iOS everything works fine and the `WebView` redirects to a page where the image is shown, I am able to intercept this with `onShouldStartLoadWithRequest` and push a custom native screen.

However, on Android, the app simply crashes after tapping the link. Also, `onShouldStartLoadWithRequest` is not called on Android and so I am unable to intercept the link to prevent this crash.

When looking at the logs with `adb logcat` I can see the following error:
`java.lang.IllegalArgumentException: Can only download HTTP/HTTPS URIs: data:image/jpeg;base64,/9j/4A...`

Is there any way to work around this and prevent the crash?

**Expected behavior:**
`onShouldStartLoadWithRequest` should be called when clicking links that attempt to download files.

**Environment:**
 - OS: Android
 - OS version: 11.0
 - react-native version: 0.64.0
 - react-native-webview version: 11.0.2 & latest

Rajan-Savaliya-spt:
android webview pdf downalod item app craze and close the app happed: i want to downalod anyfile from the webview in android device any solution for that....

remberluyckx:
Anyone have a solution for this problem?

Issue #2073, opened 2021-06-30
Downloading from HTML attribute not working in android Webview


BallySharma:
import { WebView } from 'react-native-webview';

<WebView
      originWhitelist={['*']}
      cacheEnabled={false}
      setSupportMultipleWindows={false}
      saveFormDataDisabled={true}
      allowsBackForwardNavigationGestures={true}
      onError={console.error.bind(console, 'error')}
      incognito={true}
      cacheMode={'LOAD_NO_CACHE'}
      javaScriptEnabled={true}
      onLoadEnd={() => this.setState({ isLoading: false })}
      source={{
        html: '<script type="text/javascript"> ' +
          'window.onload=function(){' +
          'document.forms["myForm"].submit();' + '}</script>' +
          '<body >' +
          '<form id="myForm" method="POST" action="' + I_SERVICE_URL + '">' +
          '<input type="hidden" name="isSales" value="' + this.isSales + '"/>' +
          '<input type="hidden" name="jwtToken" value="' + this.accessToken + '"/>' +
          '<input type="hidden" type="submit" value="Login"/>' +
          '</form>' +
          '</body>'
      }}
/>

Webview render HTML content properly but unable to download pdf file after click on download button.

If using external browser then file downloaded successfully without any issues.

How can we download the pdf file using source as html content?

Issue #1945, opened 2021-03-30
Blob url not working in android only


kailashchand-arcgate:
## Issue : 
After upgrade version from **8.0.6 to 11.3.2** but still getting app crash. we have tested our app in android (10 ) and **iOS simulator** (iphone 11 pro, 14.1) both and found our app is crashing in android only. 
## Dependencies 
Our app is based on below dependencies :
- "react": "16.13.1",
- "react-native": "0.63.4",
- "react-native-webview": "^11.3.2",
- Android version : 11.0

## Crash Logs
```sh
Fatal Exception: java.lang.IllegalArgumentException: Can only download HTTP/HTTPS URIs: blob:https://www.emerald.com/e76b0f3e-c7db-4740-8dda-5c724e8b7981
       at android.app.DownloadManager$Request.<init>(DownloadManager.java:497)
       at com.reactnativecommunity.webview.RNCWebViewManager$2.onDownloadStart(RNCWebViewManager.java:199)
       at Ct.handleMessage(chromium-Monochrome.aab-stable-438910523:155)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loop(Looper.java:201)
       at android.app.ActivityThread.main(ActivityThread.java:6810)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:547)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:873)
```

JordaoViktor:
same problem here dude!

JordaoViktor:
i have the same problem too!

JordaoViktor:
i still with that lib, but i am making the workaround which opacity: 0.99 and androidLayerType="hardware" mixedContentMode="always", but that solution is not what i expected but works. I also saw in some issues here that you can use react-native-inappbrowser-reborn if helps in your case.

TheAlmightyBob:
This is technically a duplicate of #1090, but unfortunately that one was automatically closed by the github-actions bot (even though there was still activity on it.... I am not a fan of that bot)

There've been a few PRs attempting to resolve this: #1546 , #1817

PR #2091, opened 2021-07-14
fix(android): add onDownloadUnsupportedURL to handle android crash when download blob or base64 link

twisger:
There are lots of crash issues in android is caused by download a link not http protocol,so add a event to handle it and pass it to js layer.If necceary,it can be resolved further in the js handler.

related issue :#1090,#1760,#1945, #2073,#2078
related PR: #1546,#1817,#2065


KaungHtetHein116:
did anyone find solution yet?

twisger:
> did anyone find solution yet?

I'm working on making the blob and base64 link works well, but there are some file write permission issue need to be resolved in android 10.

rameshpatel79:
@twisger , 

Will this handle the issue mentioned at: https://github.com/react-native-webview/react-native-webview/issues/1914?

Also, you mentioned "I also have to fetch the blob in JS of webview" at https://github.com/react-native-webview/react-native-webview/pull/1817#issuecomment-795795135. Can you please tell how to do that? I want to catch the blob response when user clicks on a button that downloads blob file. Very simple sample here: https://snack.expo.io/@rameshpatell79/react-native-webview-download-error

twisger:
> @twisger ,
> 
> Will this handle the issue mentioned at: #1914?
> 
> Also, you mentioned "I also have to fetch the blob in JS of webview" at [#1817 (comment)](https://github.com/react-native-webview/react-native-webview/pull/1817#issuecomment-795795135). Can you please tell how to do that? I want to catch the blob response when user clicks on a button that downloads blob file. Very simple sample here: https://snack.expo.io/@rameshpatell79/react-native-webview-download-error

@rameshpatel79 It will not be handled because it it not a url cause crash. The code can fetch blob is like blow:




```
const getCodeOfFetchBlob = (url: string) => `
(function(){
    fetch("${url}").then(function(res) {
      var reader = new FileReader()
      res.blob().then(function(blob) {
        reader.readAsDataURL(blob)
        reader.onload = function() {
          window.ReactNativeWebView.postMessage(JSON.stringify({
            event: 'blob',
            payload: {
              uri: this.result
            }
          }))
        }
      })
    })
})();
```
If need keep compatible with lower version webView, the `fetch` should be replaced to `xhr`.

smarano:
I have found and tested a simple working solution to download "**blob:http://**" urls.
I realized that the binary content of the url 'ahead' of the "blob:" prefix has still valid downloadable content. 
The working solution is to simply remove the "blob:" prefix in the onDownloadStart() method.
In addition, for prefixes other than "blob:", just return and avoid the crash.

```
// file RNCWebViewManager.java
 webView.setDownloadListener(new DownloadListener() {
      public void onDownloadStart(String url, String userAgent, String contentDisposition, String mimetype, long contentLength) {

	// check unsupported urls
	if (!url.startsWith("http")) {   
          if (url.startsWith("blob")) {
            Log.d("RNCWebViewManager", "Binary dowload from: " + url);
            url = url.replace("blob:","");
          } else {
            // for other unsupported protocols, just do nothing and at least AVOID CRASH
            Log.w("RNCWebViewManager", "Unsupported URL protocol: " + url);
            return;
          }
        }

        webView.setIgnoreErrFailedForThisURL(url);
       ...
       ... // existing code 
       ...
	   }

karthikiyengar:
@twisger - I was wondering if this is something that is still being worked upon? 

@jamonholmgren, @jasongaare - The other promising PR that might help mitigate the crash seems to be https://github.com/react-native-webview/react-native-webview/pull/1546 - I was wondering what direction we intend to move in?

Happy to help on either front!

avinashsahoo:
**Android:**

I've just fixed the issue, by putting a if condition before DownloadManager.Request. If the Url contains "blob", then it'll simply ignore. At least it will prevent crashing the app on android.

        if(!url.contains("blob"))

**iOS** 

It's anyway handling by onShouldStartLoadWithRequest.

        onShouldStartLoadWithRequest={(req) => { 
          if (req.url.startsWith("blob")) { 
            alert("Files can't be downloaded.");
            return false; 
          } 
            return true; 
          }
        }

twisger:
> @twisger - I was wondering if this is something that is still being worked upon?
> 
> @jamonholmgren, @jasongaare - The other promising PR that might help mitigate the crash seems to be #1546 - I was wondering what direction we intend to move in?
> 
> Happy to help on either front!
> 
> P.S: I have opened #2323 that fixes the crash too.

@karthikiyengar 
I tried to make it can be downloaded successfully in android, it almost worked, But I encountered some difficulty when getting the file extension of the blob link. Feel free to send another pr about it.

hypnocill:
@twisger do you have an example or idea how can I utilize the blob url passed to onDownloadUnsupportedUrl so that I can either open the file or download it? That would be really helpful if you do.
Thank you!

twisger:
> @twisger do you have an example or idea how can I utilize the blob url passed to onDownloadUnsupportedUrl so that I can either open the file or download it? That would be really helpful if you do. Thank you!

This PR is not fully implemented and merged because of some difficulties of getting file extension of blob link file. But the main part is already included in the code of this pr. You can view file change of this pr and implement by yourself if you only want to support specific file type. The main step is:
1.Add a event to pass blob link to js webview.
2.Download this link in js webview and pass the blob data to native layer through event.
3.Download the blob data in native layer

hypnocill:
@twisger I checked the diff, hopefully the injected js on blob url download will work out. I'll patch it locally. Thanks!

hypnocill:
@twisger I patched the lib locally and it works perfectly fine with PDF from the server - it downloads it (can't open it but I don't need to). The other case I have to handle though, is when the server responds with SVG, not PDF. In this case, it doesn't work and the "window._ReactNativeWebView.downloadDataURL(dataURL)" function is never called.
I suspected it was due to "xhr.responseType = 'blob'" working only for PDF but not SVG though I tried it with 'arraybuffer', 'text' and 'document' without success.

twisger:
> @twisger I patched the lib locally and it works perfectly fine with PDF from the server - it downloads it (can't open it but I don't need to). The other case I have to handle though, is when the server responds with SVG, not PDF. In this case, it doesn't work and the "window._ReactNativeWebView.downloadDataURL(dataURL)" function is never called. I suspected it was due to "xhr.responseType = 'blob'" working only for PDF but not SVG though I tried it with 'arraybuffer', 'text' and 'document' without success.
> 
> Do you have any idea how to make this work with SVG? Thank you!
window._ReactNativeWebView.downloadDataURL, try to check if the request is callled sucessfully. It may be refused by server because of some reason, such as the responseType is not correct. Mabe `image/svg+xml` can works.

hypnocill:
@twisger I assume that the request is rejected for some reason, possibly due to response type but I tried all values  from https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType and still no result. Is image/svg+xml  valid response type?
Unfortunately it's very difficult for me to test and debug it because the react-native-webview lib is not directly inside my app but inside other app (Jitsi sdk) that's inside my app and one build of both projects so that I can run my code once takes a lot of time.

Issue #647, opened 2019-06-17
Display custom context menu entries for text selection


poltak:
**Is your feature request related to a problem? If so, Please describe.**

I'd like to be able to run my own code for the text I select on different webpages within the webview via the popup context menu. Currently the possible options there seems to be limited to whatever the default system options are.

**Describe the solutions you came up with**

Solutions would be to either:
- enable adding of new entries to the default context menu
- showing our own context menu made up only of defined entries

**Additional context**

- There was this closed, but unresolved, issue on the `react-native` repo with the same idea: https://github.com/facebook/react-native/issues/19111
- The last post in there links to this package to customize context menu entries. It works great for general text, but does not work with `<WebView>`: https://github.com/Astrocoders/react-native-selectable-text
- I found some stackoverflow answers detailing how to do this in native code. Alas I am not experienced with native mobile development, and they very well may be not be relevant answers anymore going by the dates. Android (https://stackoverflow.com/a/22391169) and iOS (https://stackoverflow.com/a/3596414).

Thanks for your time, and please let me know if you need any more information.

mahdidavoodi7:
@poltak Did you find anything mate ?

poltak:
Sorry, but no @mahdidavoodi7. We are no longer using WebView in our app, so have abandoned looking for a solution to this.

aliazizi:
Is any update for this?

friendRU:
@aliazizi hi, I have found some info. We are going to try to use this approach (there are several tabs inside to understand the decision better) - https://www.one-tab.com/page/G1CUcfXjRsyFJaVXkz2yWQ

dzcpy:
Any updates?

dzcpy:
https://medium.com/flutter-community/inappwebview-the-real-power-of-webviews-in-flutter-c6d52374209d
If you search for "Custom context menus" on this page, there is a way to do similar  things with `https://pub.dev/packages/flutter_inappwebview` in flutter. Maybe there's a way to borrow some code there and make it work for React Native Webview?

PR #2101, opened 2021-07-21
Allow custom menu items for selection menu

sbatson5:
## What's in this PR?

Adds the ability for users to specify custom context menu items on iOS.

Screen Shot 2021-07-21 at 5 10 04 PM

Addresses some older issues
- https://github.com/react-native-webview/react-native-webview/issues/647
- https://github.com/facebook/react-native/issues/3414

This adds two params to the WebView component:
- `menuItems` -- Array of strings, specifying the custom menu items you want to appear
- `onSelection` -- Callback function called when an item is clicked (currently, it only passes the string of the menu item)

```
<WebView 
  source={source} 
  menuItems={['Foo', 'Bar']}
  onSelection={({ nativeEvent }) => console.log(nativeEvent.customMenuKey)} />
```

## How to test

With the example above, can you add custom menu items and seem the logged when you click

### Android?

The project I'm on is iOS only, so I'm not sure about porting this to Android/Java.  Open to any suggestions!

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -23,6 +23,10 @@ static NSString *const MessageHandlerName = @"ReactNativeWebView";
 static NSURLCredential* clientAuthenticationCredential;
 static NSDictionary* customCertificatesForHost;
 
+NSString *const CUSTOM_SELECTOR = @"_CUSTOM_SELECTOR_";
+UITextPosition *selectionStart;
+UITextPosition* beginning;
+
 #if !TARGET_OS_OSX
 // runtime trick to remove WKWebView keyboard default toolbar
 // see: http://stackoverflow.com/questions/19033292/ios-7-uiwebview-keyboard-issue/19042279#19042279
@@ -191,6 +195,48 @@ static NSDictionary* customCertificatesForHost;
 - (void)dealloc
 {
   [[NSNotificationCenter defaultCenter] removeObserver:self];
+}
+
+- (void)tappedMenuItem:(NSString *)eventType
+{
+    self.onSelection(@{
+        @"customMenuKey": eventType
+    });
+}
+
+- (NSMethodSignature *)methodSignatureForSelector:(SEL)sel
+{
+    if ([super methodSignatureForSelector:sel]) {
+        return [super methodSignatureForSelector:sel];
+    }
+    return [super methodSignatureForSelector:@selector(tappedMenuItem:)];
+}
+
+- (void)forwardInvocation:(NSInvocation *)invocation
+{
+    NSString *sel = NSStringFromSelector([invocation selector]);
+    NSRange match = [sel rangeOfString:CUSTOM_SELECTOR];
+    if (match.location == 0) {
+        [self tappedMenuItem:[sel substringFromIndex:17]];
+    } else {
+        [super forwardInvocation:invocation];
+    }
+}
+
+- (BOOL)canBecomeFirstResponder
+{
+    return YES;
+}
+
+- (BOOL)canPerformAction:(SEL)action withSender:(id)sender
+{
+    NSString *sel = NSStringFromSelector(action);
+    NSRange match = [sel rangeOfString:CUSTOM_SELECTOR];
+
+    if (match.location == 0) {
+        return YES;
+    }
+    return NO;
 }
 
 /**
@@ -775,6 +821,21 @@ static NSDictionary* customCertificatesForHost;
 - (void)layoutSubviews
 {
   [super layoutSubviews];
+
+  UIMenuController *menuController = [UIMenuController sharedMenuController];
+  NSMutableArray *menuControllerItems = [NSMutableArray arrayWithCapacity:self.menuItems.count];
+  
+  for(NSString *menuItemName in self.menuItems) {
+      NSString *sel = [NSString stringWithFormat:@"%@%@", CUSTOM_SELECTOR, menuItemName];
+      UIMenuItem *item = [[UIMenuItem alloc] initWithTitle: menuItemName
+                                                    action: NSSelectorFromString(sel)];
+      
+      [menuControllerItems addObject: item];
+  }
+  
+  menuController.menuItems = menuControllerItems;
+  [menuController setTargetRect:self.bounds inView:self];
+  [menuController setMenuVisible:YES animated:YES];
 
   // Ensure webview takes the position and dimensions of RNCWebView
   _webView.frame = self.bounds;
--- a/apple/RNCWebView.h
+++ b/apple/RNCWebView.h
@@ -69,6 +69,8 @@
 @property (nonatomic, copy) NSString * _Nullable allowingReadAccessToURL;
 @property (nonatomic, assign) BOOL pullToRefreshEnabled;
 @property (nonatomic, assign) BOOL enableApplePay;
+@property (nonatomic, copy) NSArray<NSString *> * _Nullable menuItems;
+@property (nonatomic, copy) RCTDirectEventBlock onSelection;
 #if !TARGET_OS_OSX
 @property (nonatomic, weak) UIRefreshControl * _Nullable refreshControl;
 #endif
--- a/apple/RNCWebViewManager.h
+++ b/apple/RNCWebViewManager.h
@@ -8,4 +8,6 @@
 #import <React/RCTViewManager.h>
 
 @interface RNCWebViewManager : RCTViewManager
+  @property (nonatomic, copy) NSArray<NSString *> * _Nullable menuItems;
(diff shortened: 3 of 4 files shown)


apple/RNCWebView.m
```diff
@@ -185,14 +187,93 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+    self.onSelection(@{
+        @"customMenuKey": eventType
+    });
+}
+
+- (NSMethodSignature *)methodSignatureForSelector:(SEL)sel
+{
+    if ([super methodSignatureForSelector:sel]) {
+        return [super methodSignatureForSelector:sel];
+    }
```
TheAlmightyBob, line 246: Could this store `[super methodSignatureForSelector:sel]` in a local variable and return that rather than invoking the method twice?

TheAlmightyBob (changes requested):
Looks like this still needs corresponding Typescript/documentation changes.

Also, it might be good to disable most of this logic if no menu customization has actually been configured. And group/comment the methods as it's not obvious that things like `methodSignatureForSelector` and `canBecomeFirstResponder` are related to this.

sbatson5:
Would anyone be able to offer help on an edge case? The selection works as expected when you use a long press to highlight a word.  It then works fine if you take your finger off and use the dot to change the amount of text selected.  However, if you use a long press to highlight then move your finger to select more in the same gesture, the menu never appears.  If afterwards, you slightly move the dot, it appears as expected.

I'm not sure what event is being triggered to prevent it from firing, but as far as I can tell, the gesture `ended` event is firing, so it _should_ appear.

Anyone have any insights?

TheAlmightyBob:
> Would anyone be able to offer help on an edge case? The selection works as expected when you use a long press to highlight a word. It then works fine if you take your finger off and use the dot to change the amount of text selected. However, if you use a long press to highlight then move your finger to select more in the same gesture, the menu never appears. If afterwards, you slightly move the dot, it appears as expected.
> 
> I'm not sure what event is being triggered to prevent it from firing, but as far as I can tell, the gesture `ended` event is firing, so it _should_ appear.
> 
> Anyone have any insights?

Does tapping on the selection (rather than moving the dot) cause the menu to appear?

The behavior you describe actually seems to match the normal iOS text selection behavior.

sbatson5:
> 
> Does tapping on the selection (rather than moving the dot) cause the menu to appear?
> 
> The behavior you describe actually seems to match the normal iOS text selection behavior.

Yes, it does appear if I just tap the highlight. 🤦 I wasn't aware that was the default behavior, but you're right. It's working as it should

sbatson5:
@TheAlmightyBob Added checks to disable the logic if these props aren't passed, as well as some comments and documentation in the README

sbatson5:
As a note, I have no idea how to get rid of "Copy, Look up, Share..." on the context menu. It doesn't look like those get triggered by the `canPerformAction` of this class.

Titozzz:
I think you forgot to declare the props as optional in typescript (with a ?)

docs/Reference.md
```diff
@@ -1385,6 +1385,31 @@ Example:
...
+Example:
+
+```javascript
+<WebView menuItems={['Tweet', 'Save for later']} />
+
+```
+
+### `onCustomMenuSelection`
+
+Function called when a custom menu item is selected.  It receives a Native event, which includes two custom keys: `customMenuKey` and `selectedText`.
```
TheAlmightyBob, line 1401: Small detail, but with localization it could be nice if the menu `key` and `label` were separate things.

apple/RNCWebView.m
```diff
@@ -185,14 +187,119 @@ - (instancetype)initWithFrame:(CGRect)frame
...
 #endif // !TARGET_OS_OSX
+
+  // Allow this object to recognize gestures
+  UILongPressGestureRecognizer *longPress = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(startLongPress:)];
+  longPress.delegate = self;
+
+  longPress.minimumPressDuration = 0.4f;
+  longPress.numberOfTouchesRequired = 1;
+  longPress.cancelsTouchesInView = YES;
+  [self addGestureRecognizer:longPress];
```
TheAlmightyBob, line 198: The checks you added are good, but I was _kind of hoping_ that even the gesture recognizer could be skipped if the menu item properties weren't set.

I'm not sure how complicated that would be though, or how likely it is that this could actually cause any trouble, it just makes me slightly nervous that this new gesture recognizer would be in place for everybody even if they're not using this new feature...

sbatson5: I'm really not overly familiar with Objective-C or the lifecycle of these components.  `self.menuItems` isn't defined at this point, so I can't do a check on whether or not the object exists.  Is there another lifecycle hook I can use that would have the same effect?

docs/Reference.md
TheAlmightyBob: I meant the array of objects approach. IMO it makes the API clearer... usually, in my experience, IDs and UI text are not the same. The label might be a bunch of different things depending on the language, which with the "array of strings" approach means your code needs to be making sure to perform logic (e.g. switch statements, etc) with the localized strings, and logging could get confusing (might need to write logic to translate back from the display string into a string that can be logged in the developer's language), etc...

phuongwd: Sorry for interrupting :), what do you think we use this array?

```
const menuItems = [{
    title: 'Highlight',
    onPress: () => {}
  },
  {
    title: 'Share',
    onPress: : () => {}
  },
  {
    title: 'Copy',
    onPress: : () => {}
  }
]
```

sbatson5: @phuongwd I couldn't find a way to include a function in within an object and pass it to objective-c.

@TheAlmightyBob Updated this to include to keys -- `key` and `label`.  I think this PR should be ready for review now 🤞

sbatson5:
@TheAlmightyBob Could I bump to get a re-review on this? It should be good to go

docs/Reference.md
```diff
@@ -1425,24 +1425,25 @@ Example:
...
 ### `menuItems`
 
-An array of custom menu items that will be appended to the UIMenu that appears when selecting text (will appear after 'Copy' and 'Share...').  Used in tandem with `onCustomMenuSelection`
+An array of custom menu item objects that will be appended to the UIMenu that appears when selecting text (will appear after 'Copy' and 'Share...').  Used in tandem with `onCustomMenuSelection`
 
 Example:
 
 ```javascript
-<WebView menuItems={['Tweet', 'Save for later']} />
+<WebView menuItems={[{ label: 'Tweet', key: 'tweet', { label: 'Save for later', key: 'saveForLater' }]} />
```
TheAlmightyBob, line 1433: I think you're missing a `}`

TheAlmightyBob (changes requested):
Aside from that little documentation typo this looks reasonable to me now, thanks for figuring out the key/label thing

docs/Reference.md
sbatson5: Good catch. Fixed

TheAlmightyBob (dismissed):

Titozzz (approved):

danilo900:
Hello! It is really good. But we need a option to override the default menu items, like "copy", "search".

sbatson5:
> Hello! It is really good. But we need a option to override the default menu items, like "copy", "search".

I could not figure out how to do this in Objective-C. I believe default options are coming from WKWebView, which this component extends from. To override it would require updating the parent class.

However, we ended up building our own WebView in Swift to accomplish this. I outlined most of the steps in this video:
https://www.youtube.com/watch?v=UsGjTZEaV0w&t=118s

TheAlmightyBob:
@sbatson5 That's a great tutorial!

FYI, the same approach seems to work in Objective-C, and there is already precedent in react-native-webview for overriding the WKWebView (the OSX code does it here: https://github.com/react-native-webview/react-native-webview/blob/master/apple/RNCWebView.m#L52)

So it should be possible to add that capability if somebody wants to take it on.

andreialecu:
Hey @sbatson5 it appears this PR created a conflict with a different lib we're using:

https://github.com/react-native-webview/react-native-webview/blob/a0f1e2641a733777e124e8396ee3b86ec4f3deb8/apple/RNCWebView.m#L26

The `react-native-selectable-text` library has an identical one:

https://github.com/Astrocoders/react-native-selectable-text/blob/master/ios/RNSelectableTextView.m#L28

This results in:
```
duplicate symbol '_CUSTOM_SELECTOR' in:
    .../react-native-selectable-text/libreact-native-selectable-text.a(RNSelectableTextView.o)
    .../react-native-webview/libreact-native-webview.a(RNCWebView.o)
ld: 1 duplicate symbol for architecture arm64
```

congnguyen91:
same issue

PR #2116, opened 2021-07-30
feat(apple): userAgent can be changed at runtime

trcoffman:
Make it so that if you change the userAgent prop, this change is propagated to
the native webview on iOS/macOS. Android already has this functionality.

Sometimes it is useful for WebView apps to be able to change their user agent
dynamically to communicate custom things to the web app they are hosting.
Since we can't add custom headers to every single request made by the webview,
the user agent can be used to send things like feature flags to the web app, so
it can have the appropriate behavior depending on what state the mobile app is
in.

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -311,9 +311,7 @@ static NSDictionary* customCertificatesForHost;
     [_webView addObserver:self forKeyPath:@"estimatedProgress" options:NSKeyValueObservingOptionOld | NSKeyValueObservingOptionNew context:nil];
     _webView.allowsBackForwardNavigationGestures = _allowsBackForwardNavigationGestures;
 
-    if (_userAgent) {
-      _webView.customUserAgent = _userAgent;
-    }
+    _webView.customUserAgent = _userAgent;
 #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000 /* __IPHONE_11_0 */
     if ([_webView.scrollView respondsToSelector:@selector(setContentInsetAdjustmentBehavior:)]) {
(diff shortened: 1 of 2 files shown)


apple/RNCWebView.m
```diff
@@ -311,9 +311,7 @@ - (void)didMoveToWindow
     [_webView addObserver:self forKeyPath:@"estimatedProgress" options:NSKeyValueObservingOptionOld | NSKeyValueObservingOptionNew context:nil];
     _webView.allowsBackForwardNavigationGestures = _allowsBackForwardNavigationGestures;
 
-    if (_userAgent) {
-      _webView.customUserAgent = _userAgent;
-    }
+    _webView.customUserAgent = _userAgent;
```
trcoffman, line 314: I'm not entirely clear on the lifecycle of this class and if didMoveToWindow could theoretically occur more than once. But it's perfectly valid to set customUserAgent to nil.

trcoffman, line 314: Removed unnecessary branch. It's perfectly valid to set customUserAgent to `nil`.

apple/RNCWebView.m
```diff
@@ -701,6 +699,12 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
 }
 #endif // !TARGET_OS_OSX
 
+- (void)setUserAgent:(NSString*)userAgent
```
dvicory, line 702: I wonder why it was only done at view creation time before. I can't find any docs saying it's not valid to update at any point. 💯

dvicory (approved):

apple/RNCWebView.m
dvicory: If you set it to `nil` does it leave the default user-agent in place?

trcoffman: `nil` is the default value of this property. I tested this and setting to `nil` restores the default user agent.

Titozzz (approved):

PR #2148, opened 2021-08-30
chore: migrate to react-native-test-app

tido64:
Migrate test apps to [`react-native-test-app`](https://github.com/microsoft/react-native-test-app).

| Android | iOS | macOS | Windows |
| - | - | - | - |
| Screenshot_1630317438 | Simulator Screen Shot - iPhone 11 - 2021-08-30 at 11 57 19 | image | image |

### Remaining work:

- [x] Fix react-native-macos build - #2313
- [x] Align react-native-macos to 0.62 to match with others - #2314
- [x] Bump react-native* to 0.64 (or whatever version we can bump to without leaving other platforms behind)
- [x] Figure out why `autolink-windows` doesn't find `RNCWebView`


tido64:
This is currently blocked by #2218. I will rebase and continue working on this after it's merged.

Titozzz:
I've already reverted the 0.65 bump in https://github.com/react-native-webview/react-native-webview/commit/f8a1ba2ebc56a52030443b491bde14b1c86ec189, isn't that enough ?

tido64:
> I've already reverted the 0.65 bump in [f8a1ba2](https://github.com/react-native-webview/react-native-webview/commit/f8a1ba2ebc56a52030443b491bde14b1c86ec189), isn't that enough ?

Ah, I had missed this. But I think this is still going to be problematic because react-native dependencies are now out of sync. We have react-native 0.65 pulling in react 17 and cli 6, which again pulls in metro 0.66. react-native-windows is still on 0.62, which won't work with react 17 or metro above 0.60. react-native-macos is even further behind. I think we still need to downgrade. The best we can do is 0.64, since react-native-macos is still lagging behind on releases. What do you think?

tido64:
Regardless, this is going to be a pretty big undertaking 😕

Titozzz:
Well, I do think react-native should almost always be latest in the devDeps 👀... Just to make sure the libs works with the latest versions.. but I'm not sure what's the best path forward. We should check what other libs are doing

tido64:
> Well, I do think react-native should almost always be latest in the devDeps 👀... Just to make sure the libs works with the latest versions.. but I'm not sure what's the best path forward. We should check what other libs are doing

I think you can still test the latest versions, but that should be on CI if possible. You can bump to latest, and disable the incompatible packages. This is how we do it in react-native-test-app: https://github.com/microsoft/react-native-test-app/blob/a18cf3418a87d1b2355113cfee6b676e53ca500e/scripts/set-react-version.js#L41-L48

Otherwise, AsyncStorage is still on an older version because of this: https://github.com/react-native-async-storage/async-storage/blob/0ec517c3b6e0020c4edc5a59519c4e3938707d76/package.json#L88

react-native-menu is still on 0.64: https://github.com/react-native-menu/menu/blob/191a1f73416329887275d6ebcfa8fb0974e2f90a/package.json#L58

tido64:
Updated description with remaining work.

Titozzz:
I've merged the PR to fix macOS, anything else I can do to help ?

tido64:
> I've merged the PR to fix macOS, anything else I can do to help ?

One thing I missed in the list above is re-enabling macOS on CI. Do you know if this has ever existed? Would it be something you can bring back?

I have another branch with the react-native-macos bump, but I'm blocked by an issue where it thinks it's on 0.63 instead of 0.62 and refuses to load the JS. This was last night and I haven't investigated further.

tido64:
With @chiaramooney's help, I was able to get the Windows test app booting again. I'm still trying to figure out the CI deploy failure, but the changes should be ready for review now.

Issue #2142, opened 2021-08-24
remove usages of jcenter()


dlwo159:
WARNING:: Please remove usages of jcenter() Maven repository from your build scripts and migrate your build to other Maven repositories.
This repository is deprecated and it will be shut down in the future.
See http://developer.android.com/r/tools/jcenter-end-of-service for more information.
Currently detected usages in: project ':react-native-webview'

Bibazavr:
@Titozzz , this is important - jcenter shut down soon

![image](https://user-images.githubusercontent.com/51177449/134893379-0d2ab3a4-946e-4b77-8065-4d486a932cf5.png)

Titozzz:
Please create a PR and @ me there

Bibazavr:
i guess this issue may closed after new release?

PR #2178, opened 2021-09-27
fix(jcenter): remove jcenter

Bibazavr:
Builds will no longer be able to resolve artifacts from JCenter after February 1st, 2022 because jcenter servers shut down

so, i remove jcenter and update react-native (since 0.65.1 they don't use jcenter)


Bibazavr:
@Titozzz, PR is ready

#2142

Titozzz (approved):

Issue #1979, opened 2021-05-06
Also support cacheMode on iOS


wayne-tt:
Thank you for making open source.

In Android, it was confirmed that the cached data was changed immediately following the http cache policy.

But there seems to be no logic followed by the http cache policy on iOS.

If you are okay, apply for iOS as well?

wayne-tt:
Before this feature was added, I am going to use cacheEnabled props that are false.

bfelbo:
See https://github.com/react-native-webview/react-native-webview/issues/1866#issuecomment-782562104. Adding `incognito={true}` works.

PR #2181, opened 2021-09-30
Update cacheEnabled doc

ibash:
`cacheEnabled` doesn't actually work on iOS, the closest one can get is setting `incognito={true}`.

Ref: https://github.com/react-native-webview/react-native-webview/issues/1979


gangli66:
> `cacheEnabled` doesn't actually work on iOS, the closest one can get is setting `incognito={true}`.
> 
> Ref: #1979

Only on iOS, incognito={true} can delete all web content including cache, local storage and session storage.
On Android, it can only delete cache and the local storage is been remained.
How can I delete cache only on iOS?

PR #2189, opened 2021-10-06
When we get a request for a pop-up, open in a pop-up

flaviut:
Previously, this code meant that iOS popups got opened up in the
webview. This doesn't match the Android behavior, where pop-ups are
opened in the browser, or the intent of the programer: if they didn't
want a pop-up, they wouldn't have used window.open()


TheAlmightyBob:
This seems like a reasonable option to have, but is a breaking change that might not work for everybody. I'd recommend making it an opt-in mode.

flaviut:
I disagree that it should be opt-in, since in my opinion the current behavior is buggy.

However, I would be happy to update this pull request however the maintainers of the project would like me to.

TheAlmightyBob:
It occurs to me that this could be tied into the existing `setSupportMultipleWindows` property (which is currently Android-only).

That goes against my previous recommendation to make it opt-in (since that property defaults to `true`, which was itself a previous breaking change due to security concerns), but as long as this is communicated as a breaking change (with major version bump), the benefit of making that property apply to both iOS and Android seems worth it.

flaviut:
This is still a real inconsistency between the Android & iOS versions of this component.

@Titozzz would you mind taking a look? I'd be happy to make changes & rebase onto latest master.

flaviut:
.

flaviut:
@Titozzz @jamonholmgren

flaviut:
Bump

flaviut:
.

flaviut:
Still around

flaviut:
.

o-alexandrov:
@Titozzz Could you please share your view on this PR?
I think https://github.com/react-native-webview/react-native-webview/pull/2189#issuecomment-970718631 above summarizes it well:
- this PR would resolve the inconsistency between Android and iOS for `setSupportMultipleWindows`

flaviut:
.

Crash--:
@flaviut Since the #2640 has been merged, does this PR is still needed? Can't you use the new onOpenWindow() to achieve what you want?

PR #2236, opened 2021-11-14
Update RNCWebViewManager.java

eladgel:
Changed from protected to public so the implementation can be overridden


ravindarconcent (approved):

TheAlmightyBob:
What does this allow you to override that you couldn't previously override? We already extend and override `RNCWebView` as per the [docs](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Custom-Android.md)

eladgel:
@TheAlmightyBob you are so right!
I didn't see this
Closing the pr

Issue #2177, opened 2021-09-27
Disabling Magnifier on OS version 15+ in iOS


amanBhawsar:
**Bug description:**
Magnifier is appearing on OS version 15+ on iOS on doing a long press on Webviews. Any way to disable it ?

**To Reproduce:**
Long press within any webview on iOS, OS version 15 (latest) -> Magnifier appears

**Expected behavior:**
Need a way to control whether Magnifier should be shown or not. As for most of the use cases, it is not required.

**Screenshots/Videos:**
https://user-images.githubusercontent.com/40451824/134856219-15031589-4b01-444a-93ac-28885aa096f9.mov

**Environment:**
 - OS: iOS
 - OS version: 15+
 - react-native version: 0.63.3
 - react-native-webview version: 8.1.2

cristianoccazinsp:
Wondering the same, extremely annoying.

cristianoccazinsp:
Keep open please.

98gzi:
Any updates on this?

Nebula-Developer:
Any updates two years later?

PR #2252, opened 2021-11-24
Added textInteractionEnabled prop

Arjan-Zuidema:
Added prop `textInteractionEnabled`. Only supported on iOS 15+

When set to `false` disables all text interaction.

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -329,6 +329,10 @@ NSString *const CUSTOM_SELECTOR = @"_CUSTOM_SELECTOR_";
     [prefs setValue:@TRUE forKey:@"javaScriptCanOpenWindowsAutomatically"];
     _prefsUsed = YES;
   }
+  if (!_textInteractionEnabled) {
+    [prefs setValue:@FALSE forKey:@"textInteractionEnabled"];
+    _prefsUsed = YES;
+  }
   if (_prefsUsed) {
     wkWebViewConfig.preferences = prefs;
   }
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -333,10 +333,14 @@ NSString *const CUSTOM_SELECTOR = @"_CUSTOM_SELECTOR_";
     [prefs setValue:@TRUE forKey:@"javaScriptCanOpenWindowsAutomatically"];
     _prefsUsed = YES;
   }
-  if (!_textInteractionEnabled) {
-    [prefs setValue:@FALSE forKey:@"textInteractionEnabled"];
-    _prefsUsed = YES;
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 140500 /* iOS 14.5 */
+  if (@available(iOS 14.5, *)) {
+    if (!_textInteractionEnabled) {
+      [prefs setValue:@FALSE forKey:@"textInteractionEnabled"];
+      _prefsUsed = YES;
+    }
   }
+#endif
   if (_prefsUsed) {
     wkWebViewConfig.preferences = prefs;
   }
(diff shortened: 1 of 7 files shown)


clemensmol:
Merging would be awesome, this is also a nice workaround for #2177.

amanBhawsar (approved):

apple/RNCWebView.m
```diff
@@ -333,6 +333,10 @@ - (WKWebViewConfiguration *)setUpWkWebViewConfig
     [prefs setValue:@TRUE forKey:@"javaScriptCanOpenWindowsAutomatically"];
     _prefsUsed = YES;
   }
+  if (!_textInteractionEnabled) {
+    [prefs setValue:@FALSE forKey:@"textInteractionEnabled"];
+    _prefsUsed = YES;
+  }
```
TheAlmightyBob, line 339: I think this also needs the `#if` check around it?

And possibly also a runtime `if (@available` check?

TheAlmightyBob (changes requested):
Please update `WebViewTypes` also.

And according to the [docs](https://developer.apple.com/documentation/webkit/wkpreferences/3727362-textinteractionenabled?changes=la_1&language=objc), it looks like this was actually introduced with iOS14.5, not iOS15?

apple/RNCWebView.m
Arjan-Zuidema: Fixed this with both a #if and @availabel check for iOS 14.5

Arjan-Zuidema:
> Please update `WebViewTypes` also.
> 
> And according to the [docs](https://developer.apple.com/documentation/webkit/wkpreferences/3727362-textinteractionenabled?changes=la_1&language=objc), it looks like this was actually introduced with iOS14.5, not iOS15?

Updated the `WebViewTypes`

apple/RNCWebViewManager.m
```diff
@@ -104,6 +104,7 @@ - (RCTUIView *)view
 #endif
 
 #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 150000 /* iOS 15 */
+RCT_EXPORT_VIEW_PROPERTY(textInteractionEnabled, BOOL)
```
TheAlmightyBob, line 107: This is still gating it under iOS 15...?

docs/Reference.md
```diff
@@ -1377,6 +1378,24 @@ Example:
 
 ---
 
+### `textInteractionEnabled`[⬆](#props-index)<!-- Link generated with jump2header -->
+
+If false indicates to WebKit that a WKWebView will not interact with text, thus not showing a text selection loop. Only applicable for iOS 15 or greater.
```
TheAlmightyBob, line 1383: Possibly should update this to say 14.5

src/WebViewTypes.ts
```diff
@@ -672,6 +673,15 @@ export interface IOSWebViewProps extends WebViewSharedProps {
    */
   limitsNavigationsToAppBoundDomains?: boolean;
 
+  /**
+   * If false indicates to WebKit that a WKWebView will not interact with text, thus
+   * not showing a text selection loop. Only applicable for iOS 15 or greater.
```
TheAlmightyBob, line 678: Should probably update to say 14.5

TheAlmightyBob (changes requested):
Looking pretty close!

It would also be good to test on a 14.5 simulator and make sure it really does still work as you expect there, if you haven't already

src/WebViewTypes.ts
Arjan-Zuidema: Updated to version 14.5

docs/Reference.md
Arjan-Zuidema: Updated to version 14.5

apple/RNCWebViewManager.m
Arjan-Zuidema: Created an iOS 14.5 section for this

Arjan-Zuidema:
> Looking pretty close!
> 
> It would also be good to test on a 14.5 simulator and make sure it really does still work as you expect there, if you haven't already

Fixed the problems you pointed out, also tested on a simulator running iOS 14.5 this time :)

TheAlmightyBob (approved):

TheAlmightyBob:
Thanks @Arjan-Zuidema!

Issue #2256, opened 2021-11-28
Avoid repetitive permission prompts on iOS


fbruckhoff:
Problem: By default, iOS quickly forgets about permissions granted to requests originating from `WKWebView`. Every time the user relaunches the app and performs an action within WKWebView such as turning on their microphone or camera, `WKWebView` asks for their permission. What's more, when `WKWebView` initiates a permission request, iOS will prompt twice: Once for the native layer, and once for the website loaded within `WKWebView`. 

We wanted our WebRTC-based audio & video communications app to be kind to users and remember their preferences instead of bothering them each time with repetitive prompts. Also, we wanted to get rid of the double-prompt issue. The second prompt related to the loaded website was especially irritating because it displayed long URLs in the description, as opposed to the reason strings specified in `info.plist`.

**The solution:** Implement WKUIDelegate's new method: `webView:requestMediaCapturePermissionForOrigin:initiatedByFrame:type:decisionHandler:`

This allows iOS to a) remember the permission decisions and b) get rid of the initial double-prompt by inheriting the grant from the native layer.

A PR follows immediately after this with the implemented solution.

Cheers 🥂

efeakaroz13:
Related docs:https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#mediaCapturePermissionGrantType

saeid-elio:
I wonder why this was not implemented in a generic way for all permission dialogs and only for media capture. Other permission  dialogs such as permission to use user's location still suffer from the same problem.

PR #2257, opened 2021-11-28
Enhanced permissions handling to prevent repetitive prompts #2256

fbruckhoff:
Referencing issue #2256 

**Problem:** By default, iOS quickly forgets about permissions granted to requests originating from `WKWebView`. Every time the user relaunches the app and performs an action within WKWebView such as turning on their microphone or camera, `WKWebView` asks for their permission. What's more, when `WKWebView` initiates a permission request, iOS will prompt twice: Once for the native layer, and once for the website loaded within `WKWebView`. 

We wanted our WebRTC-based audio & video communications app to be kind to users and remember their preferences instead of bothering them each time with repetitive prompts. Also, we wanted to get rid of the double-prompt issue. The second prompt related to the loaded website was especially irritating because it displayed long URLs in the description, as opposed to the reason strings specified in `info.plist`.

**The solution:** Implement WKUIDelegate's new method: `webView:requestMediaCapturePermissionForOrigin:initiatedByFrame:type:decisionHandler:`

This allows iOS to a) remember the permission decisions and b) get rid of the initial double-prompt by inheriting the grant from the native layer.

A PR follows immediately after this with the implemented solution.

Cheers 🥂

### DIFF
--- a/apple/RNCWebView.h
+++ b/apple/RNCWebView.h
@@ -9,6 +9,14 @@
 #import <React/RCTDefines.h>
 #import <WebKit/WebKit.h>
 
+typedef enum RNCWebViewPermissionGrantType : NSUInteger {
+    RNCWebViewPermissionGrantType_GrantIfSameHost_ElsePrompt,
+    RNCWebViewPermissionGrantType_GrantIfSameHost_ElseDeny,
+    RNCWebViewPermissionGrantType_Deny,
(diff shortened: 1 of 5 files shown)


saeid-elio:
Problem still exists if in Webview page tries to get permission to use user's current location. Then permission prompt shows twice. Any solution for that?

PR #2258, opened 2021-11-28
Enable webview to open new window on IOS

hlta:


TheAlmightyBob:
What's the difference between this and #2189?

hlta:
> What's the difference between this and #2189?

This is for open a new webview when target="blank" same behaviour like Android which has already been supported  but #2189 is open in external browser

nantianyishao:
great!!!!!!

TheAlmightyBob:
Looks like this PR now contains a bunch of unrelated Android changes, I'm guessing because you created it off of your master branch instead of a dedicated feature branch.

hlta:
> Looks like this PR now contains a bunch of unrelated Android changes, I'm guessing because you created it off of your master branch instead of a dedicated feature branch.

Created the branch just for this fix
https://github.com/react-native-webview/react-native-webview/pull/2346

PR #2279, opened 2021-12-07
feat(android): add minimum font size

theofournier:
# Summary
Android WebView enforces a default minimum font size of 8 (see https://developer.android.com/reference/android/webkit/WebSettings#setMinimumFontSize(int)), causing scaling issues when using smaller font sizes. Content that may be deliberately scaled down to fit 100% width of the device may overflow. This property allows for customizing the minimum font size so scaling appears similar as on iOS.

In cases where grids of content/text are displayed (such as a seating chart), font size may need to be scaled down to fit the entire grid onto one screen width.

Previous Pull Request : https://github.com/react-native-webview/react-native-webview/pull/986

Running on Android.

Reference HTML code that uses font sizes < 8. 

For example:
`<WebView source={{ html: '<html><body><span style="font-size: 4px;">Font size 4</span></body></html>' }} minimumFontSize={1} />`

## Compatibility

OS | Implemented
-- | --
iOS | ❌
Android | ✅

## Checklist
- [x] I added the documentation in Reference.md

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -646,6 +646,11 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
     }
   }
 
+  @ReactProp(name = "minimumFontSize")
+  public void setMinimumFontSize(WebView view, int fontSize) {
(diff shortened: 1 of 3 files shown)


jtheoof:
@Titozzz This looks good. Any chance to get this approved and merged?

TheAlmightyBob (approved):

PR #2304, opened 2021-12-28
Add WebView2 Control Support to Windows

### LINKED ISSUE #1848: Migrate to WebView2 in Windows

EdgeHTML (the "old edge" rendering engine, used in this library) has many quirks and unsupported features, and will stop being supported by Microsoft on March 9, 2021 (https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge). This library renders using WebView, which uses EdgeHTML as a rendering engine. As end of life happens for this rendering engine, sites rendered using this webview may not function.

I believe this can be done by using https://docs.microsoft.com/en-us/microsoft-edge/webview2/, which uses the Edge chromium rendering engine.

**Additional context**

chiaramooney:
## Adding WebView2 to WebView community module

**_What is the goal of this feature work?_**
Give developers access to both the WinUI WebView and WebView2 controls via the WebView community module. Solution will allow developers to use both WebView and WebView2 within the same application, switch an instance of the WebView component between the WinUI WebView and WebView2 controls at runtime, and requires minimal JSX changes to the original WebView API.
With these changes, we will continue to provide support for original WebView control in this community module, due to deploy limitations with WebView2. WebView2 can only target Desktop. WebView2 also currently does not have an API for projecting a WinRT object, but this feature work is in progress.

_**How will this feature work be executed?**_
**Toggling Between the WebView and WebView2 Contro**l
The toggle from WebView to WebView2 will be handled from the JavaScript. In this solution, a Windows specific prop would be added to the WebView component called useWebView2. Next, a separate native module for WebView2 based off of the original WebView code is created. Now we have two windows projects. Then both modules are imported to the JavaScript and, based on the value of the useWebView2 prop, which defaults to false if not specified, then the ReactNativeWebView or ReactNativeWebView2 module is returned as the WebView component. With this solution, developers can toggle between WebView and WebView2 at runtime by adjusting the value of the useWebView2 prop. Note this solution matches the design of the react-native-datetimepicker community module which provides both the TimePicker and DateTimePicker components.

**The ReactNativeWebView2 Native Module**
We have exposed the WebView2 control through WinUI 2.7-prerelease. Currently, WebView (like most native modules) gets its copy of WinUI through the React Native Windows app that it's added to. And the app gets its WinUI version from RNW (unless the WinUI override is used). We need to configure the WebView module similar to react-native-xaml where the module can declare a dependency on a specific WinUI version, so that the module can have access to the WebView2 control.
The source for the ReactNativeWebView2 native module will be based off of the source for the original ReactNativeWebView native module. There will need to be a few adjustments made to the code such as swapping out the WebView control type for WebView2, renaming files and classes, and adjusting functionality to account for behavior that is not supported for the WebView2 control.

**Diagrams**
_Original Module Structure_
image

_New Module Structure_
image

NOTES:
- Change restores previous upgrade of module to use RN and RNW v0.65.
- Change includes upgrades to Windows example app needed for bump to 0.65.

- [x] Update ReactNativeWebView.sln file to expect RNW 0.65.

Resolves #1848


docs/Getting-Started.md
```diff
@@ -86,7 +85,20 @@ Add `PackageProviders().Append(winrt::ReactNativeWebView::ReactPackageProvider()
...
 
-## 3. Import the webview into your component
+## 3. Decide if you want access to WebView2
+The WebView2 control is a WinUI control that renders web content using the Microsoft Edge (Chromium) rendering engine. We have added support for the WebView2 control to the WebView community module. If you want access to the WebView2 control in your app, follow these steps:
+
+  i. Ensure your React Native Windows version is v0.65 or higher.
+
+  ii. Let autolinking handle adding the `ReactNativeWebView2` project to your app.
+
+  iii. Customize your app's WinUI 2.x version to version 2.8.0-prerelease.210927001. See here for instructions.
```
asklar, line 95: or higher; let's also file an issue for updating this when webview2 goes stable in winui2

docs/Getting-Started.md
```diff
@@ -86,7 +85,20 @@ Add `PackageProviders().Append(winrt::ReactNativeWebView::ReactPackageProvider()
...
+  
+  iv. Customize your app's WebView2 version to version 1.0.1020.30. See here for instructions.
```
asklar, line 97: can we do something to remove or automate this?

docs/Reference.md
```diff
@@ -1504,6 +1505,19 @@ An object that specifies the credentials of a user to be used for basic authenti
...
+Use WinUI WebView2 control instead of WebView control as the native webview. The WebView2 control is a WinUI control that renders web content using the Microsoft Edge (Chromium) rendering engine.
+
+| Type    | Required | Platform |
+| ------- | -------- | -------- |
+| boolean | No       | Windows  |
+
+Example:
+
+```javascript
+<WebView useWebView2={true} />
```
asklar, line 1519: is it expected that this value can be toggled at runtime? e.g.

```jsx
const [wv2, setwv2] = useState(false);
// ...
/// something happens and calls setwv2(true)
// ...
<WebView useWebView2={wv2} />
```

example/windows/example/example.vcxproj
```diff
@@ -1,31 +1,30 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(SolutionDir)packages\Microsoft.Windows.CppWinRT.2.0.200729.8\build\native\Microsoft.Windows.CppWinRT.props" Condition="Exists('$(SolutionDir)packages\Microsoft.Windows.CppWinRT.2.0.200729.8\build\native\Microsoft.Windows.CppWinRT.props')" />
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="16.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(SolutionDir)\ExperimentalFeatures.props" Condition="Exists('$(SolutionDir)\ExperimentalFeatures.props')" />
+  <Import Project="..\packages\Microsoft.Windows.CppWinRT.2.0.210312.4\build\native\Microsoft.Windows.CppWinRT.props" Condition="Exists('..\packages\Microsoft.Windows.CppWinRT.2.0.210312.4\build\native\Microsoft.Windows.CppWinRT.props')" />
```
asklar, line 4: should probably use SolutionDir here

example/windows/example/example.vcxproj
```diff
@@ -142,41 +147,56 @@
...
+    <Error Condition="!Exists('..\packages\Microsoft.Windows.CppWinRT.2.0.210312.4\build\native\Microsoft.Windows.CppWinRT.targets')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\Microsoft.Windows.CppWinRT.2.0.210312.4\build\native\Microsoft.Windows.CppWinRT.targets'))" />
+    <Error Condition="!Exists('..\packages\ReactNative.Hermes.Windows.0.8.0-ms.0\build\native\ReactNative.Hermes.Windows.targets')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\ReactNative.Hermes.Windows.0.8.0-ms.0\build\native\ReactNative.Hermes.Windows.targets'))" />
+    <Error Condition="!Exists('..\packages\$(WinUIPackageName).$(WinUIPackageVersion)\build\native\$(WinUIPackageName).targets')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\$(WinUIPackageName).$(WinUIPackageVersion)\build\native\$(WinUIPackageName).targets'))" />
+    <Error Condition="!Exists('..\packages\Microsoft.Web.WebView2.1.0.1020.30\build\native\Microsoft.Web.WebView2.targets')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\Microsoft.Web.WebView2.1.0.1020.30\build\native\Microsoft.Web.WebView2.targets'))" />
```
asklar, line 200: SolutionDir?

package.json
```diff
@@ -67,9 +67,9 @@
     "metro": "0.56.4",
     "metro-react-native-babel-preset": "^0.59.0",
     "react": "16.11.0",
-    "react-native": "0.62.2",
+    "react-native": "^0.65.0",
```
asklar, line 70: this feels wrong to require 0.65+ for everything?

windows/ReactNativeWebView.sln
```diff
@@ -11,194 +11,207 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Chakra", "..\node_modules\r
...
+		{729D9AF8-CD9E-4427-9F6C-FB757E287729}.Debug|Any CPU.ActiveCfg = Debug|Win32
+		{729D9AF8-CD9E-4427-9F6C-FB757E287729}.Debug|ARM.ActiveCfg = Debug|Win32
 		{729D9AF8-CD9E-4427-9F6C-FB757E287729}.Debug|ARM64.ActiveCfg = Debug|Win32
 		{729D9AF8-CD9E-4427-9F6C-FB757E287729}.Debug|x64.ActiveCfg = Debug|x64
 		{729D9AF8-CD9E-4427-9F6C-FB757E287729}.Debug|x64.Build.0 = Debug|x64
 		{729D9AF8-CD9E-4427-9F6C-FB757E287729}.Debug|x86.ActiveCfg = Debug|Win32
 		{729D9AF8-CD9E-4427-9F6C-FB757E287729}.Debug|x86.Build.0 = Debug|Win32
-		{729D9AF8-CD9E-4427-9F6C-FB757E287729}.Release|ARM.ActiveCfg = Release|ARM
-		{729D9AF8-CD9E-4427-9F6C-FB757E287729}.Release|ARM.Build.0 = Release|ARM
+		{729D9AF8-CD9E-4427-9F6C-FB757E287729}.Release|Any CPU.ActiveCfg = Release|Win32
```
asklar, line 63: why do we have anycpu here?

windows/ReactNativeWebView2/PropertySheet.props
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ImportGroup Label="PropertySheets" />
+  <PropertyGroup Label="UserMacros" />
+  <!--
+    To customize common C++/WinRT project properties:
```
asklar, line 6: why do we need 2 solutions/projects? I thought the idea was a single project would target both? if not, does the app always us the reactnativewebview2 project?

windows/ReactNativeWebView2/ReactWebView2.cpp
```diff
@@ -0,0 +1,126 @@
...
+    using namespace Windows::UI::Xaml::Controls;
+    using namespace Windows::UI::Xaml::Input;
+    using namespace Windows::UI::Xaml::Media;
+    using namespace Microsoft::Web::WebView2::Core;
+} // namespace winrt
+
+namespace winrt::ReactNativeWebView2::implementation {
+
+    ReactWebView2::ReactWebView2(winrt::IReactContext const& reactContext) : m_reactContext(reactContext) {
+#ifdef CHAKRACORE_UWP
```
asklar, line 28: what's this for?

windows/ReactNativeWebView2/ReactWebView2.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    void ReactWebView2::WriteWebViewNavigationEventArg(winrt::WebView2 const& sender, winrt::IJSValueWriter const& eventDataWriter) {
+        auto tag = this->GetValue(winrt::FrameworkElement::TagProperty()).as<winrt::IPropertyValue>().GetInt64();
+        WriteProperty(eventDataWriter, L"canGoBack", sender.CanGoBack());
+        WriteProperty(eventDataWriter, L"canGoForward", sender.CanGoForward());
+        if (Is17763OrHigher()) {
+          WriteProperty(eventDataWriter, L"loading", !sender.IsLoaded());
+        }
+        WriteProperty(eventDataWriter, L"target", tag);
+        //WriteProperty(eventDataWriter, L"title", sender.DocumentTitle());
```
asklar, line 79: remove?

windows/ReactNativeWebView2/ReactWebView2.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include "ReactWebView2.g.cpp"
+#include <winrt/Windows.Foundation.Metadata.h>
+#include <optional>
+
+namespace winrt {
+    using namespace Microsoft::ReactNative;
+    using namespace Windows::Data::Json;
+    using namespace Windows::Foundation;
+    using namespace Windows::UI;
+    using namespace Windows::UI::Popups;
```
asklar, line 16: some of these usings and corresponding includes don't seem to be needed, remove where you can

windows/ReactNativeWebView2/ReactWebView2Manager.cpp
```diff
@@ -0,0 +1,145 @@
...
+                        continue;
+                    }
+
+                    bool isPackagerAsset = false;
+                    if (srcMap.find("__packager_asset") != srcMap.end()) {
+                        isPackagerAsset = srcMap.at("__packager_asset").AsBoolean();
+                    }
+                    if (isPackagerAsset && uriString.find("file://") == 0) {
+                        auto bundleRootPath = winrt::to_string(ReactNativeHost().InstanceSettings().BundleRootPath());
+                        uriString.replace(0, 7, bundleRootPath.empty() ? "ms-appx-web:///Bundle/" : bundleRootPath);
```
asklar, line 77: consider replacing the 7 with std::size("file://"), maybe have that string defined somewhere to avoid repeating it (you can make it constexpr)

docs/Getting-Started.md
chiaramooney: #2347 Filed!

docs/Reference.md
chiaramooney: Yes toggle at runtime and works with fast refresh! I'll add to the option description to make that clearer.

docs/Getting-Started.md
chiaramooney: Hm I'm not sure. Maybe for now I can file an issue for it and we can come back to it to improve dev experience later?

package.json
chiaramooney: This was the change that was already merged in and we asked to revert it back to 0.62 until we could fix the example app for 0.65. So it's just restoring the upgrade that was already done and approved!

windows/ReactNativeWebView.sln
chiaramooney: Not sure why that was added to the solution. I'll remove.

windows/ReactNativeWebView2/ReactWebView2.cpp
chiaramooney: That line was taken from the original ReactNativeWebView source. Unclear what its use case is. I can remove if you don't think it's needed.

windows/ReactNativeWebView2/ReactWebView2.cpp
chiaramooney: These were taken from the original ReactWebView source, but I can investigate.

windows/ReactNativeWebView2/PropertySheet.props
chiaramooney: Currently we chose this design: 
"Toggling Between the WebView and WebView2 Control
The toggle from WebView to WebView2 will be handled from the JavaScript. In this solution, a Windows specific prop would be added to the WebView component called useWebView2. **Next, a separate native module for WebView2 based off of the original WebView code is created. Now we have two windows projects. Then both modules are imported to the JavaScript and, based on the value of the useWebView2 prop, which defaults to false if not specified, then the ReactNativeWebView or ReactNativeWebView2 module is returned as the WebView component**. With this solution, developers can toggle between WebView and WebView2 at runtime by adjusting the value of the useWebView2 prop. Note this solution matches the design of the react-native-datetimepicker community module which provides both the TimePicker and DateTimePicker components.

The app will use the reactnativewebview project when the app is using `<WebView />` and will use the reactnativewebview2 project when using `<WebView useWebView2={true} />`
image


I should be able to take the two modules and house them in once vcxproj if we'd prefer to only deal with one project file. We could also save this work as feature work for the future if the two project structure gives people issues.

windows/ReactNativeWebView2/pch.h
```diff
@@ -0,0 +1,19 @@
...
+#include <winrt/Windows.UI.Xaml.Controls.h>
+#include <winrt/Windows.UI.Xaml.Markup.h>
+#include <winrt/Windows.UI.Xaml.Navigation.h>
+#include <winrt/Windows.Web.Http.h>
+#include <winrt/Windows.Web.Http.Headers.h>
+#include <winrt/Microsoft.ReactNative.h>
+#include <winrt/Microsoft.UI.Xaml.Controls.h>
+#if __has_include(<VersionMacros.h>)
+#include <VersionMacros.h>
+#endif
```
asklar, line 19: consider adding something like:
```suggestion
#else
#define RNW_VERSION_AT_LEAST(x,y,z) false
```


So that you can simplify the usage site and always directly call `RNW_VERSION_AT_LEAST`  and not check for the existence of the macro

windows/ReactNativeWebView2/ReactWebView2.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+namespace winrt {
+    using namespace Microsoft::ReactNative;
+    using namespace Windows::Foundation;
+    using namespace Windows::UI;
+    using namespace Windows::UI::Xaml;
+    using namespace Windows::UI::Xaml::Controls;
+    using namespace Microsoft::UI::Xaml::Controls;
+#if defined(RNW_VERSION_AT_LEAST)
+#if RNW_VERSION_AT_LEAST(0,68,0))
```
asklar, line 19: this and other usages of this macro have an extra trailing `)`, so this code should not compile

asklar (changes requested):

docs/Reference.md
```diff
@@ -1526,6 +1527,19 @@ An object that specifies the credentials of a user to be used for basic authenti
...
+Use WinUI WebView2 control instead of WebView control as the native webview. The WebView2 control is a WinUI control that renders web content using the Microsoft Edge (Chromium) rendering engine. Option can be toggled at runtime and supports Fast Refresh.
+
+| Type    | Required | Platform |
+| ------- | -------- | -------- |
+| boolean | No       | Windows  |
+
+Example:
+
+```javascript
+<WebView useWebView2 />
```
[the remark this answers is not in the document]
chiaramooney: It doesn't need to be! When I run locally, prettier actually prefers to leave the {true} off.

jest-setups/jest.setup.js
```diff
@@ -4,7 +4,7 @@ const { platform } = require('./jest.setup.windows');
 
 switch (platform) {
     case "windows":
-        const webViewWindowsAppId = 'WebViewWindows_3x6rhkkr9xcf6!App';
+        const webViewWindowsAppId = '40411fc5-8e92-4d46-b68d-b62df44b1366_3x6rhkkr9xcf6!App';
```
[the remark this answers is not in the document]
chiaramooney: Will update the manifest

src/WebView.windows.tsx
```diff
@@ -33,6 +33,8 @@ import {
   State,
 } from './WebViewTypes';
 
+const ReactNativeVersion = require('../node_modules/react-native/Libraries/Core/ReactNativeVersion');
```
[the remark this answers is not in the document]
chiaramooney: Yes good callout I'll adjust!

windows/ReactNativeWebView2/ReactWebView2.cpp
```diff
@@ -0,0 +1,144 @@
...
+#endif
+} // namespace winrt
+
+namespace winrt::ReactNativeWebView2::implementation {
+
+    ReactWebView2::ReactWebView2(winrt::IReactContext const& reactContext) : m_reactContext(reactContext) {
+#if defined(RNW_VERSION_AT_LEAST)
+#if RNW_VERSION_AT_LEAST(0,68,0))
+#ifdef CHAKRACORE_UWP
+            m_webView = winrt::WebView2(winrt::WebViewExecutionMode::SeparateProcess);
```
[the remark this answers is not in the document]
chiaramooney: Cool will remove that!

docs/Reference.md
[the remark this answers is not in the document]
asklar: thanks, today I learned :D
https://reactjs.org/docs/jsx-in-depth.html#props-default-to-true

windows/ExperimentalFeatures.props
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!-- Flags can be added here to effect the compilation of Microsoft.ReactNative -->
+  <PropertyGroup Label="Microsoft.ReactNative Build Flags">
+    <UseWinUI3>false</UseWinUI3>
+    <UseHermes>false</UseHermes>
+    <WinUI2xVersion>2.8.0-prerelease.210927001</WinUI2xVersion>
+    <WebView2Version>1.0.1020.30</WebView2Version>
```
asklar, line 9: note that these values won't actually be used except for when building the solution that lives in this folder. When building an app, the app's own experimentalFeatures.props will be used, not this one

chiaramooney: Yep! Just there for dev env for the module!

windows/ReactNativeWebView/ReactNativeWebView.vcxproj
```diff
@@ -158,19 +155,24 @@
       <Private>false</Private>
     </ProjectReference>
   </ItemGroup>
+  <Import Project="$(SolutionDir)\ReactNativeXaml.Imports.props" Condition="Exists('$(SolutionDir)\ReactNativeXaml.Imports.props')" />
```
asklar, line 158: why is this here?

asklar (changes requested):

windows/ReactNativeWebView.sln
```diff
@@ -45,28 +45,19 @@ Global
 		..\node_modules\react-native-windows\Shared\Shared.vcxitems*{f7d32bd0-2749-483e-9a0d-1635ef7e3136}*SharedItemsImports = 4
 	EndGlobalSection
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
 		Debug|ARM = Debug|ARM
 		Debug|ARM64 = Debug|ARM64
 		Debug|x64 = Debug|x64
 		Debug|x86 = Debug|x86
+		Release|Any CPU = Release|Any CPU
```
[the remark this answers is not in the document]
chiaramooney: Will do!

windows/ReactNativeWebView/ReactNativeWebView.vcxproj
```diff
@@ -48,7 +49,6 @@
     <PlatformToolset>v140</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)' == '15.0'">v141</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>
-    <PlatformToolset Condition="'$(VisualStudioVersion)' == '17.0'">v143</PlatformToolset>
```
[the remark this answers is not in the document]
chiaramooney: Yes will do!

windows/ReactNativeWebView/ReactNativeWebView.vcxproj
chiaramooney: I'll remove. Must have been added on accident.

docs/Getting-Started.md
```diff
@@ -95,7 +95,7 @@ If you want access to the WebView2 control in your app, follow these steps:
 
   iii. Customize your app's WinUI 2.x version to version 2.8.0-prerelease.210927001 or higher. See [here](https://microsoft.github.io/react-native-windows/docs/customizing-sdk-versions) for instructions. The WinUI 2.x support for WebView2 is not yet available in "stable" releases, so for now you will need to use a prerelease version.
   
-  iv. Customize your app's WebView2 version to version 1.0.1020.30. See [here](https://microsoft.github.io/react-native-windows/docs/customizing-sdk-versions) for instructions.
+  iv. You may need to specify the `Microsoft.Web.WebView2` package in your app's `packages.config` file. If this is needed, you will get a build error listing the version of the package that you needed to specify. Simply add the package to your `packages.config`, and you should be good to go.
```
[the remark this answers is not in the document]
chiaramooney: Yes! So there's a quirk with package reference where depending on if the app was created before 0.68 or after and uses community modules, you may need to specify packages in packages.config otherwise the community modules will fail to build. Added that note here in case users hit that case.

asklar (approved):

exotexot:
Does this WebView2 implementation allow for postMessage?

asklar:
> Does this WebView2 implementation allow for postMessage?

@exotexot not yet. See #2390

src/WebView.windows.tsx
```diff
@@ -67,34 +69,38 @@ export default class WebView extends React.Component<WebViewSharedProps, State>
 
   webViewRef = React.createRef<NativeWebViewWindows>();
 
+  RnwVersionSupportsWebView2 = (version.major>1 || version.minor>=68);
+  
+  RCTWebViewString = (this.RnwVersionSupportsWebView2 && this.props.useWebView2) ? 'RCTWebView2' : 'RCTWebView';
```
francisco-mural, line 74: Could be possible the (react native used) version is 1.68+ is running but the system native `RCTWebView2` module is not available in the system?
(could be an outdated system)
Is there any reliable way of checking if the `RCTWebView2` is available?

megamingus: how about windows versions that are not compatible with webview2? such as the `Window team` device family? (the one used for Surface Hubs)

asklar: If the runtime is not installed, the app will show a message where the webview would be, telling you to click on this link to install the runtime. 

The problem is afaik Surface Hub can only install stuff from the store, and the webview2 runtime is not available in the store. 
I'm not on the webview2 team, but if you have feedback/questions about it, the right issue to raise them is MicrosoftEdge/WebView2Feedback#1851

exotexot:
@chiaramooney 

I followed this Guide: https://github.com/react-native-webview/react-native-webview/blob/master/docs/Getting-Started.md#3-webview2-support

But how do I know if my webview component is using webview1 or webview2 ?

`import React from 'react';
import { SafeAreaView, Text } from 'react-native';

const App = () => {

const html = <script> ${alert(window.chrome)} </script>

return (
<SafeAreaView style={{flex: 1}}>
<WebView
useWebView2={true}
originWhitelist={['*']}
source={{html: html}}
/>

);
}

export default App`

With this small snippet, the alert always gives me undefined.

When trying with
const html = <script> ${alert(navigator.userAgent)} </script>

I'll get always the same return: "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36 Edg/108.0.1462.76" with useWebView2 being true or false.

Anyway to deteremine If I'm actually experiencing webview2 ???

PR #2308, opened 2021-12-30
Add iOS content blocking to WKWebView

asamiller:
`WKWebView` on iOS supports [content blocking](https://developer.apple.com/documentation/safariservices/creating_a_content_blocker). This is defined via a JSON string that is passed to the native side and then compiled into bytecode. It takes a set of rules that are used to block content. I created types for the rules based on Apple's doc. That way the rule can be written in Typescript and then stringified before being passed to the native side.

```typescript
const CONTENT_BLOCK: IOSContentRule[] = [
  {
    trigger: {'url-filter': '.*', 'resource-type': ['image']},
    action: {type: 'block'},
  },
];

return (<WebView
  originWhitelist={['*']}
  source={{uri: 'https://google.com'}}
  contentRuleList={CONTENT_BLOCK}
/>);
```

### DIFF
--- a/apple/RNCWebView.h
+++ b/apple/RNCWebView.h
@@ -83,6 +83,7 @@ typedef enum RNCWebViewPermissionGrantType : NSUInteger {
 #if !TARGET_OS_OSX
 @property (nonatomic, weak) UIRefreshControl * _Nullable refreshControl;
 #endif
+@property (nonatomic, copy) NSString * _Nullable contentRuleList;
 
 #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000 /* iOS 13 */
 @property (nonatomic, assign) WKContentMode contentMode;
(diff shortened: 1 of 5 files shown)


jamonholmgren:
@asamiller This looks really good.

> Is there a way to get React's key for the component on the native side?

I think we could just make a private property on the class instance and use a random string for that. I don't think it has to be anything special, just unique.

Like, add it here:

https://github.com/react-native-webview/react-native-webview/blob/4b3a5deaf6cba97436107f7a2c32914a77038b69/apple/RNCWebView.m#L90

```objc
@property (nonatomic, strong) NSString *contentRuleListIdentifier;
```


Set it in the `initWithFrame` initializer to something random.

```objc
_contentRuleListIdentifier = [[NSUUID UUID] UUIDString];
```

Then use it.

asamiller:
@jamonholmgren That works great! It's updated now to have the UUID.

Jonovono:
Amazing, whats the status on merging this bad boi? Should I work off the branch or wait for merge?

@asamiller @jamonholmgren

OrkhanAlikhanov:
I think this should have been merged. Looks great to me

PR #2309, opened 2021-12-30
Bump Node to Restore Pipeline

chiaramooney:
**_What_**
Current Node version is not compatible with @jest/create-cache-key-function. Bumping to restore pipeline.


chiaramooney:
Issue solved in #2312

asklar:
hadn't seen that you had a pr out for this as well, sorry :)

PR #2312, opened 2022-01-04
chore: Update node version in CI to 14

asklar:
CI started to fail as some minor version update of a dependency started to require a higher version of node than we were using (12.9.1):

```
yarn install v1.22.17
[1/4] Resolving packages...
[2/4] Fetching packages...
error @jest/create-cache-key-function@27.3.1: The engine "node" is incompatible with this module. Expected version "^10.13.0 || ^12.13.0 || ^14.15.0 || >=15.0.0". Got "12.9.1"
error Found incompatible module.
```


chiaramooney (approved):
👍

PR #2275, opened 2021-12-06
fix(ios): build failing in Xcode 12

grit96:
Add version check around `_mediaCapturePermissionGrantType`.

Fixes #2262 and #2270

PR #2313, opened 2022-01-05
fix(macos): mediaCapturePermissionGrantType is iOS only

tido64:
macOS fails to build because `_mediaCapturePermissionGrantType` is defined only for iOS:
https://github.com/react-native-webview/react-native-webview/blob/55b8a7702c82cc7b1713474daf6edf117f934d83/apple/RNCWebView.h#L95-L97


Titozzz (approved):

grit96:
Any reason #2275 wasn't merged when it implements the same change a month ago?

PR #2323, opened 2022-01-12
Prevent crash when trying to download blob urls

karthikiyengar:
This PR is a simple fix that addresses crashes on Android when trying to Download blobs. I've tested this with the example app, here's a video of the user experience:

https://user-images.githubusercontent.com/8260207/149129687-842c8751-2499-4fae-8006-c56b519cdad2.mp4

I have also tested downloading regular, non-blob files and it works fine. 

It would be great to get this merged, given that it's a simple (hopefully non-controversial) fix for a severe problem that affects a lot of users 🙏 

Prior art: 
https://github.com/react-native-webview/react-native-webview/pull/1216
https://github.com/react-native-webview/react-native-webview/pull/1546
https://github.com/react-native-webview/react-native-webview/pull/2091

Prior issues:  
https://github.com/react-native-webview/react-native-webview/issues/1090
https://github.com/react-native-webview/react-native-webview/issues/295
https://github.com/react-native-webview/react-native-webview/pull/1760
https://github.com/react-native-webview/react-native-webview/issues/1945
https://github.com/react-native-webview/react-native-webview/issues/2073

PR #2328, opened 2022-01-14
Prevent crash from unsupported URI downloads

karthikiyengar:
As discussed [here](https://github.com/react-native-webview/react-native-webview/pull/2323#issuecomment-1012654602), here's a `try/catch` implementation to mitigate [blob crashes](https://github.com/react-native-webview/react-native-webview/issues/1090), with the block restricted to the call site of the exception. Here's a video of the current behaviour. It **might** make sense to send a toast about the download being unsupported. Please let me know if you feel like that's desirable.

https://user-images.githubusercontent.com/8260207/149483121-5a31a03e-7d00-4b31-a9d2-a01249c5997b.mp4

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -90,6 +90,7 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.io.UnsupportedEncodingException;
+import java.lang.IllegalArgumentException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLEncoder;
@@ -215,7 +216,14 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
         RNCWebViewModule module = getModule(reactContext);
 
-        DownloadManager.Request request = new DownloadManager.Request(Uri.parse(url));
+        DownloadManager.Request request;
+        try {
+          request = new DownloadManager.Request(Uri.parse(url));
+        } catch (IllegalArgumentException e) {
+          System.out.println("Unsupported URI, aborting download" + e.toString());
+          e.printStackTrace();
+          return;
+        }
 
         String fileName = URLUtil.guessFileName(url, contentDisposition, mimetype);
         String downloadMessage = "Downloading " + fileName;
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -220,8 +220,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
         try {
           request = new DownloadManager.Request(Uri.parse(url));
         } catch (IllegalArgumentException e) {
-          System.out.println("Unsupported URI, aborting download" + e.toString());
-          e.printStackTrace();
+          Log.w(TAG, "Unsupported URI, aborting download", e);
           return;
         }
 
@@ -236,8 +235,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
           String cookie = CookieManager.getInstance().getCookie(baseUrl);
           request.addRequestHeader("Cookie", cookie);
         } catch (MalformedURLException e) {
-          System.out.println("Error getting cookie for DownloadManager: " + e.toString());
-          e.printStackTrace();
+          Log.w(TAG, "Error getting cookie for DownloadManager", e);
         }
 
         //Finish setting up request
@@ -1000,7 +998,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
         if (!topWindowUrl.equalsIgnoreCase(failingUrl)) {
           // If error is not due to top-level navigation, then do not call onReceivedError()
-          Log.w("RNCWebViewManager", "Resource blocked from loading due to SSL error. Blocked URL: "+failingUrl);
+          Log.w(TAG, "Resource blocked from loading due to SSL error. Blocked URL: "+failingUrl);
           return;
         }
 
@@ -1109,10 +1107,10 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
         super.onRenderProcessGone(webView, detail);
 
         if(detail.didCrash()){
-          Log.e("RNCWebViewManager", "The WebView rendering process crashed.");
+          Log.e(TAG, "The WebView rendering process crashed.");
         }
         else{
-          Log.w("RNCWebViewManager", "The WebView rendering process was killed by the system.");
+          Log.w(TAG, "The WebView rendering process was killed by the system.");
         }
 
         // if webView is null, we cannot return any event


TheAlmightyBob (changes requested):
While a toast is a very nice idea (I was pondering it as well), it runs into the problem of appropriate messaging/localization. And different developers/apps may wish to handle this in different ways (e.g. in-app error handling, logging, analytics, etc). So I'm going to make the same suggestion I did on #1546, which is to add a new error event (perhaps `onDownloadError`) to raise when this happens so apps can handle it however they wish.

karthikiyengar:
@TheAlmightyBob - Thanks for your review. While I agree that this should be the preferred way to handle the error, IMHO the addition of a new API should be handled in a separate issue. I have created one [here](https://github.com/react-native-webview/react-native-webview/issues/2335). I am happy to help with that, however, I don't have as much context around non-Android platforms.

This PR simply aims to achieve parity with the behaviour on iOS and mitigates an easily avoidable app crash which many users are suffering from. The solution proposed in this PR was very close to getting merged in the previous review, only barring the scope of the `try/catch` [here](https://github.com/react-native-webview/react-native-webview/pull/1546#pullrequestreview-467943002). There has also been https://github.com/react-native-webview/react-native-webview/pull/2065 which also closed due to inactivity. 

As highlighted in the previous PR, a lot of users have resorted to patching the library with this exact same fix (or some variant of it), as the primary concern seems to be mitigating crashes on Android. I am not sure what your stance is on merging this solution, but if the maintainers feel like the addition of `onDownloadError` is a blocker for releasing this fix, we will sadly have to resort to patching as well.

TheAlmightyBob:
> I don't have as much context around non-Android platforms

I think the event could be documented as Android-only, at least for now. Or do you mean that you don't have much context with the React Native side?

Thanks for the iOS example. That at least has the yellowbox warning to communicate the problem to developers, rather than a silent failure. Although, to the previous point, I suppose that if it is showing a yellowbox warning that suggests there is an error that could theoretically also be raised on iOS via event... (I'm actually rather curious how that warning is being shown... I don't see any code in react-native-webview that would do it)

I'm aware of the history of this issue and would also like to see it addressed, I'm just a bit wary of hiding information from developers that could help them fix the root cause.

Using the Android logging class to log a warning (and passing the exception to it) seems like it would at least be better than using `println`/printStackTrace`, no? (I realize `onDownloadStart` currently uses that pattern, but it's the only place in `RNCWebViewManager` that does so I'm not sure it's the example to follow)

(also I'm sorry I didn't start this conversation as a review comment in the code so this could be a thread...)

karthikiyengar:
> I'm actually rather curious how that warning is being shown... I don't see any code in react-native-webview that would do it

I can't point out the exact callsite for now, because I'm fighting with XCode 13 incompatibilities, but it definitely looks like the yellow box is coming from `react-native-webview`. The `iOS` code seems to be using the `RCTLogWarn` API in multiple places, allowing it to create a yellow box from native. I did not have any userland code doing `console.warn`s above example - it was exactly the same as the one I ran on Android.

```jsx
<WebView
  source={{
  }}
  automaticallyAdjustContentInsets={false}
/>
```
Here is a longer stacktrace for the iOS yellowbox:
image

If we want to do do yellow boxes on Android, we could use `com.facebook.react.util.RNLog`. However, this is only available starting react-native 0.63 and adding it might break compatibility for users on older versions. I'm not entirely sure how we're enforcing React Native versions in this project.

Aside: I gave some further thought to adding an `onDownloadError`. Upon reading @avinashsahoo's comment from [here](https://github.com/react-native-webview/react-native-webview/pull/2091#issuecomment-1012994881), it looks like `onShouldStartLoadWithRequest` is only called during page navigation and not on file downloads (contrary to what is described in the comment for iOS). Maybe that would be a better way to delegate choice to the user.

TheAlmightyBob:
I thought about this more and I think you made a reasonable point for why the error event, while useful, does not strictly need to be part of this PR. I think I'm good with this if the logging switches to use the Android logger (unless you have an argument for why it shouldn't?).

> The iOS code seems to be using the RCTLogWarn API in multiple places

Yeah, but I looked through those and none of them seemed related... I'm not blocking this PR on that question though, I'm just curious.

> If we want to do do yellow boxes on Android, we could use com.facebook.react.util.RNLog. However, this is only available starting react-native 0.63

Okay, that seems like a decent reason to hold off for now.

> Aside: I gave some further thought to adding an onDownloadError. Upon reading @avinashsahoo's comment from here, it looks like onShouldStartLoadWithRequest is only called during page navigation and not on file downloads (contrary to what is described in the comment for iOS). Maybe that would be a better way to delegate choice to the user.

Interesting thought... I'm not sure if I'm quite following you though. "Delegating choice" seems like a different thing from "handling and reporting errors."

karthikiyengar:
> Interesting thought... I'm not sure if I'm quite following you though. "Delegating choice" seems like a different thing from "handling and reporting errors."

Yes, you're right. We shouldn't mix up the two. 
- onShouldStartLoadWithRequest - Here, I believe that the user should be able to prevent **downloads** too, depending on the URL in question. Currently, we don't allow this - the user can allow/disallow navigation, but not downloads.  
- onDownloadError - My concern is that this API might give users the impression that we're handling more than we really are (e.g failed downloads due to bad network). `blob` downloads not working seems like a limitation more than an error to me.

If I understand your comment correctly, it looks like I've managed to address most of your concerns around this PR. If that's not the case, please feel free to let me know.

TheAlmightyBob:
> onDownloadError - My concern is that this API might give users the impression that we're handling more than we really are (e.g failed downloads due to bad network). blob downloads not working seems like a limitation more than an error to me.

That's a fair concern about the name, but I still think the issue should be communicated in some way so that developers can handle it (e.g., fixing the website so it doesn't try to download via blob scheme, provide some friendly error to the end-user, etc). Perhaps there's a better name... #2091 used `onDownloadUnsupportedUrl`, but I worry that could have the opposite problem, that maybe there are other reasons that the download manager could throw an error?

(at this point though this conversation should probably be happening in #2335 rather than here...)

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -215,7 +216,14 @@ public void onDownloadStart(String url, String userAgent, String contentDisposit
 
         RNCWebViewModule module = getModule(reactContext);
 
-        DownloadManager.Request request = new DownloadManager.Request(Uri.parse(url));
+        DownloadManager.Request request;
+        try {
+          request = new DownloadManager.Request(Uri.parse(url));
+        } catch (IllegalArgumentException e) {
+          System.out.println("Unsupported URI, aborting download" + e.toString());
+          e.printStackTrace();
```
TheAlmightyBob, line 224: This is my remaining concern at this point. I think this should be using the Android Logger to log a warning (and pass the exception to it, which the logger accepts as a parameter) rather than using `println` & `printStackTrace`.

TheAlmightyBob (changes requested):

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
karthikiyengar: Could you please elaborate on what you mean by the Android logger? I was able to see this stack trace printed out in Android Studio.

If you mean RNLog from the previous discussion, we would need a React Native version bump first.

TheAlmightyBob: No, I don't mean RNLog (I also don't mean `FLog` either... I'm not familiar with that one and unsure why it's used). I mean just `Log.w`. If you search the code you'll find examples, such as https://github.com/react-native-webview/react-native-webview/blob/b3cba2bb8e7d2ddca4cf852ceef6169c82699a33/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L1003

So I mean something like (I didn't notice GitHub supported "suggestions" 'til now):
```suggestion
          Log.w(TAG, "Unsupported URI, aborting download", e);
```

I admit I haven't actually tried using `println` on Android (I've always used `Log`), so I don't know exactly how they compare, but my understanding from what I read is that `println`/`printStackTrace` outputs in a way that could be easy to miss (e.g., if you filter to only your app, those lines might not show up?).

karthikiyengar: TIL. Thanks. I took the liberty of cleaning up a few of the other logs.

TheAlmightyBob: Thanks! _Normally_ I'd vote to have other cleanup in a separate PR, but this looks pretty small & straightforward... especially switching to `TAG` uses

TheAlmightyBob (approved):

TheAlmightyBob:
Thanks @karthikiyengar!

PR #2332, opened 2022-01-17
Adding french translation for the README

InTheNooB:


docs/README.french.md
```diff
@@ -0,0 +1,114 @@
...
+
+Ce projet suit les spécifications [all-contributors](https://github.com/all-contributors/all-contributors). Les contributions de toute nature sont les bienvenues !
+
+## License
+
+MIT
+
+## Traductions
+
+This readme is available in:
```
TheAlmightyBob, line 109: Should this sentence not also be translated?

TheAlmightyBob (changes requested):

docs/README.french.md
InTheNooB: It could indeed. Thinking about it, a better solution would be not to translate any of this last part (even the title), because if someone needs to change the language of the README, it means that the dont speak the current language, and thus they would have to still be able to navigate through READMEs using this part which would stay in english in every translation.

TheAlmightyBob: Interesting logic. I would think though that if they didn't speak French they wouldn't have found themselves on the French version of the README in the first place, so I'm not quite sure this would help...? But it doesn't bother me either. I just wanted to make sure this was intentional and not just a line you missed by accident.

TheAlmightyBob (approved):

TheAlmightyBob:
Merci @InTheNooB!

PR #2346, opened 2022-01-25
new windows issue on ios

hlta:


hlta:
> @hlta can you describe the issue that this fixes more clearly?

@milkman4  

This addresses two issues:

1. Open a link in new windows 
2. Fix the problem on the checkout page if you are suing (Afterpay , PayPal ... ). These payment provider required to have a checkout page instance open and open their security page in the new window

jackkinsella:
I think this is still relevant.

milkman4:
@jackkinsella yes, i agree

PR #2362, opened 2022-02-02
Mark android-only methods as possibly undefined

### LINKED ISSUE #2361: Fix types for Android-only WebView methods
**Bug description:** The TypeScript types allow writing unsafe code that can cause runtime errors

**To Reproduce:** Write code that calls `webview.clearHistory()` and run the typescript compiler, confirm it passes. Then run it in iOS, confirm you get a runtime error.

**Expected behavior:** Android-only methods should be `| undefined` on the interface so that developers are forced to write safer code.

**Environment:**
 - OS: N/A
 - OS version: N/A
 - react-native version: N/A
 - react-native-webview version: 11.17.1

wbobeirne:
Closes #2361. Makes android-only methods possibly undefined properties on `WebView`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -46,19 +46,19 @@ declare class WebView<P = {}> extends Component<WebViewProps & P> {
      * (Android only)
      * Removes the autocomplete popup from the currently focused form field, if present.
      */
-    clearFormData: () => void;
+    clearFormData?: () => void;
 
      /**
      * (Android only)
      * Clears the resource cache. Note that the cache is per-application, so this will clear the cache for all WebViews used.
      */
-    clearCache: (clear: boolean) => void;
+    clearCache?: (clear: boolean) => void;
 
      /**
      * (Android only)
      * Tells this WebView to clear its internal back/forward list.
      */
-    clearHistory: () => void;
+    clearHistory?: () => void;
 }
 
 export {WebView};


TheAlmightyBob (approved):
This makes sense to me, I'm just uncertain if this should be considered a "breaking" change since it may require folks to update code (e.g., even if they were calling it after an `if (Platform.OS === 'android')` check, that wouldn't be good enough anymore because Typescript doesn't know that the methods are defined for Android...)

A possibly safer approach could be to have these methods be defined but no-op on iOS, similar to platform-specific properties/events... that's no more "correct" than what you have here, but could be less-disruptive...?

wbobeirne:
> A possibly safer approach could be to have these methods be defined but no-op on iOS

Seems reasonable, but that actually _could_ be a breaking runtime change since some folks may have written conditionals like:

```
if (webview.clearHistory) {
  /* do android specific things here */
}
```

Given that the actual code implementation is `undefined` right now, this is more of a bug fix than a change to the API really, it just describes the actual state of the correctly.

And while your point about checking `Platform.OS` (or any other androidy check) makes total sense, it's still an implicit relationship between `Platform.OS` and `WebView`. I think most who feel strongly about structuring their TS codebases to be strict would appreciate being forced to not rely on implicit relationships like that, but that's just my 2 cents!

TheAlmightyBob:
> > A possibly safer approach could be to have these methods be defined but no-op on iOS
> 
> Seems reasonable, but that actually _could_ be a breaking runtime change since some folks may have written conditionals like:
> 
> ```
> if (webview.clearHistory) {
>   /* do android specific things here */
> }
> ```

Fair point. While that seems much less likely to me, it is possible.
 
> Given that the actual code implementation is `undefined` right now, this is more of a bug fix than a change to the API really, it just describes the actual state of the implementation correctly.
> 
> And while your point about checking `Platform.OS` (or any other androidy check) makes total sense, it's still an implicit relationship between `Platform.OS` and `WebView`.

But a documented relationship, at least.

> I think most who feel strongly about structuring their TS codebases to be strict would appreciate being forced to not rely on implicit relationships like that, but that's just my 2 cents!

Also a reasonable point...

PR #2368, opened 2022-02-07
docs: add more details on onContentProcessDidTerminate

Almouro:
We were logging every call to `onContentProcessDidTerminate` as an unexpected error on our app. 

Turns out it was completely expected to happen at some point. 
I added more details from [the Apple docs](https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455639-webviewwebcontentprocessdidtermi) to make users aware that WKWebView live in completely different processes than the app. As such they are not included in the total RAM of the app, and can be terminated independently of the app to liberate some new apps.

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -573,11 +573,14 @@ url
 
 ### `onContentProcessDidTerminate`[⬆](#props-index)<!-- Link generated with jump2header -->
 
-Function that is invoked when the `WebView` content process is terminated.
+Function that is invoked when the `WebView` content process is terminated. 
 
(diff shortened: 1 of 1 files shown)


TheAlmightyBob (approved):

rohitiglas:
App is crashing for IOS , Is it fixed or not ?

Almouro:
Hi @rohitiglas, this PR is only a documentation change, if you have a crash that you think is related to the lib, I think you can submit a separate issue

TheAlmightyBob:
Thanks @Almouro !

PR #2419, opened 2022-03-26
feat(windows): Add WebView2 UWP Desktop support

asklar:
making this pr to trigger publish, we didn't use the right tag in the PR: #2304

in the process, it turns out the certificate for the test app had expired so I removed it (and made CI deploy from layout instead).


AgneLukoseviciute (approved):

AgneLukoseviciute (approved):

chiaramooney (approved):
👍

jeveloper:
hey @chiaramooney  Not a big deal but for non Windows platforms, there is now a warning that pops up which may confuse a lot of developers. 
May i suggest or shall i add a note somewhere in the readme? 
"Cannot find module 'react-native-test-app/package.json'" 
This can be safely ignored, this will occur when metro starts up.

Cheers and thank you

asklar:
> hey @chiaramooney Chiara Mooney FTE Not a big deal but for non Windows platforms, there is now a warning that pops up which may confuse a lot of developers. May i suggest or shall i add a note somewhere in the readme? "Cannot find module 'react-native-test-app/package.json'" This can be safely ignored, this will occur when metro starts up.
> 
> Cheers and thank you

@tido64 any idea why that might be?

tido64:
> "Cannot find module 'react-native-test-app/package.json'" This can be safely ignored, this will occur when metro starts up.

How do you repro this? I've tested 11.18.0 on both macOS and Windows and don't see this message.

PR #2432, opened 2022-03-31
Don't crash if trying to download "invalid" URIs on Android

### LINKED ISSUE #2428: Downloading unsupported filenames on Android 10 crashes app
**Bug description:**

On Android 10 specifically (not 9, not 11, not 12), downloading a file with an extra dot before the extension (e.g. `my..file`) crashes the app. There may be other filename quirks that also cause this, that happens to be the specific scenario that one of our users hit and we haven't taken the time to investigate what else might also fail.

This is related to, but not quite the same as, #1090. In this case, the root cause appears to be an Android bug, not really an intentionally "unsupported" input that would be expected to throw. However, `react-native-webview` still shouldn't _crash_ when it happens.

Also, the error is coming from a different place... this time it's being thrown by the `enqueue` call in `RNCWebViewModule.downloadFile`.

The error is
> java.lang.IllegalArgumentException: Invalid file URI

It would be nice to say "Android fixed it, job done!", but of course we all know we can't just assume everybody is running the latest version of Android.... 😞 

**To Reproduce:**

You can use the example app in this repo. Modify the `App.tsx` so that the `Download` option is enabled on Android, and change the URL to download in `Downloads.tsx` from 7zip to `https://github.com/TheAlmightyBob/react-native-webview/raw/test/Test..zip`. Run in an Android 10 emulator (or device if you happen to have one!), try to download, watch it crash. Run on Android 9/11/12, and it should successfully download.

**Expected behavior:**

No crash. I'm not sure we can do anything about it failing to download on Android 10, but it shouldn't crash.

**Environment:**
 - OS: Android
 - OS version: 10
 - react-native version: 0.66.4 and whatever the Example app is using (I'm not too familiar with `react-native-test-app`, though it sure seems interesting!)
 - react-native-webview version: 11.18.0

TheAlmightyBob:
Fixes #2428.

This is based on @karthikiyengar's previous fix in #2328.

If desired, I could also include the tweaks in my [test branch](https://github.com/TheAlmightyBob/react-native-webview/tree/test) for reproing the issue in the example app.... however, since it is an _example_ app and not a _bug regression test_ app, I'm not sure that would be appropriate.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
@@ -311,7 +311,12 @@ public class RNCWebViewModule extends ReactContextBaseJavaModule implements Acti
   public void downloadFile(String downloadingMessage) {
     DownloadManager dm = (DownloadManager) getCurrentActivity().getBaseContext().getSystemService(Context.DOWNLOAD_SERVICE);
 
-    dm.enqueue(this.downloadRequest);
+    try {
+      dm.enqueue(this.downloadRequest);
+    } catch (IllegalArgumentException e) {
+      Log.w("RNCWebViewModule", "Unsupported URI, aborting download", e);
+      return;
+    }
 
     Toast.makeText(getCurrentActivity().getApplicationContext(), downloadingMessage, Toast.LENGTH_LONG).show();
   }


matinzd (dismissed):
This has been a pain for some users, AFAIK. Let’s merge it!

Titozzz (approved):

Issue #2414, opened 2022-03-24
Crash in Production - RNCWebView.m -[RNCWebView webView:decidePolicyForNavigationAction:decisionHandler:]


e-pradeep:
Fatal Exception: NSInvalidArgumentException
CoreFoundation                 0x119180 __exceptionPreprocess
libobjc.A.dylib                0x59f8 objc_exception_throw
CoreFoundation                 0x92bec _CFArgv
CoreFoundation                 0x1a898 -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]
0xb474 +[NSDictionary dictionaryWithObjects:forKeys:count:]
0x7e6618 -[RNCWebView webView:decidePolicyForNavigationAction:decisionHandler:] + 1142 (RNCWebView.m:1142)
0x29dfdc __72-[DTXWKWebView webView:decidePolicyForNavigationAction:decisionHandler:]_block_invoke + 324 (DTXWKWebView.m:324)
[DTXWKWebView webView:decidePolicyForNavigationAction:decisionHandler:canCallOriginal:callOriginal:] + 369 (DTXWKWebView.m:369)
[DTXWKWebView webView:decidePolicyForNavigationAction:decisionHandler:] + 327 (DTXWKWebView.m:327)
WebKit                         0x170e28 (Missing UUID 3a7b114aadeb3ae2b02caf882114d8ef)
WebKit                         0x1d7ddc (Missing UUID 3a7b114aadeb3ae2b02caf882114d8ef)
WebKit                         0x1d8480 (Missing UUID 3a7b114aadeb3ae2b02caf882114d8ef)
WebKit                         0x1d8344 (Missing UUID 3a7b114aadeb3ae2b02caf882114d8ef)
WebKit                         0x3e5d08 (Missing UUID 3a7b114aadeb3ae2b02caf882114d8ef)
WebKit                         0x3db820 (Missing UUID 3a7b114aadeb3ae2b02caf882114d8ef)
WebKit                         0x39ea0 (Missing UUID 3a7b114aadeb3ae2b02caf882114d8ef)
WebKit                         0x1f9918 (Missing UUID 3a7b114aadeb3ae2b02caf882114d8ef)
WebKit                         0x2da30 (Missing UUID 3a7b114aadeb3ae2b02caf882114d8ef)
WebKit                         0x2a96c (Missing UUID 3a7b114aadeb3ae2b02caf882114d8ef)
[...]

Issue #2051, opened 2021-06-15
Consider checking for nil URLs in WKNavigationDelegate


msbit:
When looking into https://stackoverflow.com/questions/67778548/using-react-native-webview-in-ios-app-is-crashing-intermittently it appears to me that in the delegate methods:

* [`webView:decidePolicyForNavigationAction:decisionHandler:`][1], and
* [`webView:decidePolicyForNavigationResponse:decisionHandler:`][2]

there is the potential for `request.URL` and `request.mainDocumentURL` to be `nil`, which very likely causes the error in that issue:

`-[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[x]'`

when using them in the dictionary literal construction (`@{...}`).

In the [same file][3], a default empty NString is used to guard against this when creating the base event, perhaps something similar could used here too?

[1]: https://github.com/react-native-webview/react-native-webview/blob/e33eea6dcf43b43e96b3a2dfffc689fbd54e98ea/apple/RNCWebView.m#L947
[2]: https://github.com/react-native-webview/react-native-webview/blob/e33eea6dcf43b43e96b3a2dfffc689fbd54e98ea/apple/RNCWebView.m#L1018
[3]: https://github.com/react-native-webview/react-native-webview/blob/e33eea6dcf43b43e96b3a2dfffc689fbd54e98ea/apple/RNCWebView.m#L786

msbit:
@e-pradeep no more movement from my end. As noted it was something I'd observed as a possible issue, but addressing this at this level wouldn't necessarily solve what you are describing, it would likely just ensure that the error isn't thrown; `request.mainDocumentURL` would still be `nil`.

PR #2442, opened 2022-04-11
Added a null check. Noticed that sometimes these values were getting nil but still added to dictionary causing a crash

DSFlash:
Potential breaking change fix. kindly review.

Added a null check. Noticed that sometimes these values were getting nil but still added to dictionary causing a crash.

https://github.com/react-native-webview/react-native-webview/issues/2414
https://github.com/react-native-webview/react-native-webview/issues/2051

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -1137,7 +1137,7 @@ NSString *const CUSTOM_SELECTOR = @"_CUSTOM_SELECTOR_";
   NSURLRequest *request = navigationAction.request;
   BOOL isTopFrame = [request.URL isEqual:request.mainDocumentURL];
 
-  if (_onShouldStartLoadWithRequest) {
+  if (_onShouldStartLoadWithRequest && request != nil && request.mainDocumentURL != nil) {
     NSMutableDictionary<NSString *, id> *event = [self baseEvent];
     [event addEntriesFromDictionary: @{
       @"url": (request.URL).absoluteString,
@@ -1153,7 +1153,7 @@ NSString *const CUSTOM_SELECTOR = @"_CUSTOM_SELECTOR_";
     }
   }
 
-  if (_onLoadingStart) {
+  if (_onLoadingStart && request != nil && request.URL != nil) {
     // We have this check to filter out iframe requests and whatnot
     if (isTopFrame) {
       NSMutableDictionary<NSString *, id> *event = [self baseEvent];


TheAlmightyBob:
Thanks for digging in. This is quite curious. You say "sometimes"... were you able to figure out why or have any kind of repro? (how did you check that this resolved the issue?)

As you say, it's "potentially breaking"... that's more concerning if we have no idea how this happens. Perhaps it would be safer to still raise the event but just skip adding the blank entries?

HudiDev:
I have no idea why this was closed!! 

It's a real issue.

Why can't this be approved and merged? all it's adding is null checks!

TheAlmightyBob:
> I have no idea why this was closed!!

It was closed because the submitter disappeared and did not respond to questions/feedback.

> It's a real issue.

Do you know why the issue is happening? Do you have a repro? Have you tried patching this change into your app and verifying that it fixes and issue for you and does not introduce a new one?

> all it's adding is null checks!

The submitter themself identified this as a "potentially breaking change." It skips logic that apps may be relying upon, potentially just trading one bug for another (and the new bug may be even harder to identify & fix if end users hit it).

HudiDev:
> Do you know why the issue is happening?
Well judging based on our crashlytics logs it seems that in some radical cases the request might be null when`- (void)                  webView:(WKWebView *)webView
  decidePolicyForNavigationAction:(WKNavigationAction *)navigationAction
                  decisionHandler:(void (^)(WKNavigationActionPolicy))decisionHandler` get's invoked. 

>  Do you have a repro?
Due to the crash inconsistency we were encountering we were unable to log a clear repro path. Nevertheless a crash was happening from time to time. Like I said, we have crashlytics logs to show this and we are happy to share some of the logs.

> Have you tried patching this change into your app and verifying that it fixes an issue for you and does not introduce a new one?
Yes we have patched. the patch seems to work simple because we are not receiving any crash reports from the problematic area since the patch was deployed.

TheAlmightyBob:
@HudiDev Appreciate the response, and thanks for reporting back that patching this change in helped you.

I agree that it seems `mainDocumentURL` being `nil` is the culprit here and should be handled. My concern is only with the specific way that was done in this PR.

For example, with the change patched into your app, you've confirmed that it no longer crashes. But presumably your app is still seeing requests with `mainDocumentURL` being `nil`... the difference now is that it's completely bypassing `onShouldStartLoadWithRequest` when that happens. Perhaps your app doesn't rely on that event, so it wouldn't matter. But another app might rely on that event, and could wind up with some other mysterious bug that they don't even know about (since it's not reported as a crash). You are right that currently such an app would crash... I'm not suggesting that this would break a currently-working case, only that it might replace a crash with a different bug.

If `mainDocumentURL` is optional in `NSURLRequest`, I don't see why we shouldn't still raise the `onShouldStartLoadWithRequest` event and just make it optional there too.

HudiDev:
@TheAlmightyBob Thanks for the detailed response. 
I indeed opened a new PR here: [https://github.com/react-native-webview/react-native-webview/pull/2662](url)
This PR only checks for null on `mainDocumentURL`. 
`onShouldStartLoadWithRequest` runs as usual

Kindly review it.

PR #2444, opened 2022-04-12
fix ex, add js ex, add line, add closing backtick

HyunseungLee-Travis:
## webview2 example
I have noticed 3 other example used `={true}` syntax. 
1. Other examples used `={true}`, so I believed using this syntax is correct.
2. I also think it is much more beginner friendly.

## Removed JS example for clearCache()
There were no example for clearCache().

## Add line
The *note* section seemed to be seperate from other explanation. 
I added a new line that was joined by other explanations.
It was also a bit hard to read when those two were connected.

## Unclosed backtick
The bactick for an example seemed to be missing. 
A whole section was shown as a code.

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -1478,6 +1478,7 @@ Example:
 ### `forceDarkOn`
 
 Configuring Dark Theme
+
 *NOTE* : The force dark setting is not persistent. You must call the static method every time your app process is started.
 
 *NOTE* : The change from day<->night mode is a configuration change so by default the activity will be restarted and pickup the new values to apply the theme. Take care when overriding this default behavior to ensure this method is still called when changes are made.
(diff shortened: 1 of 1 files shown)


TheAlmightyBob (approved):

TheAlmightyBob:
I'm not sure why CircleCI isn't seeing this or how to make it do so.... this branch doesn't even appear in CircleCI, so I can't run it manually I don't think.

Perhaps just force pushing your branch again might "refresh" things so CircleCI sees it? I just rebased an older PR and it ran there, so it's not _completely_ broken...

HyunseungLee-Travis:
I tried closing and re-opening the pull request.

TheAlmightyBob:
Thanks @HyunseungLee-Travis!

PR #2450, opened 2022-04-21
Deprecated issue fixed.

lalitkumarext:
mediaPlaybackRequiresUserAction change to mediaTypesRequiringUserActionForPlayback


TheAlmightyBob:
_Theoretically_ the change in this PR shouldn't be necessary because that branch of code should only be executed if the newer API is unavailable.

If you are actually seeing that branch get executed and call the deprecated API, that suggests the real problem is that `WEBKIT_IOS_10_APIS_AVAILABLE` is defined incorrectly....

I see there was an old issue on React Native ([#25411](https://github.com/facebook/react-native/issues/25411)) claiming as much, but it was resolved without action and I haven't looked any further into it...

Issue #2435, opened 2022-04-04
windows project broke on installing the library


francisco-mural:
**Bug description:**
Installing `react-native-webview` on a react native windows project makes it stop building

**To Reproduce:**
Windows 10

* `npx react-native init react_native_project`
* `cd react_native_project`
* `npx react-native-windows-init --overwrite`
* if you run here `npm run windows` it works flawlessly
* `npm install react-native-webview`
* `npm run windows` 

**Expected behavior:**
Starts the dev app as before

**Screenshots/Videos:**
```
C:\Users\myuser\workspace\react_native_project\node_modules\react-native-windows\Mso\dispatchQueue\dispatchQueue.h(728,17): error C2607: static assertion failed [C:\Users\myuser\workspace\react_native_project\node_modules\react-native-windows\Microsoft.ReactNative\Microsoft.ReactNative.vcxproj]

C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\XamlCompiler\Microsoft.Windows.UI.Xaml.Common.targets(486,5): error MSB4181: The "CompileXaml" task returned false but did not log an error. [C:\
Users\myuser\workspace\react_native_project\node_modules\react-native-windows\Microsoft.ReactNative\Microsoft.ReactNative.vcxproj]

Done Building Project "C:\Users\myuser\workspace\react_native_project\node_modules\react-native-windows\Microsoft.ReactNative\Microsoft.ReactNative.vcxproj" (default targets) -- FAILED.
Done Building Project "C:\Users\myuser\workspace\react_native_project\node_modules\react-native-webview\windows\ReactNativeWebView\ReactNativeWebView.vcxproj" (default targets) -- FAILED.
Done Building Project "C:\Users\myuser\workspace\react_native_project\windows\react_native_project\react_native_project.vcxproj" (default targets) -- FAILED.
Done Building Project "C:\Users\myuser\workspace\react_native_project\windows\react_native_project.sln" (default targets) -- FAILED.
```

**Environment:**
 - OS: Windows 11
 - OS version: 22000.556
 - react-native version: tried with 0.68 and 0.67.4 (same result)
 - react-native-webview version: 11.18.1
 - "react-native-windows":  0.68.0-preview.8 and 0.67.6 (same result)

francisco-mural:
> @francisco-mural Could you try building in VS and paste in the error messages you're getting there, so we can get a little more info!
[...]

PR #2459, opened 2022-04-26
Fix Break From Package Ref Changes

chiaramooney:
Fix break in Windows build on RNW apps v0.68+ consuming the webview module. 

Tested fix on apps upgraded to 0.68, fresh 0.68 apps, and apps 0.67-.

Resolves https://github.com/microsoft/react-native-windows/issues/9879


windows/ReactNativeWebView/ReactNativeWebView.vcxproj
```diff
@@ -170,7 +170,6 @@
   </Target>
   <ImportGroup Label="ExtensionTargets">
     <Import Project="$(SolutionDir)packages\Microsoft.Windows.CppWinRT.2.0.200316.3\build\native\Microsoft.Windows.CppWinRT.targets" Condition="Exists('$(SolutionDir)packages\Microsoft.Windows.CppWinRT.2.0.200316.3\build\native\Microsoft.Windows.CppWinRT.targets')" />
-    <Import Project="$(SolutionDir)packages\Microsoft.Web.WebView2.1.0.1020.30\build\native\Microsoft.Web.WebView2.targets" Condition="Exists('$(SolutionDir)packages\Microsoft.Web.WebView2.1.0.1020.30\build\native\Microsoft.Web.WebView2.targets')" />
```
asklar, line 173: aren't these required when not using PackageReference? In other words, apps on 67 and below would still require the packages.config and manual webview2 versions and imports, no?

chiaramooney: Nope not needed because apps v0.67 and below shouldn't be using WinUI 2.8-prerelease or higher since it isn't supported in RNW (we didn't backport the change needed to pull in webview2). So they will be using WinUI 2.7 and below and won't need the webview dependency. Then the macros we added to the WebView module prevent building the code that relies on WebView2.

asklar: I'm not following... if I'm on 0.67 and I want to use react-native-webview, without webview2, and I want to target winui 2.8 (not prerelease), how will this work?

asklar (approved):

NickLeippe:
This might also address #2435

PR #2469, opened 2022-05-04
support message passing and keeping native WebView in memory when react component is unmounted

donaldchen:
**PR Status**
Ready for review. We'd like this PR to get merged if it gets approved.

**Overview of Changes**
By default, moving a React `WebView` component in the view hierarchy creates a new native WebView and refreshes the iframe. In Discord's mobile React Native apps which use `react-native-webview`, we need the ability to move the `WebView` without refreshing the iframe, and this PR adds support for that on iOS and Android by reusing the native WebView instance. We also need to be able to pass messages between the WebView and the react-native app without a react tag (e.g. when the React component for the WebView is not mounted), and this PR adds new APIs that support that, too. To test these changes, we added a "Portals" button to the RN-webview sample app. We've also been testing these changes in the Discord app.

**High level react-native-webview API changes**
This PR updates react-native-webview with the following new APIs
- the `webViewKey` react prop- a key for accessing specific WebView instances. When this is defined, the native WebView stays in memory until explicitly released.
- `releaseWebView` - a function that explicitly releases native WebViews given a WebView key.
- the `messagingWithWebViewKeyEnabled` react prop - enables message passing without a react tag (i.e. without needing the react component to be mounted)
- `injectJavaScriptWithWebViewKey` - supports passing messages / injecting javascript to the native WebView given a WebView key even if the react component for the native WebView is not mounted anywhere
- `addOnMessageListenerWithWebViewKey` - When the `webViewKey` is defined, `onMessage` and similar callbacks like `onLoadStarted` don't work. This method adds a listener for messages from the WebView given a WebView key, and this handles messages even when the react component is unmounted.

**How This Works**

iOS Changes:
We have a singleton map where the keys are strings, and the values are `WKWebViews`. When React Native asks for a new view, we create a new `RNCWebView` instance. `RNCWebView` is a parent of `WKWebView`, and when the props say to reuse the native WebView instance, `RNCWebView` looks for the existing `WKWebView` instance from the map before it decides whether it needs to create a new `WKWebView` instance. Additionally, when we move a `WKWebView` between two `RNCWebView` parents, we need to remove the reference to the old parent that was held through the observer. We also have a map of web view keys  to `RNCWebView`s, so we can remove the correct observer from the `WKWebView`. If we don't remove the observer when we change the `WKWebView`'s parent, the app will crash with an `EXC_BAD_ACCESS` error.

Android Changes:
We renamed the old `RNCWebView`, which extended the Android `WebView` to `InternalWebView`. Then we were able to  wrap `InternalWebView` with a new view, which we named `RNCWebView`, which is a `FrameLayout`. Now that `InternalWebView` is wrapped by the `RNCWebView` view group, the Android architecture can match the iOS architecture where the `RNCWebView` has a one-to-one relationship with the react component, and the inner / native WebView does not have a one-to-one relationship with the react component. Similar to the iOS architecture, we also have singleton maps that hold references to `RNCWebView`s and `InternalWebView`s for WebView keys.

**Before and After**

Here's the behavior where the iframe refreshes when `webViewKey` undefined.

https://user-images.githubusercontent.com/3821698/165408230-849f75bc-771d-4658-8171-5b64fb1bf277.mov

Here's the behavior where the iframe maintains its state as the react WebView moves in the view hierarchy when `webViewKey` is defined.

https://user-images.githubusercontent.com/3821698/166070390-78b6e6be-701b-4e03-9bd2-b5c9c57cda17.mov

**Testing messaging passing without a react tag**
See https://github.com/discord/react-native-webview/pull/6 and https://github.com/discord/react-native-webview/pull/12 for screen recordings and explanations on how we tested the message passing between the WebView and the react-native app without a react tag.


donaldchen:
In https://github.com/react-native-webview/react-native-webview/pull/2469/commits/03d597c5008c1e132bc9c91ef249c46c2b382fa5 , I added the changes from https://github.com/discord/react-native-webview/pull/4 (and changes from some of the commits on https://github.com/discord/react-native-webview/commits/11.18.1-discord-1 that came after that PR ). While integrating this with the Discord iOS app, I realized that I needed the ability to release the WebView with a web-view-key via a static method when the React `WebView` class component wasn't available.

https://user-images.githubusercontent.com/3821698/166984290-98c15af3-716d-4844-a127-868e9a54e7a4.mp4

donaldchen:
https://github.com/react-native-webview/react-native-webview/pull/2469/commits/3a8d9c7bd076798d50ea60b484fabc972e4b246a pulls in a change from https://github.com/discord/react-native-webview/pull/5 that fixes an `EXC_BAD_ACCESS` crash that I was running into when navigating away from the WebView, backgrounding the app, foregrounding the app, and navigating back to the WebView.

mattapperson:
Overall I really like this approach, at work we have been developing this feature set as well. The API is very similar and we have it fully working on Android. Hopefully we can work together to get this feature set stable and in the open source codebase here.

donaldchen:
@mattapperson , that's exciting! Do you have this working on iOS also, or do you only have it working on Android so far? Do you have the code for your changes publicly viewable (e.g. in a fork of the repo)? If not, are you willing to share that e.g. via a PR into this repo or into https://github.com/discord/react-native-webview/commits/11.18.1-discord-1 ? I'm curious about the similarities and differences between your current approach versus my iOS changes in this PR and my proposed Android changes in the PR description

donaldchen:
> Nothing is public yet for this for our effort, but I hope to very soon.

@mattapperson , sounds good! Feel free to tag me when you all are ready to make that work public. I'm interested in collaborating to get the feature set stable and merged into this open-source / upstream react-native-webview code

donaldchen:
@TheAlmightyBob , thanks for the feedback!

@Titozzz and @jamonholmgren , do you have any thoughts on the proposal in the PR description?

Titozzz:
Fyi I'm in a major rewrite of the lib for the new architecture, so you'll probably have a lot of conflicts to handle.

donaldchen:
> Fyi I'm in a major rewrite of the lib for the new architecture

@Titozzz , in the new architecture, on Android, is there a plan to make the Android `WebView` get wrapped by an `RNCWebView` similar to what's currently on iOS? I'm wondering, since I would've had to do that anyway to make the functionality in this PR work for Android

apple/RNCWebView.m
```diff
@@ -402,50 +405,84 @@ - (void)didMoveToWindow
...
 #else
-    _webView = [[RNCWKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
+      _webView = [[RNCWKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
 #endif // !TARGET_OS_OSX
-
+        
+        if (_webView != nil && _webViewKey != nil) {
+          sharedWKWebViewDictionary[_webViewKey] = _webView;
+          NSMutableDictionary *sharedRNCWebViewDictionary= [[RNCWebViewMapManager sharedManager] sharedRNCWebViewDictionary];
+          sharedRNCWebViewDictionary[_webViewKey] = self;
```
TheAlmightyBob, line 434: What's the respective purpose of these 2 different dictionaries?

src/WebViewTypes.ts
```diff
@@ -721,6 +721,25 @@ export interface IOSWebViewProps extends WebViewSharedProps {
...
+   * It's important to call `release` on the React WebView with the corresponding webViewKey
+   * when the native WebView is no longer needed.
+   */
+  keepWebViewInstanceAfterUnmount?: boolean;
+
+  /**
+   * When keepWebViewInstanceAfterUnmount is true, if two React components use the same
+   * key for the WebView, they will use the same native WebView instance.
+   */
+  webViewKey?: string;
```
TheAlmightyBob, line 742: What happens if `keepWebViewInstanceAfterUnmount` is `true` but `webViewKey` is `undefined`?

donaldchen:
> I have implemented this feature for Android by referring to this PR and https://github.com/discord/react-native-webview/pull/2 (These helped us a lot). I've forked react-native-webview/react-native-webview to a private repo on my company's organization and added diffs like this: https://github.com/cou929/react-native-webview/pull/1
(Note: This https://github.com/cou929/react-native-webview/pull/1 branch is just for sharing, so this is not the actual code I internally use in our production. Also, in this branch, we only implemented the requirements we needed, which might be not sufficient for common use cases.)
I hope this PR is going to be merged into the master eventually due to the maintenance cost of our forked branch. But it looks like it hasn't been updated in months. Do you have any plans for this PR? Is there anything I can do to help move forward?

@cou929 , that's great that you got this working! Coincidentally, @nealmanaktola (my team member / coworker) got this working on Android in Discord's fork of `react-native-webview` last week in https://github.com/discord/react-native-webview/pull/11 and https://github.com/discord/react-native-webview/pull/12 .

@TheAlmightyBob , if we were to get this PR updated with both iOS and Android changes, and after this PR gets reviewed, would you be open to merging this PR? @Titozzz mentioned a new architecture for react-native-webview. Would we need to wait for those changes to complete before merging this PR?

> Do you have any plans for this PR? Is there anything I can do to help move forward?

I'll get back to you on that after I talk with my team and after @TheAlmightyBob responds to the question above. I'd like this PR, eventually with both iOS and Android changes, to get merged into react-native-webview/react-native-webview.

apple/RNCWebView.m
donaldchen: At a high level, the `RNCWebView`'s lifecycle is one-to-one with the lifecycle of the corresponding react-component, and the `WKWebView`'s lifecycle is not one-to-one with the lifecycle of the corresponding react-component. `sharedWKWebViewDictionary` lets us keep the `WKWebView` alive when the first react-component unmounts and when its corresponding `RNCWebView` gets destroyed.

`sharedRNCWebViewDictionary` exists because it's possible to begin with one `RNCWebView` and one `WKWebView` for a given web view key. Then it's possible to mount a second react component for the webview with the same webview key before the first react component for the first `RNCWebView` unmounts. At this point, there will be two `RNCWebView`s with the same web view key but only one `WKWebView`. In `didMoveToWindow` for the second `RNCWebView`, we need to call 

```
RNCWebView *rncWebView = sharedRNCWebViewDictionary[_webViewKey];
if (rncWebView != nil) {
  [self removeWKWebViewFromSuperView:rncWebView];
}
```

and remove the `WKWebView` from the first `RNCWebView` before adding it as a child of the second `RNCWebView`. Otherwise, the app will crash.

cc @nealmanaktola

src/WebViewTypes.ts
donaldchen: As currently written, then in that scenario, the `WKWebView` will have the default behavior: its lifecycle will be roughly one to one with the lifecycle of the corresponding react component.

`keepWebViewInstanceAfterUnmount` needs to be true and `webViewKey` must be non-nullish for the `WKWebView` to stay alive after its corresponding react component unmounts. I wrote it this way to make the behavior more explicit. It's theoretically more flexible because then we can use the `webViewKey` for other cases too if necessary. I haven't actually thought of concrete examples that would need just the `webViewKey` though, so I'm open to changing this if that's simpler.

An alternative option is to remove `keepWebViewInstanceAfterUnmount` and then assume that if `webViewKey` is non-nullish, then we should keep the `WKWebView` alive after react-component-unmount.

nealmanaktola:
> if we were to get this PR updated with both iOS and Android changes, and after this PR gets reviewed, would you be open to merging this PR?

@TheAlmightyBob, would you prefer a PR per platform (one for iOS and one for Android) or a single PR with both iOS and Android changes?

donaldchen:
> Do you have any plans for this PR? Is there anything I can do to help move forward?
I'll get back to you on that after I talk with my team and after TheAlmightyBob responds to the question above.

@cou929 , for now, you can use Discord's fork of react-native-webview if you'd like to use these features on both iOS and Android https://github.com/discord/react-native-webview/commits/11.18.1-discord-1 . Later, depending on the response to https://github.com/react-native-webview/react-native-webview/pull/2469#issuecomment-1238404004 , we can hopefully get these iOS and Android changes merged into react-native-webview/react-native-webview.

Titozzz:
You can check my work for new arch the branch is published already and 95% of features are good

apple/RNCWebView.m
TheAlmightyBob: Oh that's interesting. Thanks for the explanation.

So, if you were you just put two WebViews on a screen with the same key, I'm guessing whichever one mounts last will "win" and the other will just be blank?

src/WebViewTypes.ts
TheAlmightyBob: The alternative option is what I was originally thinking, but being explicit isn't a bad idea as long as it's documented that you need to use both.

The other use of `webViewKey` that I've been thinking of (and mentioned in #2640) would be to leverage this for supporting popup windows.... e.g., if web content calls `window.open`, the native "create window" handler could potentially create a new `WKWebView` in the dictionary and raise an event with the (generated?) key, which the app could then use (in RN code) to render a new WebView with that key. _Theoretically_, this could mean that the two WebViews would be connected as they would be in a browser.

I haven't fully thought through or tested the idea though, just mentioning as "_maybe_ this is a scenario where you would set `webViewKey` but not `keepWebViewInstanceAfterUnmount`...."

TheAlmightyBob:
> would you be open to merging this PR?

Yes, but...

> @Titozzz mentioned a new architecture for react-native-webview. Would we need to wait for those changes to complete before merging this PR?

...I haven't been involved with this work and don't know the scope/timeline. I guess I'd encourage checking out his branch as he mentions above?

> would you prefer a PR per platform (one for iOS and one for Android) or a single PR with both iOS and Android changes?

Although I expect it'll be very large, I think I'd lean towards a single PR. But if there's a logical way to organize the commits for reviewing incrementally, that might help.

apple/RNCWebView.m
donaldchen: > So, if you were you just put two WebViews on a screen with the same key, I'm guessing whichever one mounts last will "win" and the other will just be blank?

Yeah, that's what I expect to happen. Ideally, the react / JS part of the app will typically be written in a way such that there is only one web view component on screen at a time for a given key. However, I've run into this scenario with two webview components with the same key in my earlier testing, and I felt that blanking was better than crashing

donaldchen:
@Titozzz , do you have a rough estimate on when https://github.com/react-native-webview/react-native-webview/commits/new-arch will merge to master?

@nealmanaktola and I will check out that branch and see if there are any major conflicts with out changes. Depending on when that new-arch branch is expected to merge, @nealmanaktola and I could either make our PR against `master` or against `new-arch`. @TheAlmightyBob , do you have any preferences on which branch we should make the PR against? If we were to make the PR against `master`, would you prefer us updating everything on this PR, or would making a completely new PR be easier?

> Although I expect it'll be very large, I think I'd lean towards a single PR. But if there's a logical way to organize the commits for reviewing incrementally, that might help.

Sounds good! @nealmanaktola and I will initially plan on creating a single PR for both iOS and Android.

Titozzz:
I need to understand what to do with new iOS lifecycle on Fabric, it's actually more about deciding than code left

src/WebViewTypes.ts
nealmanaktola: @TheAlmightyBob I think that would be a good use for the webViewKey. However, @donaldchen and I believe that for now it might just make sense to keep `webViewKey` alone and remove `keepWebViewInstanceAfterUnmount`. We can reintroduce that property at a later point when needed (perhaps, when supporting pop-up windows).

TheAlmightyBob:
> Depending on when that new-arch branch is expected to merge, @nealmanaktola and I could either make our PR against master or against new-arch. @TheAlmightyBob , do you have any preferences on which branch we should make the PR against? 

Not a strong one. I'm not sure how "stable" the `new-arch` branch is right now, whether you would have trouble rebasing on top of that. I also don't know how realistic the possibility is of this PR getting merged _first_.

> If we were to make the PR against master, would you prefer us updating everything on this PR, or would making a completely new PR be easier?

Er, did you mean to say if you were to make it against `new-arch`? It's already against `master`...?

In general I think it depends on the nature/scope of the changes. If you're just pushing new commits, then updating this PR seems to make sense since there's already been some discussion here. If you're basically redoing everything (which might be essentially what happens if you target `new-arch`?), or other such drastic changes, then a new PR might make sense.

cou929:
@donaldchen 

> for now, you can use Discord's fork of react-native-webview if you'd like to use these features on both iOS and Android https://github.com/discord/react-native-webview/commits/11.18.1-discord-1 . Later, depending on the response to [#2469 (comment)](https://github.com/react-native-webview/react-native-webview/pull/2469#issuecomment-1238404004) , @nealmanaktola or I can take care of updating this PR, so we can hopefully get the iOS / Android changes merged into react-native-webview/react-native-webview.

That’s amazing. Thank you so much!

donaldchen:
> Er, did you mean to say if you were to make it against new-arch? It's already against master...?
In general I think it depends on the nature/scope of the changes. If you're just pushing new commits, then updating this PR seems to make sense since there's already been some discussion here. If you're basically redoing everything (which might be essentially what happens if you target new-arch?), or other such drastic changes, then a new PR might make sense.

@TheAlmightyBob , I was asking if it'd be easier to create a brand new PR against master, but since we're not going to redo everything in this PR, @nealmanaktola and I will tentatively plan on updating this PR (still against master) with our new Android changes and some small adjustments on iOS. We can revisit creating a PR against new-arch some time in the future - perhaps after the new-arch and its timeline for merging is more stable

apple/RNCWebView.m
donaldchen: I'm in the process of updating the PR description, and I came across this other reason, which I had documented in an earlier version of the PR description, for having two maps: one for `WKWebView`s and one for `RNCWebView`s.

>  We also have a map of web view keys  to `RNCWebView`s, so we can remove the correct observer from the `WKWebView`. If we don't remove the observer when we change the `WKWebView`'s parent, the app will crash with an `EXC_BAD_ACCESS` error.

This may have been one source of crashes before we added that RNCWebView map.

cc @nealmanaktola , since we previously discussed whether we needed the two maps on iOS

donaldchen:
@Titozzz  and @TheAlmightyBob , @nealmanaktola  and I have updated the PR title, description, and commits with our updated iOS and Android changes. This PR is ready for a full review, and we're hoping to merge this to `master` after we've addressed any requested changes you might have. The PR description has a "High level react-native-webview API changes" section, and if the high level API looks good to you, @nealmanaktola or I can update this PR with simple docs for these new APIs under https://github.com/react-native-webview/react-native-webview/tree/master/docs

donaldchen:
Screen Shot 2022-09-13 at 2 04 25 PM

Whoops, I didn't mean to remove anyone as a reviewer, and I can't seem to explicitly add anyone as a reviewer. @Titozzz , we'd like your review as well if you're available. Thanks!

TheAlmightyBob:
> We also need to be able to pass messages between the WebView and the react-native app without a react tag (e.g. when the React component for the WebView is not mounted)

> When the webViewKey is defined, onMessage and similar callbacks like onLoadStarted don't work

Could you explain a bit more about this scenario?

nealmanaktola:
> Could you explain a bit more about this scenario?

@TheAlmightyBob 

To clarify, the way it's currently built is that any message event gets published to the `EventEmitter` when a `webViewKey` is set, so instead of messages being sent to the regular bound onMessage fn, it gets sent via the `EventEmitter` 

like here: https://github.com/react-native-webview/react-native-webview/pull/2469/files#diff-e9700631fc9fa5600d2e1ee70eb2d28aa2a1fbc695b85f99981972d3904ca485R1816

We added `addOnMessageListenerWithWebViewKey` because with these changes, it's now possible for messages to be dropped when listening to messages via the `onMessage` prop fn.  Message drops happen when the react component is unmounted but the WebView instance is still alive. 

Now with respect to onLoadStarted and other callbacks, those callbacks will not trigger when the react component is unmounted. Since those callbacks rely on the viewId/viewTag, they will not trigger when the react component is unmounted.

When the WebView react component is mounted, these other callbacks do trigger.

.eslintrc.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
   plugins: ['react', 'react-native', 'import', '@typescript-eslint'],
   rules: {
     'no-console': 'off',
+    'no-unused-vars': 'off',
```
TheAlmightyBob, line 8: Why this change?

.eslintrc.js
```diff
@@ -61,7 +62,10 @@ module.exports = {
     ],
     '@typescript-eslint/no-unused-vars': [
       'error',
-      { ignoreRestSiblings: true },
+      { 
+        ignoreRestSiblings: true,
+        argsIgnorePattern: "^_"
```
TheAlmightyBob, line 67: Why this change?

TheAlmightyBob:
@nealmanaktola Thanks for the explanation.

> When the WebView react component is mounted, these other callbacks do trigger.

Does that mean that this line from the description is slightly inaccurate, and it should just say that the callbacks don't work _when the component is unmounted_?

> When the webViewKey is defined, onMessage and similar callbacks like onLoadStarted don't work


I'm still wrapping my head around things, but I'm kind of curious if it could make sense to have a new kind of "proxy" object instead of these separate `___WithWebViewKey` functions.... like, something that could sort of mimic a webViewRef (not necessarily the full API, but in a more familiar way)...?

I haven't fully thought this through, but I'm _vaguely_ imagining being able to do something like

```typescript
const webViewProxy = getWebViewProxy('key', { onMessage: this.handleMessage, onLoadStart: this.handleLoadStart });
...
webViewProxy.injectJavascript('...');
...
webViewProxy.release();
```

I'm a little concerned about the `src` folder ballooning with tons of little files for each of these methods/events and it becoming difficult to keep them straight with their "normal" counterparts... I at least appreciate the problem you're solving now, but trying to imagine how best to keep this manageable going forward...

nealmanaktola:
> Does that mean that this line from the description is slightly inaccurate, and it should just say that the callbacks don't work when the component is unmounted?

Yeah that line from the description is slightly inaccurate. We'll update that. Good call. `onMessage` does not work at all (even when the component is mounted or unmounted) when WebView has a webViewKey. However, the other callbacks do indeed work when the component is mounted.

> I'm still wrapping my head around things, but I'm kind of curious if it could make sense to have a new kind of "proxy" object instead of these separate ___WithWebViewKey functions.... like, something that could sort of mimic a webViewRef (not necessarily the full API, but in a more familiar way)...?

Agreed, I think right now the `withWebViewKey` paradigm doesn't really scale that well. A single proxy object does make sense to me and provides some familiarity as with a webViewRef as you noted. In terms of implementation, I am thinking that `WebViewProxy` would be a wrapper object over these functions that we've already defined. 

I have to think a bit more on how this would actually work though.

.eslintrc.js
nealmanaktola: I figured we could turn this off because `typescript-eslint/no-unused-vars` already checks for this.

.eslintrc.js
nealmanaktola: For some of our implementations (i.e macOS, windows), the functions are no-op. The function parameters are unused, so wanted a way to avoid the lint error.

For example, over here: https://github.com/react-native-webview/react-native-webview/blob/ac1b17eaec16af20523d76adf771a4adec2c303d/src/injectJavaScriptWithWebViewKey.macos.ts#L1

`_webViewKey` is not used

TheAlmightyBob:
@nealmanaktola 

> onMessage does not work at all (even when the component is mounted or unmounted) when WebView has a webViewKey

_could it_? i.e., could it still trigger the regular `onMessage` if that exists? Is there a technical reason, or is it more just that you want to steer people to use `addOnMessageListenerWithWebViewKey` if they're using `webViewKey`?

nealmanaktola:
@TheAlmightyBob it definitely could. It's more that we wanted to steer people to use `addOnMessageListenerWithWebViewKey` if they're using a `webViewKey`

.eslintrc.js
TheAlmightyBob: I see.

1. Do you need the `.macos`/`.windows` versions? I'd expect the generic `injectJavaSCriptWithWebViewKey.ts` to be used in those cases.
2. Is it insufficient to use `// eslint-disable-next-line @typescript-eslint/no-unused-vars, no-unused-vars` above the function declaration?

(I don't strongly object to this change, I'm just looking for any opportunities to narrow the scope of what is a very large PR 🙂 )

nealmanaktola: > Do you need the .macos/.windows versions? I'd expect the generic injectJavaScriptWithWebViewKey.ts to be used in those cases.

Ah that's true. I can remove those files and remove this change. 

> Is it insufficient to use // eslint-disable-next-line @typescript-eslint/no-unused-vars, no-unused-vars above the function declaration? (or I guess just the typescript-eslint one if turning off the other)

We can definitely use those, I figured that there may be a case in the future where you'd want to indicate an unused var without cluttering the code with `// es-lint-disable-next-line`.  

I'll remove these changes to keep the changes smaller for now.

lazaronixon:
Hi guys, awesome work, I’m waiting it for a long time. Do you think if I could use this implementation to create a custom navigation where each link opens in a new screen with a single webview instance ?

lazaronixon:
Do you think there's a way to move the webview instance back ?
https://user-images.githubusercontent.com/2651240/191662096-0f1ee169-0f27-44ef-a9fc-d2f8f6b237d5.mov

donaldchen:
@lazaronixon , to clarify, are you asking if the changes in this PR will allow you to use the same native WebView instance in different screens within a react-navigation Navigator both when opening new links and when going back? If so, then I think that should be possible with these changes.

For opening new links, when the newest link / screen mounts its React WebView component with a given WebView key, it'll reuse the native WebView. For going back, currently the previous screens will only reuse the native WebView if their React WebView components remount. With the current API, you might be able to use https://reactnavigation.org/docs/use-focus-effect/ and remount the React WebView component when the screen refocuses.

@nealmanaktola  and I have also discussed the possibility of adding a `rebind` method for using the shared native WebView with a React WebView component that is already mounted (without needing to remount the react component). That API isn't ready in this PR though

nealmanaktola:
@TheAlmightyBob 

Would you like to see the `WebViewProxy` changes described in https://github.com/react-native-webview/react-native-webview/pull/2469#issuecomment-1247357261 in this PR? Trying to determine if this is a suggested improvement that could happen in a separate PR or if this is a blocking requested change for this PR.

TheAlmightyBob:
@nealmanaktola I still need to get through the rest of this PR, but my main concern with making the Proxy a separate PR is that it might be a breaking change for anybody who starts using this.

Otherwise, or if you think the functions you've already defined would need to stay anyway, I wonder if they could be moved to a subfolder? _Possibly_ exported as a module?

Just pondering how to avoid filling up the `src` folder with lots of little files for each of these functions, as well as perhaps aiding discoverability...

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -0,0 +1,93 @@
...
+  @Nullable
+  public RNCWebViewManager.InternalWebView getWebView() {
+    return internalWebView;
+  }
+
+  /**
+   * Applies an action if the internal webview is non null
+   * @param action
+   */
+  public void ifHasInternalWebView(Action action) {
```
TheAlmightyBob, line 72: nit: This needs a verb. Perhaps `doIfHasInternalWebView`? Or `applyToInternalWebView`?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -272,33 +272,65 @@ private String getLackPermissionToDownloadMessage() {
...
+
+    // If there is an existing RNCWebView that has an internal webview, re-attach it to this view
+    if (rncWebViewMap.containsKey(webViewKey)) {
+      RNCWebView existingView = rncWebViewMap.get(webViewKey);
+      InternalWebView existingWebView = existingView.detachWebView();
+      view.attachWebView(existingWebView);
+
+      // The chrome client was originally setup on instance creation but might be pointing to the wrong webview
+      // so it's reset here.
+      // Not entirely sure why there is a single instance of the webchrome client for all webviews?
```
TheAlmightyBob, line 307: I don't think there is. I assume you're referring to `RNCWebViewManager` keeping its own copy of `mWebChromeClient`, but as far as I can tell the only thing that's used for is to call `onHideCustomView` on the "previous" instance when creating a new instance...? But it looks to me like a new instance is created (sometimes multiple times) for each WebView...?

I'm not familiar with the `onHideCustomView` logic.... I'm _guessing_ that implementation just never considered the possibility of having more than one WebView...

TheAlmightyBob (changes requested):
Alright, I've only barely scratched the surface of Android, but it's a start. This is a big PR!

(I know, you warned me, but it's still bigger than I expected)

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -0,0 +1,93 @@
...
+
+import android.webkit.WebView;
+import android.widget.FrameLayout;
+
+import androidx.annotation.Nullable;
+
+import com.facebook.common.logging.FLog;
+import com.facebook.react.uimanager.ThemedReactContext;
+
+public class RNCWebView extends FrameLayout {
```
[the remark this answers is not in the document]
nealmanaktola: > It would be nice to have comments on this and InternalWebView to describe their unique roles, as that seems like it will be a source of confusion.
Sounds good.

> I can't help wondering if InternalWebView should be kept as RNCWebView (because it's the thing that actually extends WebView?) and this should be called something like RNCWebViewContainer or RNCWebViewProxy...?

`RNCWebViewContainer`/ `RNCWebViewProxy` does make sense, the main reason for this name change is to be consistent with the iOS counterpart. iOS `RNCWebView` is a wrapper over `WKWebView`

> The new-arch branch from @Titozzz extracts all the classes out from RNCWebViewManager into separate files... if InternalWebView were not a nested class of RNCWebViewManager, would it still be called Internal

Yeah, I called it `Internal` mainly to differentiate it from `RNCWebView`. Internal may not be the best name here, but let me know if you have any ideas.

If we are okay to have `RNCWebView` on Android to have a different meaning than the `iOS` counterpart, I think `RNCWebViewContainer` is perfectly fine

src/addOnMessageListenerWithWebViewKey.ts
```diff
@@ -0,0 +1,17 @@
+import { NativeModules, Platform, EmitterSubscription, NativeEventEmitter } from "react-native";
+
+const scriptMessageEmitter = new NativeEventEmitter(
+  Platform.select({
+    ios: NativeModules.ScriptMessageEventEmitter,
+    default: null
+  })
```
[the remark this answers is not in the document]
nealmanaktola: This is probably not needed, we can simply specify `NativeModules.ScriptMEssageEventEmitter` on second thought. On Android, this event emitter is not defined (and it's not needed) so we can pass undefined for Android

apple/RNCWebView.m
```diff
@@ -1417,6 +1500,15 @@ - (void)requestFocus
 #endif // !TARGET_OS_OSX
 }
 
+- (void)releaseWebView
+{
+  if (_webViewKey != nil) {
```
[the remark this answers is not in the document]
nealmanaktola: I think we made the helper later at some point so not all usages were migrated over

src/WebViewTypes.ts
```diff
@@ -742,6 +742,30 @@ export interface IOSWebViewProps extends WebViewSharedProps {
...
    */
   onCustomMenuSelection?: (event: WebViewEvent) => void;
+
+  /**
+   * If two React components use the same
+   * key for the WebView, they will use the same native WebView instance.
+   * 
+   * When this is set, the native WebView will not get released when the React component
+   * unmounts. When a React component remounts, it can use a previous native WebView instance
+   * by using the same webViewKey prop that the previous React component used.
```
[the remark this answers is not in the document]
nealmanaktola: I haven't tested that just yet - will test out and see.

nealmanaktola:
>  I still need to get through the rest of this PR, but my main concern with making the Proxy a separate PR is that it might be a breaking change for anybody who starts using this.

That's a good point. I think given that some other callbacks won't be triggered when the view is unmounted. I think it might be cleaner to introduce this concept now.

I'll think about this and update the PR accordingly.

Thank you for the detailed review!

src/addOnMessageListenerWithWebViewKey.ts
```diff
@@ -0,0 +1,17 @@
...
+);
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export default function addOnMessageListenerWithWebViewKey(webViewKey: string, listener: (event: any) => void): EmitterSubscription {
```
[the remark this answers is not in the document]
nealmanaktola: Ah, this can be updated

apple/RNCWebView.h
```diff
@@ -101,18 +101,24 @@ shouldStartLoadForRequest:(NSMutableDictionary<NSString *, id> *_Nonnull)request
...
 - (void)goForward;
 - (void)goBack;
 - (void)reload;
 - (void)stopLoading;
 - (void)requestFocus;
+- (void)releaseWebView;
 #if !TARGET_OS_OSX
 - (void)addPullToRefreshControl;
 - (void)pullToRefresh:(UIRefreshControl *_Nonnull)refreshControl;
+- (void)cleanUpWebView;
```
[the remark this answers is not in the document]
donaldchen: > What's the expected MacOS support?

We're expecting zero MacOS support for the new APIs around reusing the WebView instance and sending messages to and from the WebView without a React tag. We expect existing APIs, including `removeFromSuperView`, to continue to work on MacOS.

> Why is cleanUpWebView (which already appears to check TARGET_OS_OSX in its implementation) in the !TARGET_OS_OSX block but releaseWebView (and others above) is not?

For context, you can reference our first iOS PR in our fork of react-native-webview: https://github.com/discord/react-native-webview/pull/1/files#diff-fee21637a62b4cc5dd3ff9992add07b843c999a733b03082d32cf35ef2f42bc3R529 . `removeFromSuperView` already had a `TARGET_OS_OSX` check. I pulled part of `removeFromSuperView` out into `cleanUpWebView` so that (a) `removeFromSuperView` could behave the same as before by calling `cleanUpWebView` when we're not reusing the WebView instance, and (b) we could separately call `cleanUpWebView` from `releaseWebView` when we're reusing the WebView instance.

src/addOnMessageListenerWithWebViewKey.ts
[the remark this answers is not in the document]
donaldchen: I think we may need a custom type because we also add the `webViewKey` to the event. https://github.com/discord/react-native-webview/pull/6/files#diff-c1064c73d2b45b9e3b6352aa2f27995b05ec16f5e1fd958dd2fa86070b842aa7R22

apple/RNCWebView.m
TheAlmightyBob: I just noticed that updated comment you added, which I appreciate.

I was wondering though, could you not check the `WKWebView`'s `superview` property for the first `RNCWebView` rather than looking it up in a dictionary?

apple/RNCWebView.h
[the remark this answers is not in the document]
TheAlmightyBob: Your reasoning for refactoring `removeFromSuperView`/`cleanUpWebView` makes sense. However, you say that you expect `removeFromSuperView` "to continue to work on MacOS", but it calls `cleanUpWebView`, which here in the `.h` is in the `#if !TARGET_OS_OSX` block. Is that not a problem?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -272,33 +272,65 @@ private String getLackPermissionToDownloadMessage() {
...
+      setupWebChromeClient((ThemedReactContext) existingWebView.getContext(), existingWebView);
+
+    // If there is a detached internal webview attach it to this RNCWebView
+    } else if (internalWebViewMap.containsKey(webViewKey)) {
+      InternalWebView webView = (InternalWebView) internalWebViewMap.get(webViewKey);
+      view.attachWebView(webView);
+    }
+
+    // Update all maps with the view + set/update key
+    // This means an existing webview can update it's own key
```
TheAlmightyBob, line 317: ultra-nit:
```suggestion
    // This means an existing webview can update its own key
```

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -341,34 +375,36 @@ public void setCacheMode(WebView view, String cacheModeString) {
...
+    final int layerType;
     switch (layerTypeString) {
         case "hardware":
           layerType = View.LAYER_TYPE_HARDWARE;
           break;
         case "software":
           layerType = View.LAYER_TYPE_SOFTWARE;
           break;
+      default:
+        layerType = View.LAYER_TYPE_NONE;
```
TheAlmightyBob, line 399: Just curious if there was a particular reason for this change?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -516,171 +556,189 @@ public void setIncognito(WebView view, boolean enabled) {
...
-  public void setSource(WebView view, @Nullable ReadableMap source) {
-    if (source != null) {
-      if (source.hasKey("html")) {
-        String html = source.getString("html");
-        String baseUrl = source.hasKey("baseUrl") ? source.getString("baseUrl") : "";
-        view.loadDataWithBaseURL(baseUrl, html, HTML_MIME_TYPE, HTML_ENCODING, null);
+  public void setSource(RNCWebView view, @Nullable ReadableMap source) {
+    view.ifHasInternalWebView(webView -> {
+
+      // Do not reload reload webview if the source prop has not changed
```
TheAlmightyBob, line 576: nit on "reload reload"

But, less-nitty, this could probably use some extra explanation about how it relates to WebView re-use (the comment here doesn't mention it, although the `if` statement checks for `webViewKey`)... I _assume_ the issue is that otherwise the WebView would reload when attached to a new container?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -723,74 +781,93 @@ Map<String, Integer> getCommandsMap() {
...
-        boolean includeDiskFiles = args != null && args.getBoolean(0);
-        root.clearCache(includeDiskFiles);
-        break;
-      case "clearHistory":
-        root.clearHistory();
-        break;
-    }
-    super.receiveCommand(root, commandId, args);
+  public void receiveCommand(@NonNull RNCWebView view, String commandId, @Nullable ReadableArray args) {
+    view.ifHasInternalWebView(root -> {
```
TheAlmightyBob, line 785: Why `root` here?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -723,74 +781,93 @@ Map<String, Integer> getCommandsMap() {
...
   @Override
-  public void onDropViewInstance(WebView webView) {
-    super.onDropViewInstance(webView);
-    ((ThemedReactContext) webView.getContext()).removeLifecycleEventListener((RNCWebView) webView);
-    ((RNCWebView) webView).cleanupCallbacksAndDestroy();
-    mWebChromeClient = null;
+  public void onDropViewInstance(RNCWebView view) {
+    super.onDropViewInstance(view);
+
+    // The internal webview can be null since the view may have been already reattached
```
TheAlmightyBob, line 855: Again, by "the view may have been already reattached," I assume you mean that the former internal WebView of this `view` may have been moved to a different `RNCWebView` container?

TheAlmightyBob: Also, doesn't `ifHasInternalWebView` already handle the possibility of the internal WebView being null?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -1551,6 +1635,35 @@ public void setNestedScrollEnabled(boolean nestedScrollEnabled) {
       this.nestedScrollEnabled = nestedScrollEnabled;
     }
 
+    public void setSource(ReadableMap source) {
+      this.source = source;
+    }
```
TheAlmightyBob, line 1640: Maybe worth a comment (either here or where `source` is declared) that this doesn't actually load the new source into the WebView, it's just for detecting changes?

Or perhaps worth considering if the logic for actually updating the WebView should move from `RNCWebViewManager`'s `setSource` to this function....?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -1551,6 +1635,35 @@ public void setNestedScrollEnabled(boolean nestedScrollEnabled) {
...
+
+    public boolean isNewSource(@Nullable ReadableMap newSource) {
+      if (source == null || newSource == null) {
+        return true;
+      }
+
+      // Check if any of the following string values have changed
+      String[] sourceKeys = {"uri", "method", "body", "html", "baseUrl"};
```
TheAlmightyBob, line 1648: Why not compare the entire maps?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -1693,10 +1810,17 @@ public void callInjectedJavaScriptBeforeContentLoaded() {
...
+      WebView webView = this;
+
+      if (webViewKey != null && mRNCWebViewClient != null) {
+        reactContext.runOnUiQueueThread(() -> {
+          WritableMap data = mRNCWebViewClient.createWebViewEvent(webView, webView.getUrl());
+          data.putString("webViewKey", webViewKey);
+          data.putString("data", message);
+          reactContext.getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class).emit("ReactNativeWebViewOnMessageWithWebViewKey", data);
+        });
+      } else if (mRNCWebViewClient != null) {
```
TheAlmightyBob, line 1823: nit: It looks like `onMessage` is basically a no-op if `mRNCWebViewClient` is `null`...? So perhaps instead of checking it in both the `if` and `else if` branches, there could be a higher-level check (e.g., just return if it's `null`)

TheAlmightyBob (changes requested):
Okay, finally looked more at the Android side

apple/RNCWebView.h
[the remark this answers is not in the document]
donaldchen: Ah that's a good point. `cleanUpWebView` should be safe to call whether `TARGET_OS_OSX` is true or not. In that case, I think we should move `cleanUpWebView` in `RNCWebView.h` to outside of the `#if !TARGET_OS_OSX` block. We can update that

apple/RNCWebView.m
```diff
@@ -406,18 +411,46 @@ - (void)didMoveToWindow
...
+    
+    if (!reusedWebViewInstance) {
 #if !TARGET_OS_OSX
-    _webView.scrollView.delegate = self;
+      _webView.scrollView.delegate = self;
 #endif // !TARGET_OS_OSX
-    _webView.UIDelegate = self;
-    _webView.navigationDelegate = self;
+      _webView.UIDelegate = self;
+      _webView.navigationDelegate = self;
```
TheAlmightyBob, line 453: This is curious... this means the previous `RNCWebView`'s delegates would still be in control of the `WKWebView` instance even after it's been reparented to a different `RNCWebView`? What's the intent of guarding this logic with `if (!reusedWebViewInstance)`?

nealmanaktola:
@TheAlmightyBob Thanks the additional passes, I haven't had time the past week to look through your additional feedback. I'll spend some time this week addressing it :)

apple/RNCWebView.m
donaldchen: That is a great question! You've identified a bug here, and I think we should remove the `if (!resusedWebViewInstance)` check. https://github.com/discord/react-native-webview/pull/21 cc @nealmanaktola 

This was likely a mistake from my first iteration on these changes. If I'm remembering correctly, I was trying to avoid repeating WebView-initialization logic when reusing a `WKWebView` instance with a new `RNCWebView` instance. However, I'm realizing this part of the initialization logic makes the `WKWebView` reference the `RNCWebView`, so it's important that it references the new `RNCWebView` rather than the old one

donaldchen: I ran into this issue today when I was trying to make a button-click in the web page conditionally load an iframe. Loading the iframe successfully depends on the navigation delegate as I learned from https://stackoverflow.com/questions/41250162/wkwebview-not-loading-content-from-iframe-on-a-webpage , and I realized the navigation delegate wasn't getting reconnected properly when attaching the `WKWebView` to a new `RNCWebView`

TheAlmightyBob:
@donaldchen @nealmanaktola A heads-up that, since there hasn't been recent activity on this PR, we're going to prioritize getting the new architecture support in first. Still fans of this idea though!

nealmanaktola:
@TheAlmightyBob makes sense - haven't had time recently to make these changes, but will get to it.

TheAlmightyBob:
Hey bot, let's keep this one open

apple/RNCWebView.m
donaldchen: Fixed here https://github.com/discord/react-native-webview/commit/a4322df5cdbc0ad63dcb983ece35306666c61c7e

apple/RNCWebView.m
donaldchen: @TheAlmightyBob , you're right! We should be able to check the `WKWebView`'s `superview` property for the first `RNCWebView`. In these two new commits
- https://github.com/react-native-webview/react-native-webview/pull/2469/commits/d1e443333c4300eee810eb467b65d526c02dc085
- https://github.com/react-native-webview/react-native-webview/pull/2469/commits/267b40a3d2bc011d64116302e9020671c5b8acc5

, I've updated our iOS and Android implementations to (a) remove the map of RNCWebViews and (b) get the RNCWebView through the internal WebView's parent before detaching the internal WebView from its parent

src/WebViewTypes.ts
donaldchen: It looks like Neal addressed this in https://github.com/react-native-webview/react-native-webview/pull/2469/commits/63c586af927c00e1005c754dc36e211b7f1907c3 , so I'm going to resolve this comment thread

.eslintrc.js
donaldchen: I removed the `.eslintrc.js` change here https://github.com/react-native-webview/react-native-webview/pull/2469/commits/4343e47dcc0ca479cb56da74d76c7f904e6131a2 . In a previous commit, I had replaced the `injectJavaScriptWithWebViewKey` files with `WebViewProxy` files. I've also removed the `.macos.ts` and `.windows.ts` files for `WebViewProxy`

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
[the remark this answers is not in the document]
donaldchen: Addressed here: https://github.com/react-native-webview/react-native-webview/pull/2469/commits/2eeb82f1377a4ae3f12cfb3ffad7b0f40e594f8c .

For now, I've renamed the outer WebView container to `RNCWebViewContainer`, and I've renamed the internal WebView (the one that extends the Android `WebView`) to `RNCWebView`.  This will make it slightly confusing for anyone working on both iOS and Android for this library, since now the iOS "RNCWebView" corresponds to the Android "RNCWebViewContainer" while the iOS "WKWebView" corresponds to the Android "RNCWebView". I'm okay with that if the inconsistency allows us to create better backward compatibility for the customization use case though.

Thanks for calling out that use case! Let me know if I missed anything

apple/RNCWebView.m
```diff
@@ -406,18 +411,46 @@ - (void)didMoveToWindow
 {
   if (self.window != nil && _webView == nil) {
     WKWebViewConfiguration *wkWebViewConfig = [self setUpWkWebViewConfig];
+    NSMutableDictionary *sharedWKWebViewDictionary= [[RNCWKWebViewMapManager sharedManager] sharedWKWebViewDictionary];
```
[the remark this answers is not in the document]
donaldchen: Addressed here: https://github.com/react-native-webview/react-native-webview/pull/2469/commits/e9a0e44adcb3c95a92a3c9bc2096573ee47a1e67

apple/RNCWebView.m
```diff
@@ -406,18 +411,46 @@ - (void)didMoveToWindow
...
+      
+    if ([self shouldReuseWebView]) {
+      WKWebView *webViewForKey = sharedWKWebViewDictionary[_webViewKey];
+      if (webViewForKey != nil) {
+        _webView = webViewForKey;
+        NSMutableDictionary *sharedRNCWebViewDictionary= [[RNCWebViewMapManager sharedManager] sharedRNCWebViewDictionary];
```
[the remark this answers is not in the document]
donaldchen: I ended up removing `sharedRNCWebViewDictionary`

apple/RNCWebView.h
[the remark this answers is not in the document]
donaldchen: Addressed here: https://github.com/react-native-webview/react-native-webview/pull/2469/commits/6a1ef637cbc96b27b1c68dd4e9c9813fbb20cb6f

apple/RNCWebView.m
```diff
@@ -406,18 +411,46 @@ - (void)didMoveToWindow
...
-    _webView = [[WKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
+      _webView = [[WKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
 #else
-    _webView = [[RNCWKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
+      _webView = [[RNCWKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
 #endif // !TARGET_OS_OSX
+    }
+      
+    if (_webViewKey != nil) {
+        NSMutableDictionary *sharedRNCWebViewDictionary= [[RNCWebViewMapManager sharedManager] sharedRNCWebViewDictionary];
```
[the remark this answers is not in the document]
donaldchen: I ended up deleting `sharedRNCWebViewDictionary`

apple/RNCWebView.m
```diff
@@ -406,18 +411,46 @@ - (void)didMoveToWindow
...
 #if !TARGET_OS_OSX
-    _webView = [[WKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
+      _webView = [[WKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
 #else
-    _webView = [[RNCWKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
+      _webView = [[RNCWKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
 #endif // !TARGET_OS_OSX
+    }
+      
+    if (_webViewKey != nil) {
```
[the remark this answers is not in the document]
donaldchen: addressed here: https://github.com/react-native-webview/react-native-webview/pull/2469/commits/829967139d6ee111a8269f0d6e3e7a5fa68b60b6

apple/RNCWebView.m
```diff
@@ -472,12 +513,32 @@ - (void)setAllowsBackForwardNavigationGestures:(BOOL)allowsBackForwardNavigation
 }
 
 - (void)removeFromSuperview
+{
+  bool keepWebViewInstance = _webViewKey != nil;
+  if (!keepWebViewInstance) {
```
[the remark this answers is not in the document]
donaldchen: addressed here: https://github.com/react-native-webview/react-native-webview/pull/2469/commits/829967139d6ee111a8269f0d6e3e7a5fa68b60b6 . We're now using `shouldReuseWebView`

apple/RNCWebView.m
```diff
@@ -472,12 +513,32 @@ - (void)setAllowsBackForwardNavigationGestures:(BOOL)allowsBackForwardNavigation
...
+    [self cleanUpWebView];
+  }
+  
+  if (_webViewKey != nil) {
```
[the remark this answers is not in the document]
donaldchen: I made it an if-else, since one of them is checking for `shouldReuseWebView`, and the other is checking for `!shouldReuseWebView`. https://github.com/react-native-webview/react-native-webview/commit/829967139d6ee111a8269f0d6e3e7a5fa68b60b6

donaldchen:
@TheAlmightyBob , I addressed some of the initial feedback this week. Here's some of the high level feedback I addressed in the below commits.
- Added the WebViewProxy to encapsulate all the `***WithWebViewKey` methods
- Renamed the Android container to RNCWebViewContainer and renamed the Android WebView to RNCWebView for backward compatibility (particularly around the customization use case)
- Removed the maps referencing the outer WebView containers. (We can now get the parent or superview of the internal WebView).

Screen Shot 2023-03-29 at 10 23 53 AM

I hope to address the other comments at another time. For now, are you available to review the commits above, and can you clarify if there's any other high level feedback that I'd need to address in this PR before this PR can get merged? Thanks!

flaviut:
.

apple/RNCWebView.m
TheAlmightyBob: Thanks so much for linking the commits! I see the map still exists on Android, it's just used in fewer places. Was keeping it intentional, or is that change just not finished yet?

TheAlmightyBob:
Hey @donaldchen , sorry I completely missed that there'd been an update here until that bot brought my attention to it (I guess it's good for something after all...)

Have you by chance looked at https://github.com/react-native-webview/react-native-webview/pull/2874 at all? That introduces an `RNCWebViewWrapper` on Android that subclasses `FrameLayout`, like your `RNCWebViewContainer`. It has different reasons for doing so, of course, but hopefully the approaches are compatible?

src/WebViewTypes.ts
TheAlmightyBob: @donaldchen Er, does that mean you *don't* want me to resolve threads as I review changes...?

apple/RNCWebViewManager.m
```diff
@@ -286,4 +313,21 @@ - (BOOL)          webView:(RNCWebView *)webView
   }
 }
 
+RCT_EXPORT_METHOD(releaseWebView:(nonnull NSString *)webViewKey)
+{
+  [self.bridge.uiManager addUIBlock:^(__unused RCTUIManager *uiManager, __unused NSDictionary<NSNumber *, RNCWebView *> *viewRegistry) {
```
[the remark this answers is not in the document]
TheAlmightyBob: > Is https://github.com/facebook/react-native/blob/d9f7c4c189cf6d2ccd4b6641599b5e46898f7d0d/packages/react-native/React/Fabric/Mounting/RCTComponentViewRegistry.h the view registry you’re referring to?

Well, I believe I was specifically referring to the `viewRegistry` dictionary that's passed into this function via `uiManager addUIBlock` and can return a native view via `viewRegistry[reactTag]` (example [here](https://reactnative.dev/docs/native-components-ios#handling-multiple-native-views))... but I think those are related?

> I haven’t yet found an equivalent of the releaseWebView API added in this PR for releasing the native view

I thought "releasing" was specific to the `WKWebView`, not the outer `RNCWebView`? I think I had only been suggesting the `viewRegistry` as an alternative to `sharedRNCWebViewDictionary`, not `sharedWKWebViewDictionary`

> a view that would temporarily be the parent of the native WebView

Interesting. That doesn't cause side-effects?


Well, _anyhow_, curiosities aside, this suggestion may not be so relevant now that you've started using `superview` instead of `sharedRNCWebViewDictionary`, and also relying on stuff like the `viewRegistry` may not be a good idea with the new architecture (though @Titozzz is the expert there, not me).

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewContainer.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * Provides the associated parent RNCWebView viewId for the provided
+   * webView view id.
+   * @param webView
+   * @return viewId
+   */
+  public static int getRNCWebViewId(WebView webView) {
+    Integer rncViewId = RNCWebViewMapManager.INSTANCE.getViewIdMap().get(webView.getId());
```
TheAlmightyBob, line 87: This can't use `getParent`?

TheAlmightyBob (changes requested):
Hey @donaldchen , the last batch of changes look good to me, my comments are mostly small things.

The two biggest "high-level" concerns on my mind (neither of which I think we will be news to you) are:
- The new architecture. The big refactor has been completed/released, so this will need to be made compatible (e.g. most of the Android implementation has been converted to Kotlin). I don't _expect_ any major issues/incompatibilities, but not 100% certain
- Documentation/commenting. This definitely introduces complexity, and we want to make it as easy as possible both to _use_ and to _maintain_

donaldchen:
> Hey @donaldchen , the last batch of changes look good to me, my comments are mostly small things.
The two biggest "high-level" concerns on my mind (neither of which I think we will be news to you) are:
The new architecture. The big refactor has been completed/released, so this will need to be made compatible (e.g. most of the Android implementation has been converted to Kotlin). I don't expect any major issues/incompatibilities, but not 100% certain
Documentation/commenting. This definitely introduces complexity, and we want to make it as easy as possible both to use and to maintain

Hey @TheAlmightyBob , thanks for the feedback! That makes sense. I'd like to update this PR to work with the new architecture, and I'll plan on adding more documentation, too. My availability is still TBD, but I'm hoping I can come back to this in one to two months. We'll tag you when this is ready for another review. Thanks!

DavidGOrtega:
@donaldchen I have tried to use your branch, however the webview shows a blank page. It does not load for me.

i have added 

```js
<WebView
        webViewKey="web1"
        messagingWithWebViewKeyEnabled={true}

        ref={webviewEl}

```

My code also postMessages

```js
 const js = `
window.WebViewBridge = {
  onMessage: (message) => {
    const { type, data } = JSON.parse(message || '{}');
    document.dispatchEvent(new CustomEvent(type, { detail: data }));
  }
}
```

TheAlmightyBob:
@DavidGOrtega Does it work if you use it "normally" (without the webviewkey stuff)?

I haven't used the branch with the latest changes, but a previous version was working for me.

donaldchen:
@DavidGOrtega , I haven't been available to test this branch recently, but I hope to test more when I get around to working on making this work with the new react native architecture https://github.com/react-native-webview/react-native-webview/pull/2469#issuecomment-1572325629 .

Like @TheAlmightyBob suggested, you could try an older commit such as https://github.com/react-native-webview/react-native-webview/pull/2469/commits/413cae346d84b4be0e3f3e93a1c4d5aa2586d9d5 . That one was before my last batch of changes in March.

Also, what are you passing in for the `source` prop? If you pass in what the `Portals.tsx` example uses, does it work?
https://github.com/discord/react-native-webview/blob/237796aced1f986a5f8c33d5be295c5fbf74e728/example/examples/Portals.tsx#L160

https://github.com/discord/react-native-webview/blob/237796aced1f986a5f8c33d5be295c5fbf74e728/example/examples/Portals.tsx#L82-L103

> My code also post messages
 const js = `
window.WebViewBridge = {
  onMessage: (message) => {
    const { type, data } = JSON.parse(message || '{}');
    document.dispatchEvent(new CustomEvent(type, { detail: data }));
  }
}
any hint?

Can you share more context on where this code is running? Is this code in a react-native-desktop app using a WebView, or is this code that runs inside the WebView?

You can view https://github.com/discord/react-native-webview/blob/237796aced1f986a5f8c33d5be295c5fbf74e728/example/examples/Portals.tsx#L215 for an example of posting messages to the WebView and https://github.com/discord/react-native-webview/blob/237796aced1f986a5f8c33d5be295c5fbf74e728/example/examples/Portals.tsx#L112 for an example of listening to messages sent from the WebView back to the react-native app that hosts the WebView

DavidGOrtega:
Hi! 

> @DavidGOrtega Does it work if you use it "normally" (without the webviewkey stuff)?

Yes

> If you pass in what the Portals.tsx example uses, does it work?

Let me try it.


The issue is more puzzling than I thought. Sometime loads the page, sometimes does not. For sure what is happening also is that If it loads and the device get a locked screen if I go back the blank screen is there.
Have I determined how the white screen appears at the beginning? Sort of. 
My app requests lots of permissions. Using this webview if it does not ask for all the permissions... it works, if it takes longer to reach the point where the UI should display it does not.

This is tested in android 11 within a real device

donaldchen:
Small update from my end: I had begun updating this to support the new arch in https://github.com/discord/react-native-webview/pull/27 , but I haven't had time to get those changes fully working yet. I hope to revisit those changes again some time in the future

donaldchen:
> Hello 👋, this PR has been opened for more than 2 months with no activity on it. If you think this is a mistake please comment and ping a maintainer to get this merged ASAP!

@TheAlmightyBob , I'd like to keep this PR open if possible. I'm still hoping to get back to this in the future, but I don't have a specific timeline yet

Darren120:
can this please be merged

Darren120:
> > Hello 👋, this PR has been opened for more than 2 months with no activity on it. If you think this is a mistake please comment and ping a maintainer to get this merged ASAP!
> 
> @TheAlmightyBob , I'd like to keep this PR open if possible. I'm still hoping to get back to this in the future, but I don't have a specific timeline yet

i tried this PR, it crashes if transporting into a modal or another screen

Darren120:
> > i tried this PR, it crashes if keying it into a modal, or another screen, then unmounting that screen
> 
> @Darren120 , this branch https://github.com/discord/react-native-webview/commits/11.18.1-discord-1 has a more stable version of this code. For the branch in this PR (https://github.com/discord/react-native-webview/tree/11.18.1-discord-2), I haven't had time to resolve all the errors and issues yet when making it work with the new react-native architecture

Thank you for your work. I will try that branch.
but on the previous branch i have a question
is it necessary to use global context and pass the webview memo around? Or can I just mount a new webview with an existing webview key?

I am trying to reuse webview in a new screen/ modal. but if i exit the screen where the original webview is initialized, it crashes the app.

Darren120:
> > i tried this PR, it crashes if keying it into a modal, or another screen, then unmounting that screen
> 
> @Darren120 , this branch https://github.com/discord/react-native-webview/commits/11.18.1-discord-1 has a more stable version of this code. For the branch in this PR (https://github.com/discord/react-native-webview/tree/11.18.1-discord-2), I haven't had time to resolve all the errors and issues yet when making it work with the new react-native architecture

hm i am using that version, but it seems like it keeps crashing when unmounting.

Screen A => Screen B (initilize web view) => Screen C (shared webview key)

Screen B to C navigation is no problem, once i navigate to screen A, it crashes

Darren120:
> > i tried this PR, it crashes if keying it into a modal, or another screen, then unmounting that screen
> 
> @Darren120 , this branch https://github.com/discord/react-native-webview/commits/11.18.1-discord-1 has a more stable version of this code. For the branch in this PR (https://github.com/discord/react-native-webview/tree/11.18.1-discord-2), I haven't had time to resolve all the errors and issues yet when making it work with the new react-native architecture

UPDATE:
haha sorry about that, i forgot to prebuild my expo while jumping around branches. no more crashes! excellent work! i hope we can get this merged!

Darren120:
> > i tried this PR, it crashes if keying it into a modal, or another screen, then unmounting that screen
> 
> @Darren120 , this branch https://github.com/discord/react-native-webview/commits/11.18.1-discord-1 has a more stable version of this code. For the branch in this PR (https://github.com/discord/react-native-webview/tree/11.18.1-discord-2), I haven't had time to resolve all the errors and issues yet when making it work with the new react-native architecture

hello, i am wonder what happened to 
```injectJavaScriptWithWebViewKey```? is there an alternative to this in the stable version?

Darren120:
> > i tried this PR, it crashes if keying it into a modal, or another screen, then unmounting that screen
> 
> @Darren120 , this branch https://github.com/discord/react-native-webview/commits/11.18.1-discord-1 has a more stable version of this code. For the branch in this PR (https://github.com/discord/react-native-webview/tree/11.18.1-discord-2), I haven't had time to resolve all the errors and issues yet when making it work with the new react-native architecture

An issue I am encountering is cookies are either not being shared if another webview sets them, or its being resetted
I think it broke sharedCookiesEnabled

Darren120:
> > i tried this PR, it crashes if keying it into a modal, or another screen, then unmounting that screen
> 
> @Darren120 , this branch https://github.com/discord/react-native-webview/commits/11.18.1-discord-1 has a more stable version of this code. For the branch in this PR (https://github.com/discord/react-native-webview/tree/11.18.1-discord-2), I haven't had time to resolve all the errors and issues yet when making it work with the new react-native architecture

further debugging, seems like the cookies are not being synced if a web view is activity being viewed while portaled.
for example, i am showing the web view in another screen by portaling it there, but after a while my auth cookies are not synced. if i stay on the original screen, the cookie remains synced, and navigating to the portaled webview works fine. so i guess while webview is portaled and activity being viewed, it is not syncing cookies

Titozzz:
FYI: I still like the idea behind that PR, I'm thinking of rewriting webview around new arch with this capability

Issue #2436, opened 2022-04-05
Android "Downloading" text is not localized


TheAlmightyBob:
**Bug description:**

The "Downloading" toast that appears when downloading a file on Android is [hardcoded to English](https://github.com/react-native-webview/react-native-webview/blob/5e73b2089fc80c2be7aa6eff291b18c81ad4030d/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java#L311).

**To Reproduce:**

Change the Android device locale to non-English and download a file.

**Expected behavior:**

The toast that appears either _is_ or _can be_ localized. (I'm not sure we want react-native-webview to be responsible for managing the localization itself... it may be better to allow passing in a string to use?)

**Environment:**
 - react-native-webview version: 11.18.1

TheAlmightyBob:
This was resolved by #2489 . I'm not sure why it wasn't auto-closed.

PR #2489, opened 2022-05-18
Adding downloadMessage as a prop

mmathewsTableau:
Adding downloadMessage as a prop to react native web view so that the library consumer can specify the message shown in when downloading content.

Meant to address this issue: https://github.com/react-native-webview/react-native-webview/issues/2436

For our app, this allows the message to be localized (see the toast near the bottom):

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -159,6 +159,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   protected boolean mAllowsFullscreenVideo = false;
   protected @Nullable String mUserAgent = null;
   protected @Nullable String mUserAgentWithApplicationName = null;
+  protected String mDownloadMessage = "Downloading";
 
   public RNCWebViewManager() {
     mWebViewConfig = new WebViewConfig() {
@@ -248,8 +249,8 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
         module.setDownloadRequest(request);
 
-        if (module.grantFileDownloaderPermissions()) {
-          module.downloadFile();
+        if (module.grantFileDownloaderPermissions(mDownloadMessage)) {
+          module.downloadFile(mDownloadMessage);
         }
       }
     });
@@ -285,6 +286,12 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   @ReactProp(name = "showsVerticalScrollIndicator")
   public void setShowsVerticalScrollIndicator(WebView view, boolean enabled) {
     view.setVerticalScrollBarEnabled(enabled);
+  }
+
+  @ReactProp(name = "downloadMessage")
+  public void setDownloadMessage(WebView view, String message)
+  {
+    mDownloadMessage = message;
   }
 
   @ReactProp(name = "cacheEnabled")
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -262,11 +262,11 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
     return webView;
   }
 
-  private String getDownloadingMessage(){
+  private String getDownloadingMessage() {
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModule.java
```diff
@@ -95,25 +95,27 @@ private enum MimeType {
...
+      public boolean onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
+        switch (requestCode) {
+          case FILE_DOWNLOAD_PERMISSION_REQUEST: {
+            // If request is cancelled, the result arrays are empty.
+            if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
+              if (downloadRequest != null) {
+                downloadFile(downloadMessage);
+              }
+            } else {
+              Toast.makeText(getCurrentActivity().getApplicationContext(), "Cannot download files as permission was denied. Please provide permission to write to storage, in order to download files.", Toast.LENGTH_LONG).show();
```
TheAlmightyBob, line 110: Oh dang.... somehow I missed that this string was also hardcoded. 😞 

Probably need the same solution here... although it could be a separate PR... and boy this is a much more complicated string to have to localize...

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -159,6 +159,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   protected boolean mAllowsFullscreenVideo = false;
   protected @Nullable String mUserAgent = null;
   protected @Nullable String mUserAgentWithApplicationName = null;
+  protected String mDownloadMessage = "Downloading";
```
TheAlmightyBob, line 162: I'd suggest making "Downloading" a string constant, so for example it could still be used if a custom string was specified and then reset back to `undefined`? (though admittedly I don't know how likely that is... just seems like it would be a small change to be extra safe?)

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -287,6 +288,12 @@ public void setShowsVerticalScrollIndicator(WebView view, boolean enabled) {
     view.setVerticalScrollBarEnabled(enabled);
   }
 
+  @ReactProp(name = "downloadMessage")
```
TheAlmightyBob, line 291: Do you think `downloadingMessage` might be clearer, just to avoid any possible confusing about it being a "download success" message...?

(I know the original variable name was `downloadMessage`, but since that was just a local variable in the function there was less likely to be any confusion)

TheAlmightyBob (changes requested):

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
mmathewsTableau: changed to downloadingMessage

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
mmathewsTableau: Created a separate variable for default strings

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -248,15 +253,23 @@ public void onDownloadStart(String url, String userAgent, String contentDisposit
...
+        if (module.grantFileDownloaderPermissions(getDownloadingMessage(), getLackPermissionToDownloadMessage())) {
+          module.downloadFile(getDownloadingMessage());
         }
       }
     });
 
     return webView;
   }
 
+  private String getDownloadingMessage(){
```
TheAlmightyBob, line 265: ultra-nit, but it would be good to follow the style standard of the rest of the file, which would mean adding a space before the bracket

(as well as updating the setters so that the opening bracket is on the same line as the function signature)

(also it's slightly weird to have the getters separated from the setters, but I totally get that to do otherwise would break the grouping of React props, so I don't have a better idea, just noting that quirk)

mmathewsTableau: Fixed the spacing and placement of the function brackets

TheAlmightyBob (approved):

PR #2490, opened 2022-05-19
Re-indent the Objective-C code

matiaskorhonen:
A minimal reformatting of the Objective-C code for readability using Xcode's re-indent feature (in Xcode: Editor → Structure → Re-indent).

This should make it a bit easier to work on the iOS/macOS native code…


TheAlmightyBob:
I like the idea of cleaning up the formatting with re-indent!

My one concern is that I see a lot of changes that are just doubling the amount of the indent, because it's changing 2-space indenting to 4-space. I wonder if it might be better to change the project default in XCode to 2-space?

image

I tried that very briefly locally, and it still made a lot of changes (because the current spacing is so inconsistent, as you've found), but didn't seem to be _quite_ as many changes as you have here...?

matiaskorhonen:
I went with 4 mostly because that's the Xcode default and seems to be what most people use in the Objective-C world. I'm pretty much open to either option as long as it's consistent…

PR #2517, opened 2022-05-31
fix(windows): don't build WebView2 unless on WinUI 3

tido64:


Titozzz (approved):

chiaramooney:
@tido64 Just seeing this now. This isn't the behavior we want to support on Windows. For the webview module the supported case is RNW v0.68+ with WinUI v2.8.0-prerelease.210927001+. So apps using RNW v68+ just need to upgrade their ExperimentalFeatures file in the example app accordingly once the repo move to RNW v0.68.

PR #2531, opened 2022-06-07
Revert "fix(windows): don't build WebView2 unless on WinUI 3"

tido64:
Reverts react-native-webview/react-native-webview#2517.

WebView2 should not be gated behind WinUI 3.


chiaramooney (approved):
👍 Thanks Tommy!

chiaramooney (approved):

PR #2534, opened 2022-06-09
Re-indent the Objective-C code to 2 spaces

matiaskorhonen:
The same as #2490 but with two space indents instead of four.

A minimal reformatting of the Objective-C code for readability using Xcode's re-indent feature (in Xcode: Editor → Structure → Re-indent).

This should make it a bit easier to work on the iOS/macOS native code…


matiaskorhonen:
@TheAlmightyBob here's a two space version of the re-indentation PR

TheAlmightyBob:
@matiaskorhonen Thanks! 
image
does seem a _little_ less-drastic than 
image
:)

apple/RNCWebView.m
```diff
@@ -456,7 +456,7 @@ - (void)didMoveToWindow
   if (self.menuItems != nil) {
     UILongPressGestureRecognizer *longPress = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(startLongPress:)];
     longPress.delegate = self;
-
+
```
TheAlmightyBob, line 459: Not a big deal, but weird to be that an autoformatting tool would add unnecessary whitespace like this....

TheAlmightyBob (approved):

Issue #2228, opened 2021-11-10
react native webview iOS not keep session cookies when app killed from background then webview automatically logout 


naveenr1592:
**Bug description:**

**To Reproduce:**

**Expected behavior:**

**Screenshots/Videos:**

**Environment:**
 - OS: iOS
 - OS version:13.1
 - react-native version: 64.1
 - react-native-webview version: 11.14.2

gauravrami:
@naveenr1592  **did you find any solution??**

tkmin:
I have a same issue

ANTARES-KOR:
Is there any solution to this? Any options from react-native-webview didn't resolve this issue.

My webview implementation is this
```typescript
        <WebView
          onMessage={handlePostMessage}
          geolocationEnabled={true}
          decelerationRate={"normal"}
          incognito={false}
          cacheEnabled={true}
          thirdPartyCookiesEnabled={true}
          sharedCookiesEnabled={true}
          style={{ flex: 1 }}
        />
```

allan-simon:
is related to https://github.com/react-native-webview/react-native-webview/issues/1780 ?

allan-simon:
I can confirm the issue is still there

minsu-zip:
I had the same problem.

We fixed the issue by specifying the Expires or Max-Age of the cookie.

iOS has the problem of losing cookies without Expires or Max-Age that exist in web views in the background during memory cleanup.

- I was able to find out from the cookie test on iOS safari.
- I haven't tested it in all versions of iOS, but it happens in iOS 15 and later.

RohanDhalpe05:
This worked in my Rails app.
When my webview based app is killed and reopened, the user session is lost and they must log in again. In the browser, the session persists across tab/window closes.

Add config/initializers/session_store.rb (new file):

```
# frozen_string_literal: true

Rails.application.config.session_store :cookie_store, expire_after: 14.days
```
This sets an explicit expiry on the session cookie so it persists across app restarts until the session expires or the user logs out.

The necessary WebView settings (sharedCookiesEnabled, thirdPartyCookiesEnabled, domStorageEnabled, cacheEnabled) are already configured, so the mobile app does not require any changes.

Issue #1780, opened 2020-12-10
iOS cookies: Regardless of 'sharedCookiesEnabled' value, NS/WK cookies (unreliably) synced


sampok:
**Bug description:**
Syncing of cookies on iOS between 'React Native side' (NSHTTPCookieStorage) and WKWebView seems not to follow [the documented `sharedCookiesEnabled` behavior](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Guide.md#managing-cookies) on a real device. I'm observing some type of delayed 'auto-sync' between the two sides and `sharedCookiesEnabled` prop value doesn't really have any effect. I'm seeing the 'auto-updating' happening only on a real device, not on simulator, and there seems to be some unpredictability in the behavior.

I initially observed the behavior with an app having cookies set both by Fetch requests responses and by a website inside react-native-webview WKWebKit. To make a self-contained demo, I put together a simple demo using just react-native-cookies and local react-native-webview. When you set a cookie either for 'React Native side' (NSHTTPCookieStorage) or WKWebView side, the other side often (but not always) updates with delay. Whether `sharedCookiesEnabled` is true or false doesn't seem to have an effect on this.

**Screenshots/Videos:**

<details>
<summary>Screen recordings</summary>
<p>

Setting NS cookie with 'sharedCookiesEnabled' being either undefined or true: WK sometimes updating with a delay:

![nsSet](https://user-images.githubusercontent.com/1516075/101707178-b568a800-3a58-11eb-889c-c8f835d11bc3.gif)

Setting WK cookie: The same happening with NS cookie

![wkSet](https://user-images.githubusercontent.com/1516075/101707184-b8639880-3a58-11eb-968f-e1032590f482.gif)

</p>
</details>  

**To Reproduce:**
Repository: https://github.com/sampok/cookiesync

<details>
<summary>See App.js in the project:</summary>
<p>

```js
import React, {useState, useCallback} from 'react';
import {StyleSheet, View, Text, Button} from 'react-native';
import WebView from 'react-native-webview';
import CookieManager from '@react-native-community/cookies';

const url = 'https://test.com';
const name = 'test';

const App = () => {
  const [nsCookieValues, setNsCookieValues] = useState([]);
  const [wkCookieValues, setWkCookieValues] = useState([]);

  const clearCookie = useCallback(async (useWebKit) => {
    setNsCookieValues([]);
    setWkCookieValues([]);
    CookieManager.clearByName(url, name, useWebKit);
[...]

PR #2535, opened 2022-06-09
Cookie sync improvements

matiaskorhonen:
Use a `WKHTTPCookieStoreObserver` to sync the cookies from the webview to `NSHTTPCookieStorage`.

Most developers shouldn't notice any difference in behaviour but based on my testing this seems to be more reliable than the old way where the cookies were synced in the `didFinishNavigation` callback. This also fixes a race condition bug that we hit with our production app.

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -37,7 +37,7 @@ NSString *const CUSTOM_SELECTOR = @"_CUSTOM_SELECTOR_";
   if (_webView == nil) {
     return nil;
   }
-
+  
   if ([_webView respondsToSelector:@selector(inputAssistantItem)]) {
     UITextInputAssistantItem *inputAssistantItem = [_webView inputAssistantItem];
     inputAssistantItem.leadingBarButtonGroups = @[];
@@ -99,7 +99,7 @@ RCTAutoInsetsProtocol>
 #endif // !TARGET_OS_OSX
   BOOL _savedHideKeyboardAccessoryView;
   BOOL _savedKeyboardDisplayRequiresUserAction;
-
+  
   // Workaround for StatusBar appearance bug for iOS 12
   // https://github.com/react-native-webview/react-native-webview/issues/62
   BOOL _isFullScreenVideoOpen;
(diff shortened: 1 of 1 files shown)


matiaskorhonen:
FYI, rebased against master now, @TheAlmightyBob & @Titozzz

TheAlmightyBob:
Can you explain any more about what unreliability or race conditions you were experiencing that this addresses?

matiaskorhonen:
Essentially rapid changes to the cookies inside the WebView would cause the NSHTTPCookieStorage to go out of sync resulting in authentication errors. Switching to this observer based approach fixed the issue for us.

allan-simon:
@matiaskorhonen  do you think it relates to one of these issues ? https://github.com/react-native-webview/react-native-webview/issues/1780  and/or https://github.com/react-native-webview/react-native-webview/issues/2228 ? (our app also has issue with session stored in cookie so any improvement on that side is more than welcomed :) )

matiaskorhonen:
> do you think it relates to one of these issues ? #1780 and/or #2228

Possibly, though I can't be sure as we mostly rely on the sync working *from* the webview to the shared cookie store (and not the other way around). 

I can say that the patch in this PR has improved the reliability of the cookie sync in our production app…

apple/RNCWebView.m
```diff
@@ -37,7 +37,7 @@ -(id)inputAccessoryView
   if (_webView == nil) {
     return nil;
   }
-
```
allan-simon, line 40: shouldn't the PR  don't touch extra whitespace to ease the size of the diff and increase chance of it being merged ?

allan-simon: @matiaskorhonen  ping ? I think removing the extra whitespace change will increase a lot the chance of it being reviewed/merged

allan-simon:
@TheAlmightyBob did you get time to review this, now that the diff is much much shorter ? :angel:

apple/RNCWebView.m
```diff
@@ -1559,7 +1565,7 @@ - (void)resetupScripts:(WKWebViewConfiguration *)wkWebViewConfig {
       if(!_incognito && !_cacheEnabled) {
         wkWebViewConfig.websiteDataStore = [WKWebsiteDataStore nonPersistentDataStore];
       }
-      [self syncCookiesToWebView:nil];
```
TheAlmightyBob, line 1562: Did you mean to remove the `syncCookiesToWebView` call here? That's not equivalent to the `addObserver` line

TheAlmightyBob (changes requested):
Sorry for the delay, I was hoping that somebody with better understanding of cookies would review, but you're right that it's at least not a big change :)

TheAlmightyBob (approved):

PR #2537, opened 2022-06-10
Implement postMessage/onMessage for Windows WebView2

### LINKED ISSUE #2390: [Webview2 Microsoft Windows] Implements postMessage and inject javascript in Webview2
[Webview2 implementation for Windows issue](Ref issue: https://github.com/react-native-webview/react-native-webview/issues/1848)

**We need Webview2 to implements javascript injection in order to communicate from ReactNative to the web view content.**
[injectJavascript](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#injectjavascriptstr)

**postMessage / onMessage**
We also need Webview2 component to support « window.postMessage » and the react native Webview2 to support « onMessage », to have a two way bridge communication channel.
[onMessage reference](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#onmessage)

(depends on implementation of AddWebAllowedObject API)

chiaramooney:
Implement postMessage/onMessage for Windows on WebView2. 

Resolves #2390


windows/ReactNativeWebView/ReactWebView2.cpp
```diff
@@ -85,6 +89,25 @@ namespace winrt::ReactNativeWebView::implementation {
...
+
+                
+        if (m_messagingEnabled) {
+            m_messageToken = webView.WebMessageReceived([this](auto const& sender, winrt::CoreWebView2WebMessageReceivedEventArgs const& args)
+                {
+                    try {
+                        auto message = args.TryGetWebMessageAsString();
+                        this->OnMessagePosted(message);
+                    }
+                    catch (std::exception e) {
```
asklar, line 101: IIRC, exceptions from winrt are not std::exception (they're e.g. `winrt::hresult_error`). Since you aren't doing anything with the exception, you can just `catch (...) ` instead to catch any type

related: in general you should catch exceptions by `const&` so we don't need a copy ctor call

windows/ReactNativeWebView/ReactWebView2.cpp
```diff
@@ -85,6 +89,25 @@ namespace winrt::ReactNativeWebView::implementation {
                 WriteWebViewNavigationEventArg(webView, eventDataWriter);
                 eventDataWriter.WriteObjectEnd();
             });
+
+                
+        if (m_messagingEnabled) {
+            m_messageToken = webView.WebMessageReceived([this](auto const& sender, winrt::CoreWebView2WebMessageReceivedEventArgs const& args)
```
asklar, line 95: looks like the sender param is unused; this causes a warning "unreferenced formal parameter"; either remove the name "sender", or write it as `auto const& /* sender */, winrt::CoreWebView2WebMessageReceivedEventArgs const&`

windows/ReactNativeWebView/ReactWebView2.cpp
```diff
@@ -96,6 +119,13 @@ namespace winrt::ReactNativeWebView::implementation {
                 WriteWebViewNavigationEventArg(webView, eventDataWriter);
                 eventDataWriter.WriteObjectEnd();
             });
+
+        if (m_messagingEnabled) {
+            winrt::hstring windowAlert = L"window.alert = function (msg) {window.chrome.webview.postMessage(`{\"type\":\"__alert\",\"message\":\"${msg}\"}`)};";
+            winrt::hstring postMessage = L"window.ReactNativeWebView = {postMessage: function (data) {window.chrome.webview.postMessage(String(data))}};";
```
asklar, line 125: 1) when passing strings that have embedded quotes, use raw string literals to make them easier to write / read: https://docs.microsoft.com/en-us/cpp/cpp/string-and-character-literals-cpp?view=msvc-170

2) don't create unnecessary HSTRINGs, put everything in one :)

3) in this case it seems like it doesn't matter but in general you might want to await the ExecuteScriptAsync to make sure the control has processed adding the alert / RNWV properties to the window object.

windows/ReactNativeWebView/ReactWebView2.cpp
```diff
@@ -96,6 +119,13 @@ namespace winrt::ReactNativeWebView::implementation {
...
+            webView.ExecuteScriptAsync(windowAlert);
+            webView.ExecuteScriptAsync(postMessage);
+        }
     }
 
     void ReactWebView2::OnCoreWebView2Initialized(winrt::Microsoft::UI::Xaml::Controls::WebView2 const& sender, winrt::Microsoft::UI::Xaml::Controls::CoreWebView2InitializedEventArgs const& args) {
```
asklar, line 131: remove the args name here to remove the warnings (same in other places where a param name is not referenced)

windows/ReactNativeWebView/ReactWebView2.cpp
```diff
@@ -106,6 +136,35 @@ namespace winrt::ReactNativeWebView::implementation {
         }
     }
 
+    void ReactWebView2::HandleMessageFromJS(winrt::hstring const& message) {
+        winrt::JsonObject jsonObject;
+        if (winrt::JsonObject::TryParse(message, jsonObject) && jsonObject.HasKey(L"type")) {
```
asklar, line 141: you'll want to make sure that the json value kind is string (same in other places).

One pattern you can use instead of HasKey:

```cpp
if (auto v = jsonObject.Lookup(L"type"); v && v.ValueType() == JsonValueType::String) {
  auto type = v.GetString();
  ...
```

windows/ReactNativeWebView/ReactWebView2.cpp
```diff
@@ -106,6 +136,35 @@ namespace winrt::ReactNativeWebView::implementation {
...
+            auto type = jsonObject.GetNamedString(L"type");
+            if (type == L"__alert") {
+                auto dialog = winrt::MessageDialog(jsonObject.GetNamedString(L"message"));
+                dialog.Commands().Append(winrt::UICommand(L"OK"));
```
asklar, line 145: I think this probably should be localized so it shows up in the right language. IIRC XAML already has these strings so you can probably use those

windows/ReactNativeWebView/ReactWebView2Manager.cpp
```diff
@@ -134,13 +140,13 @@ namespace winrt::ReactNativeWebView::implementation {
             }
         }
         else if (commandId == L"reload") {
-            webView.Refresh();
+            webView.Reload();
         }
         else if (commandId == L"stopLoading") {
-            webView.Stop();
```
asklar, line 140: why is this not calling Stop anymore?

asklar: I think for WV2 you'll want `webView.CoreWebView2().Stop()`

windows/ReactNativeWebView/ReactWebView2Manager.cpp
```diff
@@ -120,7 +126,7 @@ namespace winrt::ReactNativeWebView::implementation {
         winrt::IJSValueReader const& commandArgsReader) noexcept {
         auto control = view.as<winrt::ContentPresenter>();
         auto content = control.Content();
-        auto webView = content.as<winrt::WebView>();
+        auto webView = content.as<winrt::WebView2>();
```
asklar, line 129: just to confirm, this function is only used for WV2 and not for WV1 anymore, right?

windows/ReactNativeWebView/pch.h
```diff
@@ -14,6 +14,7 @@
 #include <winrt/Windows.Web.Http.Headers.h>
 #include <winrt/Microsoft.ReactNative.h>
 #include <winrt/Microsoft.UI.Xaml.Controls.h>
+#include <winrt/Microsoft.Web.WebView2.Core.h>
```
asklar, line 17: this will only be present if using a WinUI 2.x version with WV2 (prerelease only as of today), so it needs to be guarded so that folks can use react-native-webview to use WV1 if they're using WinUI 2.6, 2.7, etc.

windows/ReactNativeWebView/ReactWebView2.h
```diff
@@ -24,23 +24,27 @@ namespace winrt::ReactNativeWebView::implementation {
     class ReactWebView2 : public ReactWebView2T<ReactWebView2> {
```
asklar, line 24: I think you will want to move the macros in lines 6-10 to the pch so it can include reason over including the webview2 cppwinrt headers too

windows/ReactNativeWebView/ReactWebView.idl
```diff
@@ -14,6 +14,7 @@ namespace ReactNativeWebView{
     runtimeclass ReactWebView2 : Windows.UI.Xaml.Controls.ContentPresenter{
         ReactWebView2(Microsoft.ReactNative.IReactContext context);
         void NavigateToHtml(String html);
+        void SetMessagingEnabled(Boolean enabled);
```
asklar, line 17: prefer a property for this since it's just setting a bool:
```
Boolean MessagingEnabled { get; set; }
```

chiaramooney: This is the structure of the original WebView code. Seems like a good change to make though. I'll make a note to cleanup the original WebView as well in a PR after this.

windows/ReactNativeWebView/ReactWebView2Manager.cpp
chiaramooney: Yes! Because this is the ViewManager for the WebView2 control.

windows/ReactNativeWebView/ReactWebView2Manager.cpp
chiaramooney: That makes sense! I combed through the webview2 APIs and couldn't find a stop method. I'll try that option.

windows/ReactNativeWebView/ReactWebView2.cpp
chiaramooney: This is the structure of the original WebView code. Seems like a good change to make though. Could you clarify what you mean by "XAML already has these strings so you can probably use those"

windows/ReactNativeWebView/ReactWebView2.cpp
chiaramooney: This section was just copied from the original WebView code. Seems like a good change to make though.

windows/ReactNativeWebView/ReactWebView2.cpp
chiaramooney: This section was just copied from the original WebView code. Seems like a good change to make though.

windows/ReactNativeWebView/ReactWebView2.cpp
chiaramooney: Will resolve later in a fresh PR.

windows/ReactNativeWebView/pch.h
```diff
@@ -1,6 +1,19 @@
...
 #define NOMINMAX
+
+#if __has_include(<VersionMacros.h>)
+#include <VersionMacros.h>
+#else
+#define RNW_VERSION_AT_LEAST(x,y,z) false
+#endif
+
+#if RNW_VERSION_AT_LEAST(0,68,0)
+#define USE_WEBVIEW2 1
```
asklar, line 12: rename to HAS_WEBVIEW2 (since you can still choose to use WebView 1 with a WinUI 2.x version that contains WV2)

windows/ReactNativeWebView/ReactWebView2.h
```diff
@@ -24,7 +18,8 @@ namespace winrt::ReactNativeWebView::implementation {
...
     public:
         ReactWebView2(Microsoft::ReactNative::IReactContext const& reactContext);
-        void SetMessagingEnabled(bool enabled);
+        void MessagingEnabled(bool enabled);
+        bool MessagingEnabled();
```
asklar, line 22: nit, since these are simple setter/getter, you can mark them noexcept/const where possible so the compiler can generate better code:

```suggestion
        void MessagingEnabled(bool enabled) noexcept;
        bool MessagingEnabled() const noexcept;
```

windows/ReactNativeWebView/ReactWebView2.cpp
```diff
@@ -160,10 +160,14 @@ namespace winrt::ReactNativeWebView::implementation {
...
     }
 
-    void ReactWebView2::SetMessagingEnabled(bool enabled) {
+    void ReactWebView2::MessagingEnabled(bool enabled) {
         m_messagingEnabled = enabled;
     }
 
+    bool ReactWebView2::MessagingEnabled() {
+        return m_messagingEnabled;
+    }
```
asklar, line 169: if you take the other feedback, update here:
```suggestion
    void ReactWebView2::MessagingEnabled(bool enabled) noexcept {
        m_messagingEnabled = enabled;
    }

    bool ReactWebView2::MessagingEnabled() const noexcept {
        return m_messagingEnabled;
    }
```

windows/ReactNativeWebView/ReactWebView2.cpp
```diff
@@ -121,14 +121,12 @@ namespace winrt::ReactNativeWebView::implementation {
             });
 
         if (m_messagingEnabled) {
-            winrt::hstring windowAlert = L"window.alert = function (msg) {window.chrome.webview.postMessage(`{\"type\":\"__alert\",\"message\":\"${msg}\"}`)};";
-            winrt::hstring postMessage = L"window.ReactNativeWebView = {postMessage: function (data) {window.chrome.webview.postMessage(String(data))}};";
-            webView.ExecuteScriptAsync(windowAlert);
-            webView.ExecuteScriptAsync(postMessage);
+            winrt::hstring message = LR"(window.alert = function (msg) {window.chrome.webview.postMessage(`{"type":"__alert","message":"${msg}"}`)}; window.ReactNativeWebView = {postMessage: function (data) {window.chrome.webview.postMessage(String(data))}};)";
```
asklar, line 124: Nit, break this up in a couple of lines within the same string, for example:

```suggestion
           winrt::hstring message = LR"(
               window.alert = function (msg) {window.chrome.webview.postMessage(`{"type":"__alert","message":"${msg}"}`)};
               window.ReactNativeWebView = {postMessage: function (data) {window.chrome.webview.postMessage(String(data))}};)";
```

windows/ReactNativeWebView/ReactWebView.idl
```diff
@@ -14,7 +14,7 @@ namespace ReactNativeWebView{
     runtimeclass ReactWebView2 : Windows.UI.Xaml.Controls.ContentPresenter{
         ReactWebView2(Microsoft.ReactNative.IReactContext context);
         void NavigateToHtml(String html);
-        void SetMessagingEnabled(Boolean enabled);
+        Boolean MessagingEnabled{ get; set; };
```
asklar, line 17: you can remove the get;set; if it's a readwrite property:
```suggestion
        Boolean MessagingEnabled;
```


(Same above)

asklar (dismissed):
lgtm; left a couple of nit/comments

asklar (approved):

PR #2548, opened 2022-06-20
fix: add properties for ios WKWebView configuration

hojunin:
Several properties for WKWebView configuration is missing. 
we pass properties below, but react-native-webview don`t to ios.

* dataDetectorTypes
* allowsAirPlayForMediaPlayback
* allowsInlineMediaPlayback

plz check this simple changes. 
if this is my mistake or am i missing something, comment freely.
thank you.

### DIFF
--- a/src/WebView.macos.tsx
+++ b/src/WebView.macos.tsx
@@ -160,6 +160,10 @@ const WebViewComponent = forwardRef<{}, MacOSWebViewProps>(({
       onContentProcessDidTerminate={onContentProcessDidTerminate}
       injectedJavaScript={injectedJavaScript}
       injectedJavaScriptBeforeContentLoaded={injectedJavaScriptBeforeContentLoaded}
+      allowsAirPlayForMediaPlayback={allowsAirPlayForMediaPlayback}
+      allowsInlineMediaPlayback={allowsInlineMediaPlayback}
+      incognito={incognito}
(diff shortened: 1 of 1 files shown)


TheAlmightyBob:
@Titozzz It looks like this is related to the recent refactor? Seems like these props were previously part of `otherProps` and so were getting passed along with that, but they aren't anymore and aren't being explicitly passed either.

Just glancing at the code, I'm _guessing_ this also affects `mediaPlaybackRequiresUserAction` & `incognito`...? possibly others I'm missing...

hojunin:
> @Titozzz It looks like this is related to the recent refactor? Seems like these props were previously part of `otherProps` and so were getting passed along with that, but they aren't anymore and aren't being explicitly passed either.
> 
> Just glancing at the code, I'm _guessing_ this also affects `mediaPlaybackRequiresUserAction` & `incognito`...? possibly others I'm missing...

i check other properties like `mediaPlaybackRequiresUserAction`, `incognito` ...and so on in  [WKWebview Config Docs](https://developer.apple.com/documentation/webkit/wkwebviewconfiguration). props that deprecated or cannot find is excepted!


스크린샷 2022-06-21 오전 11 40 12

Titozzz:
I.ll have a look today, thanks

Titozzz:
Yeah my bad while refactoring theses were not propagated due to the fact that they were not extracted from otherProps. I'll land a fix soon

Titozzz (approved):

Issue #1609, opened 2020-09-03
Code from the injectedJavaScriptBeforeContentLoaded prop runs AFTER the web code.


Kreozot:
**Bug description:**

According to official docs of `react-native-webview` (The injectedJavaScriptBeforeContentLoaded prop):

> This is a script that runs before the web page loads for the first time. It only runs once, even if the page is reloaded or navigated away. This is useful if you want to inject anything into the window, localstorage, or document prior to the web code executing.

But we can see that the script on the page runs first, and only then runs the code from the `injectedJavaScriptBeforeContentLoaded` prop.

The setup is:

* index.html with test.js script: https://github.com/Kreozot/webview-scripts-test/tree/master/android/app/src/main/assets
  The script has alert _"script executed"_

* WebView with the injectedJavaScriptBeforeContentLoaded prop: https://github.com/Kreozot/webview-scripts-test/blob/master/App.js
  The prop has alert _"injectedJavaScriptBeforeContentLoaded"_

**To Reproduce:**

1. `git clone git@github.com:Kreozot/webview-scripts-test.git`
1. `cd webview-scripts-test`
1. `yarn`
1. `yarn run android`

**Expected behavior:**

Alert "injectedJavaScriptBeforeContentLoaded" shows **before** "script executed".

**Screenshots/Videos:**

Alert "injectedJavaScriptBeforeContentLoaded" shows **after** "script executed".

**Environment:**
 - OS: Android
 - OS version: 9
 - react-native version: 0.63.2
 - react-native-webview version: 10.8.3

Kreozot:
I have tested it on 10.3.0 version of react-native-webview and it behave as expected. So, maybe it's a new bug.

Kreozot:
UPD: sometimes it works as expected, but like 50/50.

renwenci:
i have the same issue on IOS 13.6.1, anyone has ideas?

roscorcoran:
Same or similar issues here, however on Android the Javascript injected via _injectedJavaScriptBeforeContentLoaded_ never runs, even when Javascript injected via _injectedJavaScript_ does run.

Kreozot:
Dear bot, please don't close this issue because it still an issue.

ArthurSampaio:
Hi @Kreozot, did you find an away or workaround to solve this?

Kreozot:
> Hi @Kreozot, did you find an away or workaround to solve this?

[...]

PR #2549, opened 2022-06-20
docs: prop also available for Android

balsloev:
It was stated that the prop `injectedJavaScriptBeforeContentLoaded` was only available for iOS and macOS. 
But I found that it is also implemented for Android. I have updated the documentation accordingly.

**References can be found here**

[/src/WebViewTypes.ts#L287](https://github.com/react-native-webview/react-native-webview/blob/53ff8fb86691e6d143f67e3351e722e7f87f2e65/src/WebViewTypes.ts#L287)

[/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L483 ](https://github.com/react-native-webview/react-native-webview/blob/53ff8fb86691e6d143f67e3351e722e7f87f2e65/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L483)

[/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L943](https://github.com/react-native-webview/react-native-webview/blob/53ff8fb86691e6d143f67e3351e722e7f87f2e65/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L943)

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -198,9 +198,9 @@ Make sure the string evaluates to a valid type (`true` works) and doesn't otherw
 
 On iOS, see [`WKUserScriptInjectionTimeAtDocumentStart`](https://developer.apple.com/documentation/webkit/wkuserscriptinjectiontime/wkuserscriptinjectiontimeatdocumentstart?language=objc)
 
-| Type   | Required | Platform   |
-| ------ | -------- | ---------- |
-| string | No       | iOS, macOS |
+| Type   | Required | Platform            |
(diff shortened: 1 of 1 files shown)


TheAlmightyBob:
Thanks for catching this.

Have you found that the property is actually working reliably on Android? There was quite a bit of discussion in the MR (https://github.com/react-native-webview/react-native-webview/pull/1099) and at least one issue (#1609) saying it doesn't quite work as expected.

PR #2575, opened 2022-06-30
fix(windows): Add WinUI Version Opt In Functionality

chiaramooney:
Add functionality to opt into WinUI 2.8.0-prerelease for v68+ RNW apps using the Webview module.

### DIFF
--- a/windows/ReactNativeWebView/ReactNativeWebView.vcxproj
+++ b/windows/ReactNativeWebView/ReactNativeWebView.vcxproj
@@ -81,7 +81,9 @@
     <Import Project="$(ReactNativeWindowsDir)\PropertySheets\external\Microsoft.ReactNative.Uwp.CppLib.props" Condition="Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.props')" />
     <Import Project="$(SolutionDir)\packages\$(WinUIPackageProps)" Condition="'$(WinUIPackageProps)'!='' And Exists('$(SolutionDir)\packages\$(WinUIPackageProps)')" />
   </ImportGroup>
-  <PropertyGroup Label="UserMacros" />
+  <PropertyGroup Label="UserMacros">
+    <DefineConstants Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(WinUIPackageVersion)','2.8.0-prerelease.210927001'))">WINUI_VERSION_COMPAT;$(DefineConstants)</DefineConstants>
+  </PropertyGroup>
   <PropertyGroup />
   <ItemDefinitionGroup>
     <ClCompile>
@@ -112,6 +114,16 @@
     <ClCompile>
       <PreprocessorDefinitions>NDEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
     </ClCompile>
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(WinUIPackageVersion)','2.8.0-prerelease.210927001'))">
+    <ClCompile>
+      <PreprocessorDefinitions>WINUI_VERSION_COMPAT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(WinUIPackageVersion)','2.8.0-prerelease.210927001'))">
+    <Midl>
+      <PreprocessorDefinitions>WINUI_VERSION_COMPAT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </Midl>
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClInclude Include="ReactWebView.h" />
--- a/windows/ReactNativeWebView/ReactWebView.idl
+++ b/windows/ReactNativeWebView/ReactWebView.idl
@@ -9,7 +9,12 @@ namespace ReactNativeWebView{
 #else
 #define RNW_VERSION_AT_LEAST(x,y,z) false
 #endif
-#if RNW_VERSION_AT_LEAST(0,68,0)
+
+#ifndef WINUI_VERSION_COMPAT
+#define WINUI_VERSION_COMPAT 0
+#endif 
+
+#if RNW_VERSION_AT_LEAST(0,68,0) && WINUI_VERSION_COMPAT
     [default_interface]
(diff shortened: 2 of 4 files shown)


windows/ReactNativeWebView/ReactNativeWebView.vcxproj
```diff
@@ -81,7 +81,9 @@
     <Import Project="$(ReactNativeWindowsDir)\PropertySheets\external\Microsoft.ReactNative.Uwp.CppLib.props" Condition="Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.props')" />
     <Import Project="$(SolutionDir)\packages\$(WinUIPackageProps)" Condition="'$(WinUIPackageProps)'!='' And Exists('$(SolutionDir)\packages\$(WinUIPackageProps)')" />
   </ImportGroup>
-  <PropertyGroup Label="UserMacros" />
+  <PropertyGroup Label="UserMacros">
+    <DefineConstants Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(WinUIPackageVersion)','2.8.0-prerelease.210927001'))">WINUI_VERSION_COMPAT;$(DefineConstants)</DefineConstants>
```
asklar, line 85: rename WINUI_VERSION_COMPAT to something like WINUI2_HAS_WEBVIEW

asklar: actually probably better to use:
```suggestion
    <DefineConstants Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(WinUIPackageVersion)','2.8.0-prerelease.210927001'))">WINUI_VERSION_COMPAT;$(DefineConstants)</DefineConstants>
Collaborator
@asklar asklar 1 minute ago
rename WINUI_VERSION_COMPAT to something like WINUI2_HAS_WEBVIEW

 

actually probably better to use:
```suggestion

No file chosen
Attach files by dragging & dropping, selecting or pasting them.
  </PropertyGroup>
  <PropertyGroup />
  <ItemDefinitionGroup>
    <ClCompile>
@@ -113,6 +115,16 @@
      <PreprocessorDefinitions>NDEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
    </ClCompile>
  </ItemDefinitionGroup>
  <ItemDefinitionGroup Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(WinUIPackageVersion)','2.8.0-prerelease.210927001'))">
    <ClCompile>
      <PreprocessorDefinitions>WINUI_VERSION_COMPAT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
    </ClCompile>
  </ItemDefinitionGroup>
  <ItemDefinitionGroup Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(WinUIPackageVersion)','2.8.0-prerelease.210927001'))">
    <Midl>
      <PreprocessorDefinitions>WINUI_VERSION_COMPAT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
    </Midl>
  </ItemDefinitionGroup>
Commenting on lines +118 to +127
 

can replace this with:
```suggestion
<WinUI2_HasWebView2 Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(WinUIPackageVersion)','2.8.0-prerelease.210927001'))">true</WinUI2_HasWebView2>
```

windows/ReactNativeWebView/ReactNativeWebView.vcxproj
```diff
@@ -81,7 +81,9 @@
     <Import Project="$(ReactNativeWindowsDir)\PropertySheets\external\Microsoft.ReactNative.Uwp.CppLib.props" Condition="Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.props')" />
     <Import Project="$(SolutionDir)\packages\$(WinUIPackageProps)" Condition="'$(WinUIPackageProps)'!='' And Exists('$(SolutionDir)\packages\$(WinUIPackageProps)')" />
   </ImportGroup>
-  <PropertyGroup Label="UserMacros" />
+  <PropertyGroup Label="UserMacros">
+    <WinUI2_HasWebView2 Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(WinUIPackageVersion)','2.8.0-prerelease.210927001'))">true</WinUI2_HasWebView2>
```
asklar, line 85: remove the property group label (UserMacros) since they're not the user's to modify

windows/ReactNativeWebView/ReactWebView.idl
```diff
@@ -9,7 +9,12 @@ namespace ReactNativeWebView{
 #else
 #define RNW_VERSION_AT_LEAST(x,y,z) false
 #endif
-#if RNW_VERSION_AT_LEAST(0,68,0)
+
+#ifndef WINUI2_HAS_WEBVIEW2
+#define WINUI2_HAS_WEBVIEW2 0
+#endif 
+
+#if RNW_VERSION_AT_LEAST(0,68,0) && WINUI2_HAS_WEBVIEW2
```
asklar, line 17: if we don't have webview2, then WINUI2_HAS_WEBVIEW2 will be defined as 0. 
If we do have it, then it will be defined but have no value.
Therefore this line will expand to:
```
#if <something> &&
```
note the trailing `&&`, which will lead to a compilation error

You either need to make sure that macro is always defined to 0 or 1, or better yet, remove 13-15 and replace 17 with `#if RNW_VERSION_AT_LEAST(0,68,0) && defined(WINUI2_HAS_WEBVIEW2)`

windows/ReactNativeWebView/pch.h
```diff
@@ -8,7 +8,11 @@
 #define RNW_VERSION_AT_LEAST(x,y,z) false
 #endif
 
-#if RNW_VERSION_AT_LEAST(0,68,0)
+#ifndef WINUI2_HAS_WEBVIEW2
+#define WINUI2_HAS_WEBVIEW2 0
+#endif 
+
+#if RNW_VERSION_AT_LEAST(0,68,0) && WINUI2_HAS_WEBVIEW2
```
asklar, line 15: same comment here

asklar (changes requested):

windows/ReactNativeWebView/ReactWebView.idl
chiaramooney: Ok I can make that change. What's interesting though is that when I test, WINUI2_HAS_WEBVIEW2 is always evaluating to 1 or 0; I haven't encountered the case you're mentioning where the value is defined with no value and compilation fails. Based on what I read online, it seemed as if the default value is assumed to be 1 unless otherwise specified.

asklar (approved):

Issue #1190, opened 2020-02-06
FullScreen Observer listening for event that is not limited to just entering Full Screen, also triggered on KeyboardUI windows showing.


tyleragnew:
**Bug description:**

There is an Observer in RNCWebView.m that listens for UIWindowDidBecomeVisibleNotification events being triggered, and then selects the method showFullScreenVideoStatusBars to execute. I do not think this is the appropriate event to listen for when attempting to determine whether or not a FullScreenVideo view has indeed been entered. In fact, if a keyboard is shown, it will also emit a UIWindowDidBecomeVisibleNotification. This is causing issues for the Status Bar when there is a WebView component in the project, because the showFullScreenVideoStatusBars method is hardcoding the statusBarStyle to LightContent (causing it to disappear for screens where the background is white). 

I think it would be best to truly search for an event that is scoped only to only a fullScreenVideo being displayed, to take care of this. UPDATE: It looks like a feature request has come in specifically for this type of event: https://github.com/react-native-community/react-native-webview/issues/1189

**To Reproduce:**

In a component using WebView, do an action that requires a keyboard window to display. You will see that the statusBar ends up toggling its style. This is the expected behavior for entering a fullScreenVideo, but not for just any keyboard event.

**Expected behavior:**

A keyboard event should not trigger a method that is supposed to just listen for changes to the fullScreen status affect the status bar. 

**Screenshots/Videos:**
[KeyboardActionTriggersFullScreenMode.zip](https://github.com/react-native-community/react-native-webview/files/4165675/KeyboardActionTriggersFullScreenMode.zip)

**Environment:**

 - iOS: 13
 - react-native version: .60.8
 - react-native-webview version: 7.6.0 and 8.0.6

trollepierre:
this bug is still painful.
Related #2576

PR #2576, opened 2022-07-01
fix(status-bar): prevent status bar to be hidden, on KeyboardUI event.

trollepierre:
According to the not fixed bug #1190 ,
> FullScreen Observer listening for event that is not limited to just entering Full Screen, also triggered on KeyboardUI windows showing.

Inside our app that is not using FullScreenVideoOpen, our KeyboardUI events are messing up our StatusBar state.

That's why this PR prevents the StatusBar changes.
This PR is well inspired by https://github.com/react-native-webview/react-native-webview/issues/735#issuecomment-581574089 from @[tyleragnew](https://github.com/tyleragnew)

However, the best solution (to me) would be:
- Do not call `showFullScreenVideoStatusBars` and `hideFullScreenVideoStatusBars` during KeyboardUI event.

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -501,7 +501,8 @@ RCTAutoInsetsProtocol>
   
   _isFullScreenVideoOpen = YES;
   RCTUnsafeExecuteOnMainQueueSync(^{
-    [RCTSharedApplication() setStatusBarStyle:self->_savedStatusBarStyle animated:YES];
+    [RCTSharedApplication() setStatusBarHidden:NO animated:YES];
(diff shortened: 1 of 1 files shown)


TheAlmightyBob:
@trollepierre Setting `autoManageStatusBarEnabled` to `false` doesn't work for you?

https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#autoManageStatusBarEnabled

TheAlmightyBob:
@trollepierre Did you happen to look into why it doesn't work? The functions you're modifying look like they're checking for that setting and do nothing if it's set... so if that's not working then it seems like maybe that's the bug to be fixed.

Unless the problem is that you _do_ want these functions to execute (i.e. you want the status bar to be managed), but you just want them to do different things from the current implementation...?

PR #2580, opened 2022-07-04
fix(iOS): Adding the contentProcessDidTerminate property into useWebViewLogic hook arguments (fixes #2559)

### LINKED ISSUE #2559: onContentProcessDidTerminate not called in newer versions of react-native-webview + react-native
Thanks for making this library, here is my first bug report

**Bug description:**
I recently upgraded some of our packages, and noticed that the callback given to onContentProcessDidTerminate was no longer being called.  However, react-native-webview was detecting that the webview had terminated.  We use this callback to reload the webview if it is terminated due to OS memory constraints.

**To Reproduce:**
1. Create a fresh react-native project following instructions:
https://reactnative.dev/docs/environment-setup
i.e.
```
npx react-native init AwesomeProject
```
2. Install react-native-webview via npm + update pods

3. Import into App.js WebView
```
```

4. Replace App.js rendering code with:
```
  return (
    <SafeAreaView style={{ flex: 1 }}>
          <WebView
          source={{ uri: 'https://google.com' }} 
          onContentProcessDidTerminate={(event)=>{console.log("END OF A WEBVIEW!")}}
        />
    </SafeAreaView>
  );
```

5. Run on device from xcode as per:
https://reactnative.dev/docs/running-on-device

6. Trigger webViewWebContentProcessDidTerminate.  For example, by opening and refreshing many high-memory websites in Safari.
(_If anyone knows of an easier and more reliable way to trigger this please let me know_)

Result: No console message "END OF A WEBVIEW!"

**Expected behavior:**
onContentProcessDidTerminate, as provided in the react component property, is called.

**Environments:**
This occurs for me on:
node version: 16.15.1
react-native: 0.67.1
react-native-webview: 11.22.3

node version: 18.4.0
react-native: 0.69.0
react-native-webview: 11.22.3

node version: 16.15.1
react-native: 0.68.2
react-native-webview: 11.22.2

All using on-device development build + run from xcode:
xcode: 13.4.1
ios: 15.5
device: ipad 6th generation MR7J2B/A

The lack of auto-restart of the webview was also reported by users of our main app via test flight.

The issue does NOT occur for the following configuration (it runs successfully):
node version: 16.15.1
react-native: 0.67.1
react-native-webview: 11.6.4

**Further investigations:**
Further investigation proceeded with the following 'hack':
Adding the `onScroll` callback to the WebView component as follows in `App.js`
```
          onScroll={(event)=>{console.log("SCROLL!")}}
```
Then modifying `node_modules/react-native-webview/apple/RNCWebView.m` such that we disable the `_onScroll(event);` event at line 882 in the `scrollViewDidScroll` method.  And finally add the `_onScroll(event);` back in after `_onContentProcessDidTerminate(event);` on line 1186 in `webViewWebContentProcessDidTerminate`.

In this case, the `onScroll` callback fired on termination but still not onContentProcessDidTerminate.  This demonstrates that the `_onContentProcessDidTerminate` method exists (confirmed by the. debugger) and that the event can flow into the react-native component via onScroll, but not for onContentProcessDidTerminate!

**Request:**
My current thoughts are that this is something to do with react-native's upgrade limiting the props that can be called from the C function.
I would appreciate if anyone else can confirm they are experiencing this issue.  I'm not at all well versed in C or how react-native interacts with the C functions, so any suggestions on that front would be greatly appreciated.

For now we plan to downgrade react-native-webview to version 11.13.0

moonsift:
Fixes #2559 as suggested by the comments there.

Successfully tested on ipad 6th gen, iOS 15.5 following procedure in issue.

### DIFF
--- a/src/WebView.ios.tsx
+++ b/src/WebView.ios.tsx
@@ -72,6 +72,7 @@ const WebViewComponent = forwardRef<{}, IOSWebViewProps>(({
   onLoad,
   onLoadEnd,
   onLoadProgress,
+  onContentProcessDidTerminate: onContentProcessDidTerminateProp,
   onFileDownload,
   onHttpError: onHttpErrorProp,
   onMessage: onMessageProp,
@@ -117,6 +118,7 @@ const WebViewComponent = forwardRef<{}, IOSWebViewProps>(({
     originWhitelist,
     onShouldStartLoadWithRequestProp,
     onShouldStartLoadWithRequestCallback,
+    onContentProcessDidTerminateProp,
   });
 
   useImperativeHandle(ref, () => ({


devth:
Thanks @moonsift!!

Hope this can get released soon 🙏🏻

TheAlmightyBob (approved):

TheAlmightyBob:
Thanks @moonsift!

bfelbo:
For anyone else curious if this issue is affecting them (and thereby need to upgrade), it seems to only affect versions [v11.22.0](https://github.com/react-native-webview/react-native-webview/releases/tag/v11.22.0) onwards.

Issue #2578, opened 2022-07-02
 Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.6.0, expected version is 1.1.15.


nithinshiriya:
Receiving invalid kotlin versions

 Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.6.0, expected version is 1.1.15.

React : 0.69.0

jetified-kotlin-stdlib-1.6.10.jar!/kotlin/Unit.class
e: /Users/rao/Desktop/trignofy/bb-update/bellesboard-mobile-v2/node_modules/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/events/TopShouldStartLoadWithRequestEvent.kt: (28, 21): Class 'kotlin.Unit' was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.6.0, expected version is 1.1.15.

nithinshiriya:
Fix: Update  ReactNativeWebView_kotlinVersion=1.6.0

fonov:
add 

```grandle
buildscript {
    ext {
      kotlinVersion = "1.6.0"
     ...
    }
  ...
}   
```

in `android/build.gradle`

juniordeleon013:
> add
> 
> ```
> buildscript {
>     ext {
>       kotlinVersion = "1.6.0"
>      ...
>     }
>   ...
> }   
> ```
> 
> in `android/build.gradle`

This solved my problem, thanks

vorlovsky:
I've got the same issue too:  
```
> Task :react-native-webview:compileDebugKotlin FAILED
w: Runtime JAR files in the classpath should have the same version. These files were found in the classpath:
    C:/Users/LJ/.gradle/caches/transforms-3/8d4b2b3bf878949ebb0dcdeaf1145afb/transformed/jetified-kotlin-stdlib-jdk8-1.4.10.jar (version 1.4)
    C:/Users/LJ/.gradle/caches/transforms-3/e49ff03ad7eade9ef547aab0326f2bbe/transformed/jetified-kotlin-stdlib-jdk7-1.4.10.jar (version 1.4)
    C:/Users/LJ/.gradle/caches/transforms-3/a7fcf88a6a347bcc001af6c4b15a73b1/transformed/jetified-kotlin-stdlib-1.6.0.jar (version 1.6)
[...]

PR #2592, opened 2022-07-07
fix(android): bump up kotlinVersion to 1.6.0

craftzdog:
Fixes the error "Module was compiled with an incompatible version of Kotlin" #2578


jamonholmgren:
@Titozzz I don't think updating the default version is an issue, necessarily. The library will still use the `ext` version if they specify it.

```
buildscript {
    ext {
      // if they specify this, it'll be used instead
      kotlinVersion = "1.6.0"
     ...
    }
  ...
}   
```

So I think we could merge this.

jamonholmgren (approved):

PR #2609, opened 2022-07-21
chore(docs): add page navigation gesture & button support

### LINKED ISSUE #2466: Explain A Clear Way To Navigate In the Webview and Provide Example Codes In Function Components
Hi, 

There is no clear explanation of how to navigate inside the web pages that load in a webview and it frustrated me.
I had to research for hours before seeing something. [This guide](https://dev.to/mukeshmandiwal/webview-data-flow-with-webviews-in-a-react-native-app-and-navigation-4396)

It will be great for a clear explanation here in the docs. The navigation I am referring to is inside the Webview itself, if I land on the home page of a website, then click on a link to go to another page and then another, when I then click my phone's back button or swipe on the left or right sides of the screen, I expect to be taken one step backward on the webview, not exiting the app or going back to the initial route of the main stack navigation I am using.

I could not get `allowsBackForwardNavigationGestures` to work. I placed it in my Webview and even tried `allowsBackForwardNavigationGestures = {true}`. Still did not work.

And please provide examples with functional components as well as class components.

peterklingelhofer:
This closes #2466.

Here's a commit where this was successfully implemented: https://github.com/peterklingelhofer/stream-app/commit/645547530367a60b73c09aa621b66ee22478eafa.

### DIFF
--- a/docs/Guide.md
+++ b/docs/Guide.md
@@ -547,14 +547,18 @@ We can provide support for conventional page navigation for both Android and iOS
 For iOS, you'll just need to use the [`allowsbackforwardnavigationgestures`](Reference.md#allowsbackforwardnavigationgestures) prop.
 
 For Android, you can use `useRef` and `useEffect` (you'll need to import them from React if you aren't already) to allow users to navigate to the previous page when they press the back button like so:
-```js
+```jsx
 import React, {
     useEffect,
     useRef,
 } from 'react';
+import {
+    BackHandler,
+    Platform,
+} from 'react-native';
 ```
 
-```js
+```jsx
 const webViewRef = useRef(null);
 const onAndroidBackPress = () => {
   if (webViewRef.current) {
@@ -572,6 +576,13 @@ useEffect(() => {
       };
   }
 }, []);
+```
+
+And add this prop to your `WebView` component:
+```jsx
+<WebView
+  ref={webViewRef}
+/>
 ```
 
 ### Hardware Silence Switch
--- a/docs/Guide.md
+++ b/docs/Guide.md
@@ -546,7 +546,7 @@ We can provide support for conventional mobile page navigation: forward/back swi
 
 For iOS, you'll just need to use the [`allowsbackforwardnavigationgestures`](Reference.md#allowsbackforwardnavigationgestures) prop.
 
-For Android, you need to use `BackHandler.addEventListener` and hooking that up to call `goBack` on the `WebView`.
+For Android, you need to use `BackHandler.addEventListener` and hook that up to call `goBack` on the `WebView`.
 
 With functional React components, you can use `useRef` and `useEffect` (you'll need to import them from React if you aren't already) to allow users to navigate to the previous page when they press the back button like so:
 ```jsx
--- a/docs/Guide.md
+++ b/docs/Guide.md
@@ -542,11 +542,13 @@ const App = () => {
 Note that these cookies will only be sent on the first request unless you use the technique above for [setting custom headers on each page load](#Setting-Custom-Headers).
 
 ### Page navigation gesture and button support
-We can provide support for conventional page navigation for both Android and iOS.
+We can provide support for conventional mobile page navigation: forward/back swipe gestures on iOS and the hardware back button/gesture on Android.
 
 For iOS, you'll just need to use the [`allowsbackforwardnavigationgestures`](Reference.md#allowsbackforwardnavigationgestures) prop.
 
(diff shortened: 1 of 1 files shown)


docs/Guide.md
```diff
@@ -540,6 +541,50 @@ const App = () => {
 
 Note that these cookies will only be sent on the first request unless you use the technique above for [setting custom headers on each page load](#Setting-Custom-Headers).
 
+### Page navigation gesture and button support
+We can provide support for conventional page navigation for both Android and iOS.
```
TheAlmightyBob, line 545: It would be nice to specify here that you're talking about the forward/back swipe gestures on iOS and the hardware back button/gesture on Android.

docs/Guide.md
```diff
@@ -540,6 +541,50 @@ const App = () => {
...
+
+For iOS, you'll just need to use the [`allowsbackforwardnavigationgestures`](Reference.md#allowsbackforwardnavigationgestures) prop.
+
+For Android, you can use `useRef` and `useEffect` (you'll need to import them from React if you aren't already) to allow users to navigate to the previous page when they press the back button like so:
```
TheAlmightyBob, line 549: This is only true if you're using functional components and hooks. And I'm totally fine with only providing an example for hooks, I recognize that's the direction things are moving. But I think it would be worth highlighting that the _key_ parts of this solution are in using `BackHandler.addEventListener` and hooking that up to call `goBack` on the WebView.

TheAlmightyBob (changes requested):
Thanks for contributing! Just a couple small comments.

docs/Guide.md
```diff
@@ -542,11 +542,13 @@ const App = () => {
...
 ### Page navigation gesture and button support
-We can provide support for conventional page navigation for both Android and iOS.
+We can provide support for conventional mobile page navigation: forward/back swipe gestures on iOS and the hardware back button/gesture on Android.
 
 For iOS, you'll just need to use the [`allowsbackforwardnavigationgestures`](Reference.md#allowsbackforwardnavigationgestures) prop.
 
-For Android, you can use `useRef` and `useEffect` (you'll need to import them from React if you aren't already) to allow users to navigate to the previous page when they press the back button like so:
+The key parts of this solution are in using `BackHandler.addEventListener` and hooking that up to call `goBack` on the `WebView`.
+
+For Android (with functional React components), you can use `useRef` and `useEffect` (you'll need to import them from React if you aren't already) to allow users to navigate to the previous page when they press the back button like so:
```
TheAlmightyBob, line 551: Sorry, another small detail, but the way you inserted the "key parts" section, it's not clear that it's referring to Android

Maybe you could say something like "For Android, you need to use `BackHandler.addEventListener` etc etc...", then start the next paragraph, "With functional React components, ..."

Or "The key parts for Android are in using ...."

Or something else that you prefer. 🙂

TheAlmightyBob (changes requested):

docs/Guide.md
```diff
@@ -540,6 +541,52 @@ const App = () => {
 
 Note that these cookies will only be sent on the first request unless you use the technique above for [setting custom headers on each page load](#Setting-Custom-Headers).
 
+### Page navigation gesture and button support
+We can provide support for conventional mobile page navigation: forward/back swipe gestures on iOS and the hardware back button/gesture on Android.
+
+For iOS, you'll just need to use the [`allowsbackforwardnavigationgestures`](Reference.md#allowsbackforwardnavigationgestures) prop.
+
+For Android, you need to use `BackHandler.addEventListener` and hooking that up to call `goBack` on the `WebView`.
```
TheAlmightyBob, line 549:
```suggestion
For Android, you need to use `BackHandler.addEventListener` and hook that up to call `goBack` on the `WebView`.
```

TheAlmightyBob (changes requested):
tiny correction, otherwise looks good

TheAlmightyBob (approved):

TheAlmightyBob:
Hm... unfortunately I have no idea why the Android build on Windows is failing... clearly unrelated to your change or any other react-native-webview changes, and it's fine on Ubuntu.... 

@Titozzz @tido64 don't suppose either of you have any thoughts?

Titozzz:
> Hm... unfortunately I have no idea why the Android build on Windows is failing... clearly unrelated to your change or any other react-native-webview changes, and it's fine on Ubuntu.... 
> 
> @Titozzz @tido64 don't suppose either of you have any thoughts?

@chiaramooney is looking into it

TheAlmightyBob:
Thanks @chiaramooney ! (and @Titozzz for the update 🙂 )

Issue #2431, opened 2022-03-31
Load WebView on initial render


sjransom:
**Bug description:**
I'm trying to preload my WebView in a tab using React Navigations Bottom Tabs Navigator and the prop [lazy](https://reactnavigation.org/docs/bottom-tab-navigator/#lazy) set to `false` so that the screen is loaded on initial render.

Unfortunately while the screen renders the actual WebView component does not until the tab is active, this appears to only affect iOS.

**To Reproduce:**
Set `lazy: false` to screenOptions prop of Tab.navigator or options prop of Tab.Screen and render WebView as a component on the screen.

**Expected behavior:**
WebView component should be loaded on initial render.

**Environment:**
 - OS: iOS
 - OS version: 15.4
 - react-native version: 0.67.4
 - react-native-webview version: 11.17.2

If there are other known working solutions to preload WebView please let me know.

walterholohan:
We are experiencing the same issue on our React Native App which is causing a terrible user experience. When a user switches to a tab using the bottom tab bar there have to wait nearly 5 seconds for the Webview to load. If the above issue was fixed it would be amazing as the Webview would already be loaded when the user lands on that tab.

yunjeongloper:
+1

jinee525:
:+1:

sunju0123:
I am also having the same problem. It's very bad for the user experience.

dom-runna:
Also the same would be amazing if the team would be able to fix please!

gbhrdt:
I was able to achieve preloading with the here mentioned patch (using NPM patch-package): https://github.com/react-native-webview/react-native-webview/pull/1852#issuecomment-1120470181

walterholohan:
@gbhrdt do you know if the patch has been merged into master yet?

finalight:
is this still in progress?

walterholohan:
@finalight I believe @gbhrdt created a patch for it https://github.com/react-native-webview/react-native-webview/pull/1852#issuecomment-1120470181 however did not create a PR. @gbhrdt does your solution work? If so can you raise a PR?

walterholohan:
@finalight a PR was just raised with the fix, we are currently using this fix in our production app and it works great

https://github.com/react-native-webview/react-native-webview/pull/2626

ivona314:
Eagerly waiting for this to be released

gronxb:
[...]

PR #1852, opened 2021-01-30
fix(iOS): enable loading WebView even when off screen

goleary:
This is a revival of  #813 (thanks @philippeauriach)

Requests for this functionality:
#696 
#971 

I must admit that I'm developing on Windows so I haven't tested this...I just repeated the commit from #813 in the latest of this repo hoping that CI would help test and/or a generous Mac user could help verify this works as expected.

PR #2626, opened 2022-08-05
iOS: Fix webview not loading in background tab

sjransom:
Initial issue here: https://github.com/react-native-webview/react-native-webview/issues/2431#issuecomment-1203646983
Fix reference from: https://github.com/react-native-webview/react-native-webview/pull/1852#issuecomment-1120470181

This fixes the issue where the web view is not loaded during initial app render when it is on a separate tab.

Testing using:
`"react-native": "0.68.2"`
`"react-native-webview": "11.21.1"`

Thanks to @gbhrdt and @goleary

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -402,9 +402,9 @@ RCTAutoInsetsProtocol>
   return wkWebViewConfig;
 }
 
-- (void)didMoveToWindow
+- (void)didMoveToSuperview
 {
-  if (self.window != nil && _webView == nil) {
+  if (_webView == nil) {
     WKWebViewConfiguration *wkWebViewConfig = [self setUpWkWebViewConfig];
 #if !TARGET_OS_OSX
     _webView = [[WKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];


walterholohan (approved):

finalight:
So what else do we need to move forward with this ?

walterholohan:
@finalight we need a maintainer to review

@Titozzz @jamonholmgren would you be able to review please?

TheAlmightyBob:
This makes sense _to me_.... it seems _potentially_ breaking, but I can't think of a scenario where that would matter... I suppose if folks didn't want their WebViews to load on background tabs they wouldn't set `lazy=false`... and `removeFromSuperview` is already used so there's some symmetry there...

Was `didMoveToWindow` not being called, or was the problem just that `self.window` was `nil`?

I'm a little hesitant only because the previous logic seems like it was there for a reason (i.e. the fact that it's both using `didMoveToWindow` _and_ checking `self.window`). @jamonholmgren I don't suppose you recall why it was done that way?

owen12ab:
Any update on this?

jamonholmgren:
Still on our list to review. Thanks for your patience!

Faradey27:
Small note about a potential issue that we noticed.
In case if render webview in the background and webview contains  that uses srcSet  - sometimes img will not be rendered

dvlkv:
@jamonholmgren
Hello, is there any activity on this or it is considered as stale?

TheAlmightyBob:
This needs to be updated for compatibility with the new architecture changes. It should be pretty simple though, as it was a small change and the issue is just that the file was renamed, but it's not something that GitHub can resolve.

Titozzz:
Closing in favor of #2930

TheAlmightyBob:
Oh, thanks @Titozzz , I totally missed that there was an updated version.

PR #2635, opened 2022-08-11
feat: New arch Android

Titozzz:


src/WebViewTypes.ts
```diff
@@ -251,7 +247,7 @@ export interface WebViewNativeConfig {
   component?:
     | typeof NativeWebViewIOS
     | typeof NativeWebViewMacOS
-    | typeof NativeWebViewAndroid;
+    | typeof NativeWebViewComponent;
```
AndreiCalazans, line 250: Won't Fabric force normalization to a single interface? Are you keeping iOS and MacOS until it's fully migrated?

Titozzz:
> Excuse me, is there any new progress on this. We urgently need it now

If you are in a hurry, either:
- ask what can do done to help on the code
- sponsor people working on that code
- don't use fabric right now cause it's still very early

TheAlmightyBob:
@Titozzz Are there particular changes to pay attention to here? For example, it's hard to tell if the native code changed other than being extracted into separate files for the different classes (which certainly makes sense and I wonder if that should just be its own PR...)

Do you expect this to be a breaking change (e.g. require a new version of RN or enabling the new architecture flags in apps that want to use the latest version of `react-native-webview`)?

Titozzz:
@TheAlmightyBob it won't be breaking. I'm trying to get the new arch to work the same way as the old arch, but old arch users shouldn't be afraid 😅

alfonsocj:
@Titozzz do you have an estimate on when you will land this? I want to give supporting fabric on iOS a try but I want to leverage your codegen changes 😄

Titozzz:
I've already done iOS 😅, you can find my branch new-arch-ios. It only needs polishing with view lifecycle

alfonsocj:
Nice! Is there anything I can do to help you polish it?

kartikmittal1996:
> _ No description provided. _

Can I use this branch in my project with new architecture and fabric enabled?

Titozzz:
I wouldn't use it in production yet as it's not released, but you can try, also I'd start from the iOS branch that contains android and has fixes

Titozzz:
Closed in favor of https://github.com/react-native-webview/react-native-webview/pull/2686 containing both platforms

Issue #2087, opened 2021-07-09
How to open popup In App


KYEOP:
When I open a popup it opens in an external browser (window.open or _blank)

Can't I open it from within the app using onCreateWindow like in Android Native?

By the way, I don't want to use the InAppBrowser library.

react-nativr-webview ver : 11.2.4

Issue #1855, opened 2021-01-31
Links with target="_blank" do not trigger onShouldStartLoadWithRequest


rpander93:
**Bug description:**

On Android, <a /> elements with target="_blank" do not trigger onShouldStartLoadWithRequest. Absolute links will immediately open in the browser. Relative links don't do anything at all (unless you provide a baseUrl, in which case they will be immediately opened in the browser).

**To Reproduce:**
Following example code does not call onShouldStartLoadWithRequest for links with blank target.

```ts
import * as React from "react";
import { SafeAreaView } from "react-native";
import WebView from "react-native-webview";
import { ShouldStartLoadRequest } from "react-native-webview/lib/WebViewTypes";

export default function App() {
  const htmlContent = `
    <!DOCTYPE html>
    <html>
      <head>
        <style type="text/css">
          body {
            display: flex;
            flex-direction: column;
          }
        </style>
        <meta http-equiv="content-type" content="text/html; charset=utf-8">
        <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no" />
      </head>
      <body>
        <a href="https://github.com">Absolute link</a>
        <a href="https://github.com" target="_blank">Absolute link with blank target</a>
        <a href="/react-native-webview">Relative link</a>
        <a href="/react-native-webview" target="_blank">Relative link with blank target</a>
      </body>
    </html>
  `;

  const handleShoudStartLoadingWithRequest = (request: ShouldStartLoadRequest) => {
    console.log("shouldStartLoadWithRequest", request);

    return false;
  };

  return (
    <SafeAreaView style={{ flex: 1 }}>
      <WebView
        originWhitelist={["*"]}
        onShouldStartLoadWithRequest={handleShoudStartLoadingWithRequest}
        source={{ html: htmlContent }}
        style={{ flex: 1, height: 100 }}
      />
    </SafeAreaView>
  );
}
```

I added the following piece of JS on the page, and then it does work:
```html
<script>
  // Need to remove target=_blank on links or onShoudStartLoadingWithRequest doesnt get called on Android
  Array.from(document.querySelectorAll('a[target="_blank"]'))
  .forEach(link => link.removeAttribute('target'));
</script>
```

**Expected behavior:**
I expect onShouldStartLoadWithRequest to be called for absolute/ relative links in all cases, irrespective of their target.

**Environment:**
 - OS: Android
 - OS version: 11
 - react-native version: 0.63.4
[...]

Issue #139, opened 2018-11-15
target = "_blank" href can't open


cuibeihong:

jamonholmgren:
Hi @cuibeihong, thanks for the issue, please provide a minimal reproducible demo so we can test this. Thank you!

meshaabi:
I have the same issue, _blank does nothing. 

```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>Blank HTML5</title>
    <style></style>
  </head>
  <body>
    <p>This page takes you somewhere</p>
      <a href="print.html" target="_blank">TAKE ME SOMEWHERE</a>
    </p>
  </body>
</html>
```

Clicking the link within the Webview does nothing.
onLoadStart fires, onNavigationStateChange fires. and then nothing...

duyleekun:
@jamonholmgren were you able to reproduce this? Any ahref with target="_blank" won't do anything with webview

creativemind1:
> I have the same issue, _blank does nothing.
> 
> ```
> <!DOCTYPE html>
> <html>
>   <head>
>     <meta charset="utf-8" />
>     <title>Blank HTML5</title>
>     <style></style>
>   </head>
>   <body>
>     <p>This page takes you somewhere</p>
>       <a href="print.html" target="_blank">TAKE ME SOMEWHERE</a>
>     </p>
>   </body>
> </html>
> ```
> Clicking the link within the Webview does nothing.
> onLoadStart fires, onNavigationStateChange fires. and then nothing...

For me as well, it doesn't do anything.

kevinjom:
Setting `useWebkit={false}` worked for me.

jamonholmgren:
If someone wants to tackle this problem, this seems like a good place to start: https://stackoverflow.com/a/25853806/204044

timraybould:
Any ideas on how to access the "targetFrame" that is referred to in that stack overflow link? Doesn't seem to be a property on the event object that onNavigationStateChange spits out.

benjosantony:
I tried adding the ios code in https://stackoverflow.com/a/25853806/204044 to the RNCWKWebView.m  in my node_modules just to explore the behaviour and it seems to have fixed the issue (Achieving the same behaviour as UIWebView).

indapublic:
I have `react-native-webview@7.6.0` and my `target=_blank` links not opened in Android.

esynaps:
[...]

PR #2640, opened 2022-08-15
Add `onOpenWindow` event

Ldoppea:
Hi,

This PR is a proposal for handling `target=_blank` links and JS `window.open()` calls that usually ends to opening a new tab/window in a web browser.

With current implementation, those links are handled differently on iOS and Android.

The idea is to introduce a react-native `onOpenWindow` event that would be called when the WebView wants to open a new tab/window.

By doing so, the developper can control what is done with those links and (if needed) to force same behavior on iOS and Android.

This is something I need in my current project as I want to open those links using an InAppBrowser. Also I want to have control on which links I allow to open or not.

The API would be used like this:
```js
<webview
  source={source}
  onOpenWindow={(syntheticEvent) => {
    const { nativeEvent } = syntheticEvent
    const { targetUrl } = nativeEvent
    
    if (isFromMyDomain(targetUrl)) {
      navigateTo(targetUrl)
    } else {
      openInAppBrowser(targetUrl)
    }
  }}
/>
```

If `onOpenWindow` is not defined, then the old behavior is applied

By doing so, I think this would answer some of existing issues that needs to disable opening the browser (https://github.com/react-native-webview/react-native-webview/discussions/2164), open popups in the app (https://github.com/react-native-webview/react-native-webview/issues/2087) or maybe force the same behavior on both Android and iOS (https://github.com/react-native-webview/react-native-webview/issues/1855 and https://github.com/react-native-webview/react-native-webview/issues/139).

## Old behavior

- On Android
  - `<a href="https://example.com" target="_blank" />` from DOM -> opens the device's browser
  - `window.open('https://example.com', '_blank')` from JS -> opens the device's browser
- On iOS
  - `<a href="https://example.com" target="_blank" />` from DOM -> opens the link in the current webview
  - `window.open('https://example.com', '_blank')` from JS -> opens the link in the current webview

## New behavior

- On Android
  - `<a href="https://example.com" target="_blank" />` from DOM -> calls `onOpenWindow` if defined, otherwise opens the device's browser
  - `window.open('https://example.com', '_blank')` from JS -> calls `onOpenWindow` if defined, otherwise opens the device's browser
- On iOS
  - `<a href="https://example.com" target="_blank" />` from DOM -> calls `onOpenWindow` if defined, otherwise opens the link in the current webview
  - `window.open('https://example.com', '_blank')` from JS -> calls `onOpenWindow` if defined, otherwise opens the link in the current webview

## Example

I wrote an example in `example/examples/OpenWindow.tsx`. It is accessible from `OPENWINDOW` tab.

In this example you can compare behaviors with and without the `onOpenWindow` event by toggling the `Intercept OpenWindow event` switch.

In the WebView there is a button for calling `window.open()` and a link for testing both scenario.

On iOS, because the old behavior does a navigation, you can use the `RESET WEBVIEW` button to go back to the example page.

### target="_blank" on iOS

On https://github.com/react-native-webview/react-native-webview/commit/1403f01680cb6ced37a5ff16fe717731fb9e6bc6 I wrote specific code to handle `target="_blank"` links from `onShouldStartLoadWithRequest`.

This is because on my project I have a hook that react to the WebView `ref` changes. I noticed that in [createWebViewWithConfiguration](https://github.com/react-native-webview/react-native-webview/blob/master/apple/RNCWebView.m#L310-L316) the WebView's `ref` would be updated with the new URL whatever is done. So we should intercept `target="_blank"` links before this method is called (this worked fine with `window.open()`).

### `hasTargetFrame` on iOS

On iOS I added the `hasTargetFrame` boolean to `onShouldStartLoadWithRequest` parameters. I used this value to detect new tab/window links on iOS and I thought it was a good thing to expose it also in `onShouldStartLoadWithRequest`: https://github.com/react-native-webview/react-native-webview/pull/2640/commits/34375e9e82712b9d6a9d9445485f386b70e848a7

### Event registration detection on Android

In https://github.com/react-native-webview/react-native-webview/commit/6c3d521fa07ff4f182bbbe9fffb26e69819cc2e1 I tried to call the event only if defined on react-native props. I'm not sure this is the correct way, but I don't know how to do this without using a dedicated prop.

Do you know a better solution?

~~Also this does not update after first initialisation as `onDropViewInstance` is called on Android after the view is initialized. So `mWebChromeClient` becomes null and I then I don't know how to update it. I don't need this feature so I did not investigate much (note that the example works because I reset the WebView by changing its `key` prop)~~
fixed thanks to @TheAlmightyBob on [this comment](https://github.com/react-native-webview/react-native-webview/pull/2640#issuecomment-1237252073)

Note that I tried implementing an event with callback, but I failed to handle the callback from the `RNCWebChromeClient`.

### Event registration detection on iOS

I did the same on iOS in https://github.com/react-native-webview/react-native-webview/commit/3ecf879e6d43708f3f44f953d4244a9f21607ef3 but here I didn't need to use a dedicated prop.

In my project (rn `0.66.4`), I don't need this commit to detect the event registration (the `if (_onOpenWindow)` code works from scratch), but on your Example projet I need to. I don't know why the behavior is different on both project. But the version with the latest commit works on both, so I suppose it costs nothing to add this commit?

### iOS build instructions

I fixed the documentation about iOS build instruction as the previous version did not work on my computer. Tell me if you want me to extract this commit into another PR

### macOS and Windows implementation

I did not implement this event on macOS and Windows because I don't need them and it would take me a bit of time to install environment for those platforms + I don't know if my solution would be accepted by your team.

But if you think my PR is eligible for a merge and if you want me to implement those platforms, I can make a try.


whoisdominic (approved):

jackkinsella:
I'm very excited about this PR -- I think it's gonna be a great addition to the library, especially for people building hybrid web/React Native apps.

allan-simon:
Excited too, if we can do anything to help this move forward , let us know :)

l1ttps (approved):
nice

allan-simon:
@Ldoppea  can you rebase ?

Ldoppea:
> @Ldoppea can you rebase ?

Hi, I just did it, thanks for the reminder 👍

Saif-MindCoopers:
I'm too excited for this PR.
My work involves hybrid apps.
I'm in a tough spot and could really have used this feature.  `onOpenWindow` 🔥

docs/Reference.md
```diff
@@ -751,6 +783,7 @@ lockIdentifier
 mainDocumentURL (iOS only)
 navigationType (iOS only)
 isTopFrame (iOS only)
+hasTargetFrame (iOS only)
```
TheAlmightyBob, line 786: I'm wondering if perhaps this could be explained in the doc a little bit? I realize the other props aren't well-explained either... but `hasTargetFrame` seems to me like it could be especially confusing... for example, I assume that it would always be `true` if a handler is provided for `onOpenWindow`, yes?

example/examples/OpenWindow.tsx
```diff
@@ -0,0 +1,117 @@
+import React, {Component} from 'react';
+import {Button, Switch, StyleSheet, Text, View} from 'react-native';
+
+import WebView from 'react-native-webview';
+
+const HTML = `
+<!DOCTYPE html>\n
+<html>
+  <head>
+    <title>Alerts</title>
```
TheAlmightyBob, line 10: Possibly should be changed to something like "Open Window test"? (though I don't recall offhand if the title is actually ever visible when running the example app..)

example/examples/OpenWindow.tsx
```diff
@@ -0,0 +1,117 @@
...
+`;
+
+type Props = {};
+type State = {
+  shouldInterceptOpenWindow: boolean,
+  text: string,
+  webViewKey: number
+};
+
+export default class Alerts extends Component<Props, State> {
```
TheAlmightyBob, line 45: This class name should be updated 🙂

src/WebView.ios.tsx
```diff
@@ -182,6 +184,7 @@ const WebViewComponent = forwardRef<{}, IOSWebViewProps>(({
       onLoadingStart={onLoadingStart}
       onHttpError={onHttpError}
       onMessage={onMessage}
+      onOpenWindow={onOpenWindowProp && onOpenWindow}
```
TheAlmightyBob, line 187: why isn't `onOpenWindow={onOpenWindow}` enough?

src/WebViewTypes.ts
```diff
@@ -646,6 +655,16 @@ export interface IOSWebViewProps extends WebViewSharedProps {
...
 
+  /**
+   * Function that is invoked when the `WebView` should open a new window.
+   * 
+   * This happens when the JS calls `window.open('http://someurl', '_blank')`
+   * or when the user clicks on a `<a href="http://someurl" target="_blank"` link.
+   *
+   * @platform ios
+   */
+  onOpenWindow?: (event: WebViewOpenWindowEvent) => void;
```
TheAlmightyBob, line 666: Can you define this in `WebViewSharedProps` instead of having duplicate versions in `IOSWebViewProps`/`AndroidWebViewProps`?

TheAlmightyBob (changes requested):
Just a first pass review but in general I like what I'm seeing thus far. Thank you in particular for the detailed writeup, example app update, and making the behavior opt-in.

TheAlmightyBob:
> this does not update after first initialisation as onDropViewInstance is called on Android after the view is initialized. So mWebChromeClient becomes null and I then I don't know how to update it

This is very curious to me and I wonder if it's a bug? I would expect other functionality to be broken if `onDropViewInstance` is being called right after the view is initialized...??

That said, possibly you could work around it by using `WebView.getWebChromeClient` instead of directly accessing `mWebChromeClient`?

TheAlmightyBob:
I've also been pondering whether "real" multi-window support could be added (such that the caller of `window.open` would get a proper reference to the opened window and the opened window could use `window.opener`), possibly leveraging some of the same ideas as in https://github.com/react-native-webview/react-native-webview/pull/2469 for managing multiple WebViews in a dictionary....

But of course that's much more complex and I think there would still be use cases that prefer what you've got here. 🙂

docs/Reference.md
```diff
@@ -575,6 +576,37 @@ url
 
 ---
 
+### `onOpenWindow`[⬆](#props-index)<!-- Link generated with jump2header -->
+
+Function that is invoked when the `WebView` should open a new window.
+
+This happens when the JS calls `window.open('http://someurl', '_blank')` or when the user clicks on a `<a href="http://someurl" target="_blank"` link.
```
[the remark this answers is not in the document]
Ldoppea: Not sure if it was intentional or not. Sometimes I do this if the component is supposed to have more props (I could use this instead `<a href="http://someurl" target="_blank" ...>`)

As you asked the question I made the change in https://github.com/react-native-webview/react-native-webview/compare/5e7a71e0ea21bc0d1eeaa492683a306d9fbea576..1b03904b3c44b0e31fa7e7108bae9b30c1af1418. This should be less confusing.

example/examples/OpenWindow.tsx
Ldoppea: It doesn't appear in the app, but it is used as Title in browsers' devtools.

I made the fix in https://github.com/react-native-webview/react-native-webview/compare/5e7a71e0ea21bc0d1eeaa492683a306d9fbea576..1b03904b3c44b0e31fa7e7108bae9b30c1af1418

example/examples/OpenWindow.tsx
Ldoppea: Fixed in https://github.com/react-native-webview/react-native-webview/compare/5e7a71e0ea21bc0d1eeaa492683a306d9fbea576..1b03904b3c44b0e31fa7e7108bae9b30c1af1418

src/WebView.ios.tsx
Ldoppea: At some point I had to use this code so the iOS code would be able to see a null value.

I'm pretty sure I needed this on my initial code based on version `11.18.2`. Cf this PR: https://github.com/cozy/react-native-webview/pull/2 where `onOpenWindow` is always defined by construction:
```js
onOpenWindow = (event: WebViewOpenWindowEvent) => {
    const { onOpenWindow } = this.props;
    if (onOpenWindow) {
      onOpenWindow(event);
    }
  };
```

But since I rebased on master, I adapted the code to the new architecture and now the prop is declared
```
const onOpenWindow = useCallback((event: WebViewOpenWindowEvent) => {
    onOpenWindowProp?.(event);
  }, [onOpenWindowProp]);
```

But I don't remember if I tested to remove it after that refactoring 🤔 
(damn brain that cannot remember details from 3weeks ago)

src/WebViewTypes.ts
Ldoppea: Are you sure? `WebViewSharedProps` would also impact `MacOSWebViewProps` and `WindowsWebViewProps`. But I did not implement this feature on those platforms.

apple/RNCWebView.m
```diff
@@ -1140,14 +1151,24 @@ - (void)                  webView:(WKWebView *)webView
...
-  
+  BOOL hasTargetFrame = navigationAction.targetFrame != nil;
+
+  if (_onOpenWindow && !hasTargetFrame) {
+    NSMutableDictionary<NSString *, id> *event = [self baseEvent];
+    [event addEntriesFromDictionary: @{@"targetUrl": request.URL.absoluteString}];
+    decisionHandler(WKNavigationActionPolicyCancel);
+    _onOpenWindow(event);
+    return;
+  }
```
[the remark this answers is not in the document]
Ldoppea: Yes this is what I observed.

Without this check, `decidePolicyForNavigationAction` was first accepting the navigation, and after that `createWebViewWithConfiguration` was called, which would open the new window. But the webView's `ref` would be updated on the origin webView.

Initial code was:
```c
- (WKWebView *)webView:(WKWebView *)webView createWebViewWithConfiguration:(WKWebViewConfiguration *)configuration forNavigationAction:(WKNavigationAction *)navigationAction windowFeatures:(WKWindowFeatures *)windowFeatures
{
  if (!navigationAction.targetFrame.isMainFrame) {
    [webView loadRequest:navigationAction.request];
  }
  return nil;
}
```
As we converted the `new window` into a navigation to the same webview, then this changing ref "bug" was not visible as we expected it to change.

If you replace the initial code with something like:
```c
- (WKWebView *)webView:(WKWebView *)webView createWebViewWithConfiguration:(WKWebViewConfiguration *)configuration forNavigationAction:(WKNavigationAction *)navigationAction windowFeatures:(WKWindowFeatures *)windowFeatures
{
  return nil;
}
```
Then nothing would happen (no redirection nor new window), but the `ref` would still change.

src/WebView.android.tsx
```diff
@@ -174,6 +175,8 @@ const WebViewComponent = forwardRef<{}, AndroidWebViewProps>(({
     onHttpError={onHttpError}
     onRenderProcessGone={onRenderProcessGone}
     onMessage={onMessage}
+    onOpenWindow={onOpenWindow}
+    hasOnOpenWindowEvent={onOpenWindowProp !== undefined}
```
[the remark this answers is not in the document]
Ldoppea: Are you sure those are event handlers? I don't see any code calling `onContentSizeChange` nor `onScroll`.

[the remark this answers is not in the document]
Ldoppea: Hmm by reading the code this should be the case for `onScroll`. Maybe there are some internal mapping in react-native mappers? Sorry I'm new at creating react-native bridges.

src/WebView.ios.tsx
Ldoppea: I just did the test, I confirm that `onOpenWindow={onOpenWindow}` does not detect correctly if the event is defined or not, `_onOpenWindow` is alwary truthy in `RNCWebView.m`.

With `onOpenWindow={onOpenWindowProp && onOpenWindow}` the detection is done correctly.

Ldoppea:
> > this does not update after first initialisation as onDropViewInstance is called on Android after the view is initialized. So mWebChromeClient becomes null and I then I don't know how to update it
> 
> This is very curious to me and I wonder if it's a bug? I would expect other functionality to be broken if `onDropViewInstance` is being called right after the view is initialized...??
> 
> That said, possibly you could work around it by using `WebView.getWebChromeClient` instead of directly accessing `mWebChromeClient`?

This did the trick. Now the prop is propagated in realtime. Thanks for the tips @TheAlmightyBob  👍

I made the edit here: https://github.com/react-native-webview/react-native-webview/compare/1b03904b3c44b0e31fa7e7108bae9b30c1af1418..857ef17254da7c8c82f583ba0f6736fb36962205

Ldoppea:
> I've also been pondering whether "real" multi-window support could be added (such that the caller of `window.open` would get a proper reference to the opened window and the opened window could use `window.opener`), possibly leveraging some of the same ideas as in #2469 for managing multiple WebViews in a dictionary....
> 
> But of course that's much more complex and I think there would still be use cases that prefer what you've got here. 🙂

Yes, in my project we have multiple webView that interact with each other. When we encounter a link we want to react based on the link type:
- If the link redirects to another app from our ecosystem, then we want to prevent the navigation and open it in our app using a dedicated react-native route
- If the link redirects outside of our ecosystem, then we want to open it in an InAppBrowser

Those two scenario should work regardless of the link target (simple redirection, new tab or new window). That's why a generic multi-window wouldn't fit our need.

apple/RNCWebView.m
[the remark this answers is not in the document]
TheAlmightyBob: Okay. Could some brief code comments be added, just to clarify that relation between the two handlers?

src/WebView.android.tsx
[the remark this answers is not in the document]
TheAlmightyBob: `onScroll` is dispatched [here](https://github.com/react-native-webview/react-native-webview/blob/c8258bc3eea6fc26a3f01b02ed85297ce891523d/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L1762)

It looks a little different from others because `ScrollEvent` is defined by React Native, not by react-native-webview. Same story for `onContentSizeChange`, which is dispatched [here](https://github.com/react-native-webview/react-native-webview/blob/c8258bc3eea6fc26a3f01b02ed85297ce891523d/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L1582)

docs/Reference.md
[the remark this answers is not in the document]
TheAlmightyBob: The change you made was in `WebViewTypes`, not here in `Reference.md`

(it's not a big deal, but it wasn't clear to me if you meant to change this line also)

docs/Reference.md
TheAlmightyBob: > in Practice, if OnOpenWindow is defined, it should always be false because otherwise OnShouldStartLoadWithRequest would not even be called (OnOpenWindow is called instead)

I thought the opposite was the case. `if (_onOpenWindow && !hasTargetFrame) {` looks like would only call `onOpenWindow` if `hasTargetFrame` is `false`, leaving `onShouldStartLoadWithRequest` to be called with `true`...? Am I misreading this?

> It is false if the webView's navigation trigger the current frame

`false` isn't the `target="_blank"` scenario?

> Do you know what is the expected format for documenting this kind of props? I don't find any similar exemple in the existing doc. Should I just add it in a paragraph after the props list?

I think a paragraph after the props list would be a fine way to do it. `onError` appears to have an example for this.

src/WebView.ios.tsx
TheAlmightyBob: Hmmm. That makes sense.... but then confuses me because I would expect that to mean `messagingEnabled` is always true also??

Uh. Hm. Yeah. That appears to be exactly what's happening. So that seems like a bug. Dang.

Thanks for helping identify that.

TheAlmightyBob: (FYI @Titozzz ... I'm not sure if there's an observable regression that would result from `messagingEnabled` always being true, but it does seem unintended)

src/WebViewTypes.ts
TheAlmightyBob: Hm. My thinking was that's what the `@platform` annotations were for (for example, `javascriptEnabled`... although I'm now seeing that one is kind of a lie because it also works on iOS... hm...).

But then there are some properties like `userAgent` that are duplicated. So I guess there's not a strict standard here and I'm okay with this if you prefer it.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -311,6 +312,15 @@ public void setLackPermissionToDownlaodMessage(WebView view, String message) {
     mLackPermissionToDownloadMessage = message;
   }
 
+  @RequiresApi(api = Build.VERSION_CODES.O)
```
TheAlmightyBob, line 315: Huh. That's interesting that `getWebChromeClient` was introduced so long after `setWebChromeClient`... I hadn't noticed that.

I'm a little curious about the `RequiresApi` though... is that necessary/useful?

Since this function is invoked via React Prop, not directly from Java code, it doesn't seem like this would be very effective in preventing use by older Android versions....

apple/RNCWebView.m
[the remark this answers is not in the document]
Ldoppea: If added a comment here: https://github.com/react-native-webview/react-native-webview/compare/09f909649f0b42f85306f47b3b1654c94bd0054e..3ba2f992c1a6349e35796a50f13bcdf6ba80df60

Tell me if it is not clear enough

docs/Reference.md
Ldoppea: You're right, I answered too fast based on my memories but said the inverse of the real behaviour. Sorry for the confusion.

I added the doc here: https://github.com/react-native-webview/react-native-webview/compare/09f909649f0b42f85306f47b3b1654c94bd0054e..3ba2f992c1a6349e35796a50f13bcdf6ba80df60

Tell me if it is not clear enough

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
Ldoppea: This check is enforced by Android studio, other options are to surround the code by an `if` that does the same check, or to add an `ignore` comment.
image

Maybe the `if` version would be more effective based on your assumption?

apple/RNCWebView.m
```diff
@@ -309,8 +310,18 @@ - (BOOL)canPerformAction:(SEL)action withSender:(id)sender
  */
 - (WKWebView *)webView:(WKWebView *)webView createWebViewWithConfiguration:(WKWebViewConfiguration *)configuration forNavigationAction:(WKNavigationAction *)navigationAction windowFeatures:(WKWindowFeatures *)windowFeatures
 {
-  if (!navigationAction.targetFrame.isMainFrame) {
-    [webView loadRequest:navigationAction.request];
+  if (!navigationAction.targetFrame) {
```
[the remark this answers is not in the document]
Ldoppea: I'm not sure about how objective-c handle null values, but based on this documentation it should always be `nil` : https://developer.apple.com/documentation/webkit/wknavigationaction/1401918-targetframe

If just made some checks, the following code would print `targetFrame is nill`
```c
if (!navigationAction.targetFrame) {
  RCTLogWarn(@"targetFrame is nill");
} else if (!navigationAction.targetFrame.isMainFrame) {
  RCTLogWarn(@"isMainFrame is nill");
} 
```

And the following one would print `isMainFrame is nill`
```c
if (!navigationAction.targetFrame.isMainFrame) {
  RCTLogWarn(@"isMainFrame is nill");
} else if (!navigationAction.targetFrame) {
  RCTLogWarn(@"targetFrame is nill");
}
```

Seems like objective-c has natural "optional chaining" like mechanism? 🤔 

However the [original commit](https://github.com/react-native-webview/react-native-webview/commit/6643a8e38d12df8a857ae09839c94ef158e14176) links to this stackoverflow answer: https://stackoverflow.com/a/25853806 that uses `navigationAction.targetFrame.isMainFrame`
But there is another answer https://stackoverflow.com/a/27391215 that reacts to the first one and uses `navigationAction.targetFrame == nil` instead.

So I suppose my new version would be enough?

However my implementation had a useless check in the `else` block (already covered by the surrounding `if`), so I removed it here: https://github.com/react-native-webview/react-native-webview/compare/3ba2f992c1a6349e35796a50f13bcdf6ba80df60..6fe3b597c9107a88855da7b523f3970449ea3adb

src/WebViewTypes.ts
Ldoppea: The only documentation I found about `@platform` is https://reactnative.dev/docs/platform-specific-code

Seems like this is more of a documentation convention by react-native but not a type guard?

docs/Reference.md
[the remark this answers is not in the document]
Ldoppea: If forgot this one sorry. I made the change here: https://github.com/react-native-webview/react-native-webview/compare/09f909649f0b42f85306f47b3b1654c94bd0054e..3ba2f992c1a6349e35796a50f13bcdf6ba80df60

apple/RNCWebView.m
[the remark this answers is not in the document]
TheAlmightyBob: That's pretty good, but the key thing for me, that I would encourage adding, is the fact that preventing the navigation means `createWebViewWithConfiguration` won't be called, so you need to raise `onOpenWindow` directly from `decidePolicyForNavigationAction` instead.

docs/Reference.md
TheAlmightyBob: Looks good, thanks!

src/WebView.ios.tsx
TheAlmightyBob: (opened #2666 based on this conversation)

src/WebViewTypes.ts
TheAlmightyBob: You're correct, but I'm not sure that the separate interfaces provides that much of a type guard either, since they wind up getting merged with

```typescript
export type WebViewProps = IOSWebViewProps & AndroidWebViewProps & WindowsWebViewProps;
```

But I'm okay leaving this.

apple/RNCWebView.m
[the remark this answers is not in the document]
TheAlmightyBob: > Seems like objective-c has natural "optional chaining" like mechanism?

Huh! It seems you are correct. TIL. https://collindonnell.com/2022/04/28/the-brilliance-of-nil-in-objective-c/

> my implementation had a useless check in the else block (already covered by the surrounding if)

Are you _sure_ that was useless? Given that the previous logic checked `isMainFrame` and your surrounding `if` just checks `targetFrame`, wouldn't keeping that `isMainFrame` check in the `else` be an appropriate way to preserve the previous behavior?

And yes I see the other StackOverflow answer that uses `targetFrame == nil` instead of `isMainFrame`, but since that doesn't explain _why_ it makes that change, I'd be inclined to leave the existing logic just to be safe...

This reminds me of another question: your examples in the description use `window.open('https://example.com', '_blank')`. Is the behavior any different if the second parameter isn't `_blank`? (would `targetFrame` still be `nil`?)

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
TheAlmightyBob: I don't feel _super_ strongly about it, but yeah I would lean towards the `if` check, just in case anybody's still supporting older Android versions.

apple/RNCWebView.m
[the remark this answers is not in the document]
Ldoppea: > Are you sure that was useless? Given that the previous logic checked isMainFrame and your surrounding if just checks targetFrame, wouldn't keeping that isMainFrame check in the else be an appropriate way to preserve the previous behavior?

I would say yes. This code (original PR):
```c
  if (!navigationAction.targetFrame) {
    NSURL *url = navigationAction.request.URL;
    if (_onOpenWindow) {
      [event addEntriesFromDictionary: @{@"targetUrl": url.absoluteString}];
      _onOpenWindow(event);
    } else {
      if (!navigationAction.targetFrame.isMainFrame) {
        [webView loadRequest:navigationAction.request];
      }
    }
  }
```

Can be simplified as
```c
  if (!navigationAction.targetFrame) {
    //...
      if (!navigationAction.targetFrame.isMainFrame) {
        [webView loadRequest:navigationAction.request];
      }
  }
```

If we enter the first `if`, then it should be impossible to have something in `targetFrame` in the second `if`. So 2nd `if` should alway be `true`

> Huh! It seems you are correct. TIL. https://collindonnell.com/2022/04/28/the-brilliance-of-nil-in-objective-c/

> but since that doesn't explain why it makes that change, I'd be inclined to leave the existing logic just to be safe...

As your link confirm the `nil` chaining mechanism, I see no cons to add `isMainFrame` back. So lets do it: https://github.com/react-native-webview/react-native-webview/compare/6fe3b597c9107a88855da7b523f3970449ea3adb..e82aed74391b74611b143ccc73cd06d53aebcfde

> This reminds me of another question: your examples in the description use window.open('https://example.com', '_blank'). Is the behavior any different if the second parameter isn't _blank? (would targetFrame still be nil?)

Good question. I completed the tests to check that : https://github.com/react-native-webview/react-native-webview/compare/e82aed74391b74611b143ccc73cd06d53aebcfde..38ba351763ad6ddfab3b002bc227d2d0c19a233d

Result on iOS is:
- no param -> trigger OnNewWindow
- '_blank' -> trigger OnNewWindow
- '_self' -> navigate in same WebView
- '_parent' -> navigate in same WebView
- '_top' -> navigate in same WebView

I did not test yet on Android, I'll try to do it later today.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
Ldoppea: Here is the new version: https://github.com/react-native-webview/react-native-webview/compare/3788ba8aaad601212072a157bd15c714b150615a..a33abfd3d789f0da878ee9509b6644cd75bea295

apple/RNCWebView.m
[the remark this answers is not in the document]
Ldoppea: I'm made the test on Android, same behaviour 👍

apple/RNCWebView.m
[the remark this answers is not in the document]
Ldoppea: New proposal: https://github.com/react-native-webview/react-native-webview/compare/a33abfd3d789f0da878ee9509b6644cd75bea295..89189a4b26632a3f067fa07386b09d395b2e7ede

apple/RNCWebView.m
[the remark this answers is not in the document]
TheAlmightyBob: > If we enter the first if, then it should be impossible to have something in targetFrame in the second if.

🤦 
Of course. Sorry.

> Result on iOS is:
> no param -> trigger OnNewWindow
'_blank' -> trigger OnNewWindow
'_self' -> navigate in same WebView
'_parent' -> navigate in same WebView
'_top' -> navigate in same WebView

Thanks for checking. Is that the behavior you _want_? I'm not sure if those are special values that are _supposed_ refer to the current window, but I know that this parameter could instead be a name (e.g. "MySpecialPopup"), in which case a _browser_ would be expected to open the link in a new window, not navigate in the current one.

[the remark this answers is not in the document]
TheAlmightyBob: Okay, I pulled your branch and ran my own test using a custom name, and it did trigger `onOpenWindow` as expected.

TheAlmightyBob:
@Ldoppea I was checking if this was ready to merge now, but found that it still has an issue with Android and the `WebChromeClient`, so the example wasn't working without using the "reset webview" button.

It appears that in this case the issue is that `setupWebChromeClient` is called a second time via `setAllowsFullscreenVideo`, which recreates the `WebChromeClient` and resets the knowledge about the `onOpenWindow` handler.

That's a bit of a surprise to me and seems silly if the `WebChromeClient` isn't changing.... _however_, it appears that toggling `setAllowsFullscreenVideo` _does_ replace the `WebChromeClient`, I'm guessing that `hasOnOpenWindowEvent` needs to be additionally stored as a member of `RNCWebViewManager` (like `mAllowsFullscreenVideo` is) so it can be reapplied in `setupWebChromeClient`?

Or perhaps it could set a property on the `RNCWebView` instead... (which could send it to the `WebChromeClient` in `setWebChromeClient`... or possibly `onCreateWindow` could just retrieve the property from there instead of having the `WebChromeClient` own a copy....)

Ddv0623:
Hi Guys, I there any version have this feature onOpenWindow yet?

TheAlmightyBob:
@Ddv0623 Unfortunately not.

apple/RNCWebView.m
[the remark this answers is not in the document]
Ldoppea: Hi, sorry, I missed this question so I didn't realize that there were still some verifications to be done. Thank you for the test.

Ldoppea:
> @Ldoppea I was checking if this was ready to merge now, but found that it still has an issue with Android and the `WebChromeClient`, so the example wasn't working without using the "reset webview" button.
> 
> It appears that in this case the issue is that `setupWebChromeClient` is called a second time via `setAllowsFullscreenVideo`, which recreates the `WebChromeClient` and resets the knowledge about the `onOpenWindow` handler.
> 
> That's a bit of a surprise to me and seems silly if the `WebChromeClient` isn't changing.... _however_, it appears that toggling `setAllowsFullscreenVideo` _does_ replace the `WebChromeClient`, I'm guessing that `hasOnOpenWindowEvent` needs to be additionally stored as a member of `RNCWebViewManager` (like `mAllowsFullscreenVideo` is) so it can be reapplied in `setupWebChromeClient`?
> 
> Or perhaps it could set a property on the `RNCWebView` instead... (which could send it to the `WebChromeClient` in `setWebChromeClient`... or possibly `onCreateWindow` could just retrieve the property from there instead of having the `WebChromeClient` own a copy....)

Sorry for the delay. I'll try to work on this soon.

How did you test that? Is setting `allowsFullscreenVideo=true` on `examples/OpenWindow.tsx` enough to reproduce the bug?

TheAlmightyBob:
@Ldoppea 
> How did you test that? Is setting `allowsFullscreenVideo=true` on `examples/OpenWindow.tsx` enough to reproduce the bug?

Ah, sorry, good question, I should have included repro steps. I'm currently in the middle of something else so can't check right now, but I _believe_ I didn't even need to set `allowsFullscreenVideo`... I _think_ it was getting invoked and messing with things just to apply the default value... could be wrong though. But I'm _pretty sure_ that I was just using the example as-is...

Let me know if you're unable to repro.

demedos:
Is there any update on this?

Ldoppea:
> Is there any update on this?

Now I can say: yes 😅

TheAlmightyBob (dismissed):

src/WebViewTypes.ts
```diff
@@ -142,6 +142,10 @@ export interface WebViewRenderProcessGoneDetail {
   didCrash: boolean;
 }
 
+export interface WebViewOpenWindow {
+  targeturl: string;
```
sunnylqm, line 146: should be `targetUrl`

ahollocou:
Also very excited about this PR – any update on this? Thanks!

src/WebViewTypes.ts
TheAlmightyBob:
```suggestion
  targetUrl: string;
```

TheAlmightyBob (dismissed):

cbfranca:
Also very excited about this PR – any update on this? Thanks! [2]

Ldoppea:
Hi, I'm not sure if there are still things to be done on this PR. @TheAlmightyBob are you waiting something from me? Or is it in your hands?

TheAlmightyBob:
@Ldoppea Thanks for checking, but I think it's in my & @Titozzz 's hands right now. I've been holding off for the moment because I think he has a plan for how to merge a few things.

Titozzz:
This is great work with absolute awesome pr description. Please hang tight till I can review / merge this, I've been spending a lot of my bandwidth on new arch/ react native 0.71 but I'll review this hopefully soon. Thanks again

Ldoppea:
Ok, don't worry, take your time. Until the merge I'm working on my own git branch so I'm not impacted by the delay 🙂

Also thank you for maintaining this project, it helps my team a lot 😄

matusca96:
Hey @Ldoppea, amazing contribution to the project! Any chance that I can use your implementation right away? I really need this behavior, if I make a fork or just use patch-package, will work with what it's already implemented?

Ldoppea:
@matusca96 yes you can. In the package.json file you are not limited to `npm.org` packages. You can also target a git repository or, even easier, a github repository.

For example, you can fork this repository, do some code edits on it and then create a tag on the branch/commit you want to install on your project.

Then on your project's package.json replace your `react-native-webview` entry with this:
```json
"react-native-webview": "github:xxx/react-native-webview#1.0.0"
```
Replace xxx by your account or team name where you created the fork. And replace `1.0.0` by your tag name.

TheAlmightyBob:
Hi bot, please keep this alive

saifudheenpm3:
Still waiting for this feature – any update on this? Thanks.

o-alexandrov:
Could you please share your guidance:
- is this feature compatible with the v12 `new architecture` or you expect this PR would need some changes because of `new architecture`-related changes?

TheAlmightyBob:
The feature is conceptually compatible but the PR will need changes because the new architecture support involved a major refactor of the library.

I expect the general approach will still work, but, for example, things are in different files now, and the Android implementation is in Kotlin instead of Java.

Ldoppea:
Hi, I just noticed that you released the new arch rework, congrats 🎉 

I won't have much time this week but I'll try to rebase the PR soon

MolnarRolandSV:
bump

fukemy:
bump bump, Im tired of `onShouldStartLoadWithRequest` never called on Android event `setSupportMultipleWindows={false}`

asurare:
@jamonholmgren @Titozzz Any time to review this PR? This looks like a must have…

o-alexandrov:
@asurare there's nothing to review, you should refer to [this comment above](https://github.com/react-native-webview/react-native-webview/pull/2640#issuecomment-1493186639) from [TheAlmightyBob](https://github.com/TheAlmightyBob)
> The feature is conceptually compatible but the PR will need changes because the new architecture support involved a major refactor of the library.
> I expect the general approach will still work, but, for example, things are in different files now, and the Android implementation is in Kotlin instead of Java.

Ldoppea:
Hi,

Sorry for the delay, I lacked of time last month and then I forgot about this.

I just tried to do a rebase. It was easier than I expected, however did not do all the tests yet.

I'm not confident with the new arch part. I tried to implement both old and new archs, but I'm not sure which one is tested in the `example` project. Help appreciated here if someone can double check that part.

Also the implementation changed a lot on Android with the new `RNCWebChromeClient.java` and `RNCWebViewManagerImpl.kt`. I tried to adapt the old code to this, I blindly trusted the java typing errors and seems like it did the trick.
I checked that the issue from [this comment](https://github.com/react-native-webview/react-native-webview/pull/2640#issuecomment-1260189496) was still fixed, and it seems to be ok.

I have a lot of lint errors on my computer (including code that is not mine), seems like I'm missing something. So I may have missed some typing declaration. I'll check that later.

Here is the original PR content for comparison: https://github.com/react-native-webview/react-native-webview/compare/master...Ldoppea:react-native-webview:feat/on_open_window_backup_before_new_arch?expand=1

Ldoppea:
PS: on my initial PR I bulletproofed the implementation I my company's project, but today we are still using an old version of this lib so I won't be able to test this on our project until we chose to do the upgrade. So for now I just tested using the `example` project. So here again, help would be appreciated if some of you want to test this PR on their project.

android/src/main/java/com/reactnativecommunity/webview/RNCWebChromeClient.java
```diff
@@ -85,6 +89,24 @@ public RNCWebChromeClient(RNCWebView webView) {
...
+            newWebView.setWebViewClient(new WebViewClient(){
+            @Override
+            public boolean shouldOverrideUrlLoading (WebView subview, String url) {
+                WritableMap event = Arguments.createMap();
+                event.putString("targetUrl", url);
+
+                ((RNCWebView) view).dispatchEvent(
+                view,
+                new TopOpenWindowEvent(view.getId(), event)
+                );
```
TheAlmightyBob, line 103: small nit: indentation
```suggestion
                ((RNCWebView) view).dispatchEvent(
                    view,
                    new TopOpenWindowEvent(view.getId(), event)
                );
```

I hadn't noticed until now that the indentation changed from 2-space to 4-space with the new architecture refactor... not certain if that was intentional... but might as well be consistent in this case.

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -40,6 +40,7 @@
 import com.reactnativecommunity.webview.events.TopCustomMenuSelectionEvent;
 import com.reactnativecommunity.webview.events.TopLoadingProgressEvent;
 import com.reactnativecommunity.webview.events.TopMessageEvent;
+import com.reactnativecommunity.webview.events.TopOpenWindowEvent;
```
TheAlmightyBob, line 43: Can this import be removed now?

android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -199,6 +202,11 @@ public void setLackPermissionToDownloadMessage(RNCWebView view, @Nullable String
         mRNCWebViewManagerImpl.setLackPermissionToDownloadMessage(value);
     }
 
+    @ReactProp(name = "hasOnOpenWindowEvent")
+    public void setHasOnOpenWindowEvent(RNCWebView view, boolean hasEvent) {
```
TheAlmightyBob, line 206: I'm not an expert on the new architecture yet either, but I notice that *most* of the other props here have the `@Override` attribute...

TheAlmightyBob (changes requested):
Thank you for including the link to the previous version for comparison, that was indeed very helpful 😄 

I haven't done any testing yet, but the code looks good, just a couple minor comments

android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
Ldoppea: Fixed in: https://github.com/react-native-webview/react-native-webview/compare/5a835821371ba93f5a5ac120c88097fc3c918ff4..610aee3309dfd69739c28a5bb4579b91015d0394

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
Ldoppea: Fixed in: https://github.com/react-native-webview/react-native-webview/compare/5a835821371ba93f5a5ac120c88097fc3c918ff4..610aee3309dfd69739c28a5bb4579b91015d0394

android/src/main/java/com/reactnativecommunity/webview/RNCWebChromeClient.java
Ldoppea: Fixed in: https://github.com/react-native-webview/react-native-webview/compare/5a835821371ba93f5a5ac120c88097fc3c918ff4..610aee3309dfd69739c28a5bb4579b91015d0394

asurare:
@TheAlmightyBob Is it okay to be merged now ?

TheAlmightyBob:
I tested on Android new/old architecture without issue.

On iOS, with the old architecture, I see an occasional crash when trying to use the OpenWindow test. However, I was able to repro that crash on the main branch (by copying just the basic test over, minus the `openWindow` part)... so I think it has more to do with the test than with the actual library change.

On iOS, with the new architecture, I'm not getting the WebView to load at all in the test app. But again, this happens on the main branch too. @Titozzz is this familiar at all? Did you try building the test app with the new architecture on iOS?

(I have not tried a "real" app with the new architecture on iOS...)

TheAlmightyBob (dismissed):

TheAlmightyBob:
I tried creating a brand new React Native 0.72 app using the new architecture and:
- It still had the same issue of the WebView not rendering after the first render (like, when the app first loads, it's there... but if you re-render for any reason, it disappears??)
- The `openWindow` callback didn't work

Went back and tried the test app with the new architecture again, this time making the OpenWindow test the default test (since, as I said, the WebView does _initially_ render, it just disappears if you try to go to a different test or reload), and the `openWindow` callback didn't work there either. I do not know why, have not tried debugging.

It's curious to me that I haven't seen more reports of issues with the new architecture... though I do see #3003, but not sure if that's related...

TheAlmightyBob:
Okay, so:
1. The failure to render the WebView on the iOS New Architecture, #3003, and _I believe even the crashes on the old architecture_ all appear to be a regression due to #2930. Not related to this PR. (that previous PR was a small change so it was easy to locally revert and get the tests working again)
2. However, with that out of the way, this still doesn't work on the iOS New Architecture yet. Looks like additional changes are required to `RNCWebView.mm`... this whole file is specific to the new architecture, and there is some boilerplate required there to hook up events now.

TheAlmightyBob:
@Ldoppea I just added a commit that should get this working with the iOS new architecture and welcome your review of the change.

TheAlmightyBob (approved):

o-alexandrov:
This PR works with `react-native@0.72.3` and by installing directly (see below).
In case anyone else wants to test in their apps, use the following format for `npm` in your `package.json`:
```json
{
  "dependencies": {
    "react-native-webview": "github:Ldoppea/react-native-webview#667cfec3900560a724c4ad111d4bdbb76f7eba14"
  }
}
```

o-alexandrov:
@Ldoppea @TheAlmightyBob Sorry for the ping and being just a tester.
Please be informed, when `window.open` is triggered in an async function w/ a pause, `onOpenWindow` in ReactNative won't be triggered.

## `onOpenWindow` is triggered in ReactNative

```js
// your React app
const Component = () => {
  // "async" directive doesn't break, it still works. It will break in the next case when there's an "await"
  const openWindow = async () => {
    window.open(`https://youtu.be`)
  }
  return <div onClick={openWindow} style={{ width: 100, height: 100, background: "red" }} />
}
```

## `onOpenWindow` is NOT triggered

```js
// your React app
const Component = () => {
  const openWindow = async () => {
    const pauseThread = () => new Promise((resolve) => setTimeout(resolve, 1000))
    await pauseThread()
    window.open(`https://youtu.be`)
  }
  return <div onClick={openWindow} style={{ width: 100, height: 100, background: "red" }} />
}
```

TheAlmightyBob:
@o-alexandrov Are you seeing this on iOS/Android/both?

Does anything happen?

> just a tester

Testing open PRs before they get merged is _extremely_ valuable and appreciated.

Ldoppea:
Hi,

I think this is a correct behaviour. In classic browsers, `window.open()` should be called synchronously as they want to [limit the opening of popups to being in direct response to user input](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#opening_a_popup).

In reallity you can open them asynchronously but there is some "maximum delay" that should occur between user input and `window.open()` call. However there seems to be no official number about that delay.

I suppose that this behaviour is the same for webviews?

More details about this limitation in the MSDN documentation:

Ldoppea:
However I'm wondering if we can have control on this using native webview APIs 🤔

o-alexandrov:
@TheAlmightyBob I've tested on `iOS` only for now, but will test on `Android` also.
`onOpenWindow` isn't triggered at all (added a console statement in ReactNative in the beginning of the function).

@Ldoppea I agree with your points.
However, I think we should consider the behavior in `react-native-webview` as a bug, if it is not the same as in the browser.
Created a [minimal repro repo](https://github.com/o-alexandrov/test-window-open-w-delay/blob/main/index.html) for you, so you could confirm even 4 seconds delay in the browser is fine.


https://github.com/react-native-webview/react-native-webview/assets/9992724/38b3bb28-411a-4f67-8b30-11ff7c848b92

o-alexandrov:
@TheAlmightyBob @Ldoppea [this issue](https://github.com/react-native-webview/react-native-webview/pull/2640#issuecomment-1645121101) (delay below `window.open` makes `onOpenWindow` not to trigger), happens on `iOS` only. If you'd like to, I could record a screencast.
- the exact same code described in [this comment](https://github.com/react-native-webview/react-native-webview/pull/2640#issuecomment-1645121101) works on Android, but doesn't on iOS

TheAlmightyBob:
Thank you both for looking into that and following up

TheAlmightyBob:
@Ldoppea Thanks for all the work and patience on this one!

Ldoppea:
> @Ldoppea Thanks for all the work and patience on this one!

Ohh! This GH notification made my day 🎉
 
Thank you too for all your efforts maintaining this lib 🙂

jayon-niravel:
I tested with the latest release but could not open the window inside webview (using android). Can you suggest what I am doing wrong here?

`        return <WebView
            source={{
                uri: `some_randome_url`
            }}
            javaScriptCanOpenWindowsAutomatically={true}
            originWhitelist={['*']}
            onOpenWindow={(syntheticEvent) => {
            const { nativeEvent } = syntheticEvent;
            const { targetUrl } = nativeEvent
            console.log('Intercepted OpenWindow for', targetUrl)
            }}
        />;`
        
        
        The URL "some_randome_url" have below line in the html
        `<a href="https://www.google.com" target="_blank">testing link</a>`
        
        
        Clicking on above open it on external window.

Titozzz:
Huge thanks to @TheAlmightyBob for taking the time to review / merge

andophine:
Hi,  if the target url in window.open is empty, onOpenWindow event won't be triggered. This is the case, need this feature for writing data to this new window. Thanks

TheAlmightyBob:
How would you write data to the window if there's no URL open in it...?

andophine:
> How would you write data to the window if there's no URL open in it...?

An example, open the window and use javascript to write data. 
 
```
function openQR(){  
    mywin = window.open('', "target='_blank', resizable=yes, scrollbars=yes, toolbar=yes, menubar=yes, location=no, directories=no, status=yes")  
    mywin.document.write("<p>dynamic QR Code</p>")  
}  
```
Full html example below
```
<!DOCTYPE html>
<html>
<head>
    <title>New Window Example</title>
</head>
<body>
    <button onclick="openNewWindow()">Open google</button>
    <p></p>
	<button onclick="openQR()">Open QR</button>
    <script>
        function openNewWindow() {
            window.open("https://www.google.com", "_blank");
        }
        function openQR(){
            mywin = window.open('', "target='_blank', resizable=yes, scrollbars=yes, toolbar=yes, menubar=yes, location=no, directories=no, status=yes")
            mywin.document.write("<p>dynamic QR Code</p>")
        }
    </script>
</body>
</html>
```

TheAlmightyBob:
@andophine Sorry been busy and didn't see your response. Unfortunately, the flow you describe is not supported by this feature. That would require a completely different and much more complex implementation (which I would like to do but is dependent upon https://github.com/react-native-webview/react-native-webview/pull/2469).

PR #2659, opened 2022-08-30
function name correction

mayur124:


TheAlmightyBob:
@mayur124 The existing name looks correct to me. While `injectedJavaScriptBeforeContentLoaded` is _also_ mentioned above, it's just a slight variation of `injectedJavaScript`, and I believe this sentence was intentionally referring to the latter.

PR #2662, opened 2022-09-01
add null check safety for adding mainDocumentURL to event

HudiDev:
There are cases - albeit seldom - where mainDocumentURL in request is nil hence causing a crash when attempting to add it to the event object i.e [event addEntriesFromDictionary]. 
Here is the crash we were encountering: 

```
Fatal Exception: NSInvalidArgumentException
*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[1]
```

This PR aims to solve the issue.

Related issue: [2442](https://github.com/react-native-webview/react-native-webview/pull/2442)

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -1143,9 +1143,13 @@ RCTAutoInsetsProtocol>
   
   if (_onShouldStartLoadWithRequest) {
     NSMutableDictionary<NSString *, id> *event = [self baseEvent];
+    if (request.mainDocumentURL) {
+      [event addEntriesFromDictionary: @{
+        @"mainDocumentURL": (request.mainDocumentURL).absoluteString,
+      }];
+    }
     [event addEntriesFromDictionary: @{
       @"url": (request.URL).absoluteString,
-      @"mainDocumentURL": (request.mainDocumentURL).absoluteString,
       @"navigationType": navigationTypes[@(navigationType)],
       @"isTopFrame": @(isTopFrame)
     }];


apple/RNCWebView.m
```diff
@@ -1143,9 +1143,13 @@ - (void)                  webView:(WKWebView *)webView
   
   if (_onShouldStartLoadWithRequest) {
     NSMutableDictionary<NSString *, id> *event = [self baseEvent];
+    if (request && request.mainDocumentURL) {
```
TheAlmightyBob, line 1146: There's other logic that would already blow up if `request` itself were `nil`, so I think you could just make this `if (request.mainDocumentURL) {`?

TheAlmightyBob (changes requested):

TheAlmightyBob:
You referenced the related PR, but it would be good to explicitly reference the previous issues as well (even though they had been auto-closed):
https://github.com/react-native-webview/react-native-webview/issues/2414
https://github.com/react-native-webview/react-native-webview/issues/2051

apple/RNCWebView.m
HudiDev: Done

TheAlmightyBob (approved):

HudiDev:
@TheAlmightyBob what are we missing for a merge?

TheAlmightyBob:
@HudiDev Just me having time to check back after the checks ran 🙂

TheAlmightyBob:
@HudiDev Thanks for the contribution!

PR #2666, opened 2022-09-06
fix(js): messagingEnabled prop

TheAlmightyBob:
This was always being set to `true`, even if no `onMessage` handler was provided. At the point that `messagingEnabled` is set, `onMessage` actually refers to a wrapper callback from `WebViewShared`, which is always defined, not the actual `onMessage` handler that was provided (or not) by the application.

This regressed in version 11.22.0

Thanks to @Ldoppea for indirectly bringing this to my attention. 😁


TheAlmightyBob:
@Titozzz this is a very small fix related to recent refactoring that it would be nice to get a quick sanity check on 🙂

Issue #2586, opened 2022-07-05
When native postMessage to web, method of iOS and android is unmatched


zerosrat:
**Bug description:**

When calling `webView.postMessage()`, the dispatchEvent method of iOS and android is differnt as below.
android: `document.dispatchEvent(event)`

https://github.com/react-native-webview/react-native-webview/blob/6b8e4f3f5f3cc006ebca8cb3c7ab692805824685/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L758

iOS: `window.dispatchEvent(event)`

https://github.com/react-native-webview/react-native-webview/blob/a68df8526527a8fa5a21f62eea1d60e956a833db/apple/RNCWebView.m#L909

As a result, developer need to addEventListener on both of `document` and `window` target which is redundant
```
docuemt.addEventListener('message', cb)
window.addEventListener('window', cb)
```

So, is it right to modify android file to change document into window?

tkmin:
https://github.com/react-native-webview/react-native-webview/issues/356#issuecomment-1290912572

Issue #809, opened 2019-08-28
[Documentation] Clarify migration path for this.webView.postMessage removal


luisnaranjo733:
**Is your feature request related to a problem? If so, Please describe.**
Not really. The README on this project very clearly states the following upcoming removal: "this.webView.postMessage() removal (never documented and less flexible than injectJavascript)"

That seems straightforward enough. Stop using this.webView.postMessage() and start using injectJavascript. It _seems_ like anywhere you use postMessage you can just swap to injectJavascript. Would be nice to call that out in the README so it's less scary for people who have code that is using postMessage(). If there are any significant behavioral differences between postMessage and injectJavascript, would definitely be good to call them out so people know about them.

**Describe the solutions you came up with**
Add a few sentences explaining how to replace the usage of postMessage with injectJavascript. Basically, provide a migration path for people who are using the undocumented postMessage

fattomhk:
I wonder how to trigger onMessage without postMessage()

Titozzz:
When you use injectJavascript, the passed string gets evaluated in the browser window so you could use something like: 

```js
const generateOnMessageFunction = (data: string) =>
  `(function() {
    window.whateverYourListenerIsNamed.onMessage(${JSON.stringify(data)});
  })()`;
```

And then call 
```
webViewRef.current.injectJavaScript(
            generateOnMessageFunction(message),
          );
```

Titozzz:
The idea here is that injectJavascript allows for much more possibility, while not removing anything from postMessage

luisnaranjo733:
@Titozzz what is window.whateverYourListenerIsNamed? The docs seem to imply that the injected global variable is always called window.ReactNativeWebView. Is this something different?

Also, the docs talk about window.ReactNativeWebView.postMessage for web -> native communication.
How does that relate to window.whateverYourListenerIsNamed.onMessage?

fattomhk:
```
/**
 * This method is called whenever JavaScript running within the web view calls:
 *   - window.webkit.messageHandlers[MessageHandlerName].postMessage
 */
- (void)userContentController:(WKUserContentController *)userContentController
       didReceiveScriptMessage:(WKScriptMessage *)message
{
  if (_onMessage != nil) {
    NSMutableDictionary<NSString *, id> *event = [self baseEvent];
    [event addEntriesFromDictionary: @{@"data": message.body}];
    _onMessage(event);
  }
}
```
[...]

PR #2684, opened 2022-09-25
fix(android): add window.dispatchEvent when postMessage

zerosrat:
refer: https://github.com/react-native-webview/react-native-webview/issues/2586


TheAlmightyBob:
I'm not sure why the Android implementation used `document` instead of `window`, but `postMessage` has been sort of deprecated for quite some time: https://github.com/react-native-webview/react-native-webview/issues/809

So I'm kind of wondering if the better fix might be to just remove it and force everybody to switch to `injectJavascript`....

PR #2686, opened 2022-09-27
feat: New arch

Titozzz:
This pull request implement Fabric for iOS and Android:

Known bugs that need to be resolved before release:

- [x] Local Files or not working
- [x] Workaround for headers in codegen
- [x] Headers needs a small tweak on JS Side
- [x] Expose src/ in files
- [x] Handle Commands Android
- [x] Handle Commands iOS/macOS


TheAlmightyBob:
Looks like this is also showing all the changes from the `new-arch` branch... perhaps this PR should target that one instead of `master`?

jamninetyfive:
this is great. hope merged quickly.

thiavila-sigma:
@Titozzz thank you so much for this. I'm already using this branch in my project and so far it is running so smoothly that I can't understand why it is not merged yet. 

I'm happy to be your tester and report whatever bug I face (and if possible how I solved it). Regarding this, where would be the best place to report to you?

Titozzz:
As an example, it was only support uri and not local files till this morning. WebView is huge and while 99% of users only need a subset of feature, I do not want to break anything. Yet I'm happy that you can use it 🙂

kartikmittal1996:
Hey @Titozzz I am getting an issue while trying to use this branch. 

`src/WebView.ios.tsx:189:7 - error TS2322: Type 'MutableRefObject<HostComponent<NativeProps> | null>' is not assignable to type 'LegacyRef<Component<NativeProps, {}, any> & Readonly<NativeMethods>> | undefined'`

Can please take a look and help to resolve? Thanks.

agent-0028:
Thank you so much for your efforts, @Titozzz!

Question for you, the app I am developing is a long way from production, so I might be willing to start using this while still in development. I just tried to install from this branch, and the Codegen step is failing because it can't find the folder `./src`, which I see listed in the Codegen config `"jsSrcsDir": "./src",`. I believe the issue is that the folder `src` is not listed in the `files` list, which is what `npm install` uses to determine what files in a repo are installed into `node_modules`.

@thiavila-sigma - you say you have been using this branch in a project, have you seen the same? Are you installing via `npm`, or somehow otherwise pulling down this repo and including it in your project?

(I am knee deep in developing my first Turbo Module, so all these details are top of mind.)

thiavila-sigma:
@agent-0028 I'm using yarn, so...
`yarn add react-native-webview/react-native-webview#new-arch-ios`

Also, I needed to do two patches. This is how the patching file looks:

```
diff --git a/node_modules/react-native-webview/index.js b/node_modules/react-native-webview/index.js
index 55a1a38..af0ba26 100644
--- a/node_modules/react-native-webview/index.js
+++ b/node_modules/react-native-webview/index.js
@@ -1,4 +1,4 @@
 
 export { WebView };
 export default WebView;
diff --git a/node_modules/react-native-webview/react-native-webview.podspec b/node_modules/react-native-webview/react-native-webview.podspec
index 05c92ab..26095cb 100644
--- a/node_modules/react-native-webview/react-native-webview.podspec
+++ b/node_modules/react-native-webview/react-native-webview.podspec
@@ -31,7 +31,7 @@ Pod::Spec.new do |s|
 
     s.dependency "React-RCTFabric"
     s.dependency "React-Codegen"
-    s.dependency "RCT-Folly", folly_version
+    s.dependency "RCT-Folly"
     s.dependency "RCTRequired"
     s.dependency "RCTTypeSafety"
     s.dependency "ReactCommon/turbomodule/core"
```

agent-0028:
@Titozzz - So, I can confirm that with npm version 8.1.0, when I install thusly:
`"react-native-webview": "agent-0028/react-native-webview#new-arch-wip",`

The `src` folder doesn't end up being installed in `node_modules`, and so iOS won't build.

I made a quick fork to play with this and adding that line at least makes it build, the only change on my branch is to add `src` to the files list in package.json.

I can sort of get the WebView to work in Android, but at least with HEAD of this branch iOS still shows a "not implemented" error.

I know we are working with WIP code, so this is all just fine!

Now that I see @thiavila-sigma's kind reply, I am glad I made a fork. I am going to try those patches and see if I can get this running in our app.

agent-0028:
So....I have this working in our app, thank you @thiavila-sigma!

On to the next question, and @Titozzz please please feel free to ignore me. I know I am using code in a feature branch, so I have little expectation of it being "feature complete".

For our app, the ability to pass messages from the WebView back to the React Native layer will be critical. I followed the examples in the docs and tried to get `window.ReactNativeWebView.postMessage("whatever")` to work and it doesn't seem to do anything.

Should I have any expectation that this feature is implemented and working in this branch?

Really, our needs are pretty simple, as long as I can render HTML and send messages in both directions we can keep rolling on our way to production :-)

thiavila-sigma:
@agent-0028 that's great to hear! Nothing on me though, all on @Titozzz ....
Regarding window.ReactNativeWebView.postMessage, I would guess it is not something related to the new-arch branch. There are multiple github issues opened to this, so I think you have a great chance to make it working by searching them.

agent-0028:
Heyo! It is me again :-)

I am having good success working with this branch, with no patches actually.

I have the `onMessage` event being triggered nicely from within my WebView with `window.ReactNativeWebView.postMessage`, so that takes care of one direction of communication.

I am not having any luck with `injectJavaScript` though.

Is there any reason to think this would be broken at the moment?

We are using hooks, so I mostly followed the documentation but used the `useRef()` hook instead of putting a reference on `this`.

These are the most relevant parts:
```
const webViewRef = useRef<any>(null)
...
  const handleSendMessageToWebView = async () => {
    const run = `
    document.body.style.backgroundColor = 'blue';
    true;
    `

    console.log(run)

    webViewRef.current.injectJavaScript(run)
  }
...
<Button text='Send message' onPress={handleSendMessageToWebView} />
...
        <WebView
...
          ref={webViewRef}
...
        />
```

agent-0028:
☝️ Oddly, with the exact same Javascript, I can get `injectedJavaScript` to work, but calling `injectJavaScript` does nothing.

Not sure if this is related, but I made a test fixture HTML file, and when I attempt to wire up some javascript to a button in the HTML, it ALSO doesn't work. It is almost as if the `javaScriptEnabled` prop is hard coded to `false` (I did try to add the prop even though `true` is documented to be the default.

I wish I could be more useful in troubleshooting this, but I have no earthly idea how to get console output from the embedded WebView...

As always, thanks for your efforts. I know this is all open source and I do not know what kind of resources "the community" has available to keep this moving.

sanjeevyadavIT:
I try adding `yarn add react-native-webview/react-native-webview#new-arch-ios` but getting this error
```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':react-native-webview'.
> Cannot convert the provided notation to a File or URI: [/Users/abc/xyz/node_modules/react-native-webview/android/build/generated/source/codegen/java].
  The following types/formats are supported:
    - A String or CharSequence path, for example 'src/main/java' or '/usr/include'.
    - A String or CharSequence URI, for example 'file:/usr/include'.
    - A File instance.
    - A Path instance.
    - A Directory instance.
    - A RegularFile instance.
    - A URI or URL instance.
    - A TextResource instance.
```

agent-0028:
@sanjeevyadavIT - I have been working with this branch in a project. In my experience if you are going to install it via npm or yarn, you will need to add the `src` folder to the files list in `package.json`.

My guess is that the good folks that are working on this project are using something like `wml` to link a folder in `node_modules` or otherwise making the code available for a test app.

Check this commit:
https://github.com/agent-0028/react-native-webview/commit/aaa307c642b2c245a178f5bb52f2d1ae75a90bce

It should install just fine and mostly work for iOS and Android.

Of course, this is a PR in progress, so no guarantees!

Good luck!

Titozzz:
Release Candidate here: yarn add react-native-webview@new-arch

TheAlmightyBob:
I tried upgrading our app (on the old architecture) to the RC branch, and it broke at build time because these changes are no longer compatible with these techniques that we (and I'm guessing at least some others) rely upon:
- https://github.com/react-native-webview/react-native-webview/blob/master/docs/Custom-iOS.md
- https://github.com/react-native-webview/react-native-webview/blob/master/docs/Custom-Android.md

In particular, there's no longer an exported `RNCWebViewManager` to override (there's probably more to it than that, but that's the "obvious" change that breaks things... did that export need to be removed or can it be brought back?).

Hopefully customization will still be possible and just needs new instructions, but I'm not currently familiar enough with the details of the changes in this PR to know what those should be.

agent-0028:
Hey folks! So glad to see this moving along.

We have some integration screens in our "new architecture" app and we have implemented the simplest case of using `injectJavaScript`.

I just upgraded to use the release candidate @Titozzz posted, and this is still not working.

The `onMessage` handler to send messages from the WebView to RN is working, but we are currently unable to send message into the WebView using `injectJavaScript`. Is this a known issue for you? Anything we can do to help narrow it down?

We did some poking around in the native source code before this RC was released, and we don't really see `injectJavaScript` implemented in any way we can understand (but that doesn't mean much, not our expertise).

If it helps you, the behavior seems the same for iOS and for Android.

Thanks again for the efforts!

Titozzz:
@agent-0028 is that something that's working on old arch, but not the new one, or just something that you are not sure how to implement with WebView in general?

agent-0028:
@Titozzz - Thanks for checking in! Our app did not have this WebView component before we upgraded to the "new architecture", so I don't have an easy comparison. I do have a spike branch I might revive to double check.

In any case, I am pretty sure I know how this is supposed to work, I have followed the example and the `ref` I have from the mounted WebView definitely has a method on it called `injectJavaScript`, but if I inject the simplest thing, as in the Guide:

```
document.body.style.backgroundColor = 'blue';
true;
```
...nothing happens.

Also, the same JavaScript DOES work when I use `injectedJavaScript` prop.

Titozzz:
Please open your project with XCode or Android Studio and place some breakpoints within injectJavascript to see if the method is called, it should be

Titozzz:
@agent-0028 I found that a missed something ... like correctly implementing the methods. I'll ping you when it's fixed

Titozzz:
@agent-0028 are you able to test the latest commit ?

agent-0028:
@Titozzz  - Thanks for the note! I'll install that tag tomorrow and let you know how it goes.

agent-0028:
@Titozzz - Some good news, with the HEAD of this branch, `injectJavaScript` is working on Android...sadly not yet working for me in iOS. I have a better mental model of the native code now so I will do as you suggest and debug this with some breakpoints in Xcode. I am "on the clock" for a client though, so I need to wrap up some other work before I get back to this. I will try to get you some more details tomorrow or Thursday. As always, thanks for your efforts!

Titozzz:
@agent-0028 did you install rc1 ? I tried it on both iOS and Android this time 👀

agent-0028:
@Titozzz - I am sorry, I pulled down this repo and I see no tags, releases or branches labeled `rc1`, so I wasn't really sure what you meant. Did you push something to `npm` that isn't evident in this repo? I installed directly off this branch, my package.json looks like this:

```
"react-native-webview": "react-native-webview/react-native-webview#new-arch-ios",
```

That should get HEAD of this feature branch, but package.json indicates `12.0.0-rc.0`, is there at `12.0.0-rc.1` I can install from npm instead of going to GitHub?

UNIDY2002:
Hi! Running `yarn test` in my react-native project is failing since I upgraded `react-native-webview` to 12.0.0-rc.1. Do you have an idea on how to fix it? Thanks!
Here is a repro (The first CI action passes while the second fails): https://github.com/UNIDY2002/webview-repro

matinzd:
@UNIDY2002 
You need to mock the module. 

```js
jest.mock('react-native-webview', () => ({
  default: () => jest.fn() // or any mocked component instead of native view,
}))
```

UNIDY2002:
@matinzd I got it! Thank you!

agent-0028:
@Titozzz - I should have some time to work on our integration this week. The `injectJavaScript` method is working in Android for me as of `12.0.0-rc.1`, but not for iOS. We are currently using RN `0.70.4`.

Quick questions that might help troubleshoot:
* What version of React Native are you using in a test app to install this into?
* Are you definitely running with Hermes and Fabric? Can you see something like `{"fabric":true,"initialProps":{"concurrentRoot":true},"rootTag":1}` in the Metro console?
* Are you doing anything special to link the native code, or are you able to rely on auto-linking?

I have put breakpoints in Xcode in a couple places in this method:
https://github.com/react-native-webview/react-native-webview/blob/new-arch-ios/apple/RNCWebViewImpl.m#L1314

And my test screen that calls `injectJavaScript` never gets there.

So, I started going down the JavaScript side, and I can see that I am getting down to this file:
`node_modules/react-native/Libraries/Utilities/codegenNativeCommands.js`

And that file has significantly changed between `0.70.4` and the latest `0.71.3`, which is why I am asking about the RN version you are using and about Fabric. In our version there is a check on the variable `global.RN$Bridgeless`, and in latest version there is no such check.

(We are implementing a bunch of new Turbo Modules, but I don't have much experience with Objective C, and as far as I can tell you are leaning on the new Codegen stuff but not exactly implementing a Turbo Module, so I really have no context for directly using the `codegenNativeCommands` utility.)

I am going to go ahead and start down the path of upgrading our app to `0.71.3` because eventually I'll have to anyway! But definitely still shooting in the dark as far as getting the WebView `injectJavaScript` method to work.

Thanks much!

Titozzz:
I did my latest testing on 0.71, so can't confirm for project earlier. You should run the example after cloning and see if that works for you?

agent-0028:
@Titozzz - So...I upgraded our project to `0.71.3` and installed `12.0.0-rc.0` and sadly, still no success for our simple test page that attempts to use `injectJavaScript`.

I haven't been this far into the weeds with this flavor of native module, so excuse my ignorance about these details.

I did as you suggested and set some breakpoints again here:
https://github.com/react-native-webview/react-native-webview/blob/new-arch-ios/apple/RNCWebViewImpl.m#L1314

And, we never hit the native code.

Just to make sure I had my dev environment set up, I put in a break point here:
https://github.com/react-native-webview/react-native-webview/blob/new-arch-ios/apple/RNCWebView.mm#L120

And when I call `window.ReactNativeWebView.postMessage(message)` the code execution stops and I am paused at the expected breakpoint.

We tried to do as you kindly suggested and run the example from this repo, and we were unable to get it to run. We followed the instructions here:
https://github.com/react-native-webview/react-native-webview/blob/master/docs/Contributing.md#for-ios

And got an error saying:
> Could not find Xcode project files in "react-native-webview/example/ios" folder.

Also tried getting it to run from Xcode and neither worked.

This was a good exercise however, because I looked at your test files and in `examples/Injection.tsx`, I don't actually see a test that exercises `injectJavaScript`.

I am pretty sure this issue has something to do with linking the native methods to JavaScript, and I know that has changed a lot since the introduction of the Fabric renderer.

Just to double check again, are you definitely running with Hermes and Fabric? Can you see something like {"fabric":true,"initialProps":{"concurrentRoot":true},"rootTag":1} in the Metro console?

I'd love to contribute to this project by setting up some steps to reproduce, but I am afraid I am blocked at this point without some other assistance :-)

Titozzz:
> the

Thank you for trying. Can I contact you directly via Twitter / discord for faster debugging ?

agent-0028:
@Titozzz - just sent you a Twitter DM with details, thanks for the offer!

BraveEvidence:
Check this I have managed to work web view with new architecture https://www.youtube.com/watch?v=Rq-mAjkG2ME&list=PLQhQEGkwKZUrempLnmxjt7ZCZJu1W3p2i&index=1

TheAlmightyBob:
I'd like to re-raise my concern from https://github.com/react-native-webview/react-native-webview/pull/2686#issuecomment-1384642783 about breaking the native customization features of the library. I think that's a pretty important capability, including for helping reduce the maintenance cost (since it allows not having to expose every possible thing you can do with a WebView to React Native). Leaving it out for the new architecture seems fine (since that's a new opt-in thing I wouldn't consider it "breaking"), but I think we need an answer for the old architecture.

@Titozzz Do you have thoughts on this? Would MRs targeting the `new-arch-ios` branch with suggested changes be helpful...?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
```diff
@@ -56,9 +56,12 @@ class RNCWebViewManagerImpl {
     private val DEFAULT_LACK_PERMISSION_TO_DOWNLOAD_MESSAGE =
         "Cannot download files as permission was denied. Please provide permission to write to storage, in order to download files."
 
+    fun createRNCWebViewInstance(context: ThemedReactContext): RNCWebView {
+        return RNCWebView(context)
+    }
```
TheAlmightyBob, line 61: I'm not sure this will help unless you also provide a similar method to substitute the `RNCWebViewManagerImpl` itself used by `RNCWebViewManager.java`...?

(That's why I was wondering if it would make sense for `createViewInstance` to take an optional `RNCWebView` argument instead... but, subclassing the whole `RNCWebViewManagerImpl` would also probably work... or maybe you have another approach in mind)

Titozzz: Yeah in my mind you would subclass the impl

skurgansky-sugarcrm:
> Released rc.2 cc @agent-0028 !

how can i try it ?

TheAlmightyBob:
@skurgansky-sugarcrm https://www.npmjs.com/package/react-native-webview/v/12.0.0-rc.2

agent-0028:
@Titozzz - great news, as of `12.0.0-rc.2`, `injectJavaScript` is working correctly in our app!

Sorry it took me so long to try it. The backlog is never ending :-)

Titozzz:
> @Titozzz - great news, as of `12.0.0-rc.2`, `injectJavaScript` is working correctly in our app!
> 
> Sorry it took me so long to try it. The backlog is never ending :-)

Well this is still unmerged so I know the feeling about never ending backlogs but thanks for testing it actually means a lot to me

jamonholmgren:
Congrats @Titozzz! Thanks so much for your hard work on this!

I wrote a Twitter thread giving you kudos: https://twitter.com/jamonholmgren/status/1642189446115209225

PR #2691, opened 2022-09-30
feat: add `getNativeTag` API

craftzdog:
to allow communicating directly with webview from the native modules like so:

```objc
RCT_EXPORT_METHOD(runJS:(NSString* __nonnull)js
                  inView:(NSNumber* __nonnull)reactTag
                  withResolver:(RCTPromiseResolveBlock)resolve
                  rejecter:(RCTPromiseRejectBlock)reject) {
  RCTUnsafeExecuteOnMainQueueSync(^{
    RCTUIManager* uiManager = [self.bridge moduleForClass:[RCTUIManager class]];
    RNCWebView* webView = (RNCWebView*)[uiManager viewForReactTag:reactTag];
    if (webView) {
      [webView injectJavaScript:js];
    }
    resolve(@"OK");
  });
}
```

I understand it's not for normal use cases, but I needed it to improve performance by not passing large data through the JS side like images.

Please refer to the background behind this work here: https://dev.to/craftzdog/how-i-improved-my-react-native-app-50x-faster-5ffi

But it stopped working with the recent webview versions because it no longer exposes the ref of the native component.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,7 +4,6 @@ module.exports = {
   parser: '@typescript-eslint/parser',
   plugins: ['react', 'react-native', 'import', '@typescript-eslint'],
   rules: {
-    'no-underscore-dangle': 'off',
     'no-console': 'off',
     // Lines will be broken before binary operators
     'operator-linebreak': ['error', 'before'],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,6 +4,7 @@ module.exports = {
   parser: '@typescript-eslint/parser',
   plugins: ['react', 'react-native', 'import', '@typescript-eslint'],
   rules: {
+    'no-underscore-dangle': 'off',
     'no-console': 'off',
     // Lines will be broken before binary operators
     'operator-linebreak': ['error', 'before'],
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -102,6 +102,7 @@ This document lays out the current public properties and methods for the React N
 - [`clearHistory`](Reference.md#clearHistory)
 - [`requestFocus`](Reference.md#requestFocus)
 - [`postMessage`](Reference.md#postmessagestr)
+- [`getNativeWebView`](Reference.md#getnativewebview)
 
 ---
 
@@ -1673,6 +1674,14 @@ clearHistory();
 ```
 
 Tells this WebView to clear its internal back/forward list. [developer.android.com reference](<https://developer.android.com/reference/android/webkit/WebView.html#clearHistory()>)
+
+### `getNativeWebView()`[⬆](#methods-index)<!-- Link generated with jump2header -->
+
+```javascript
+getNativeWebView();
+```
+
+Returns a ref of the native WebView component.
 
 ## Other Docs
 
--- a/src/WebView.android.tsx
+++ b/src/WebView.android.tsx
@@ -119,6 +119,8 @@ const WebViewComponent = forwardRef<{}, AndroidWebViewProps>(({
     clearFormData: () => Commands.clearFormData(webViewRef.current),
     clearCache: (includeDiskFiles: boolean) => Commands.clearCache(webViewRef.current, includeDiskFiles),
     clearHistory: () => Commands.clearHistory(webViewRef.current),
+    // @ts-ignore
+    getNativeTag: () => webViewRef.current._nativeTag,
   }), [setViewState, webViewRef]);
 
   const directEventCallbacks = useMemo(() => ({
--- a/src/WebView.android.tsx
+++ b/src/WebView.android.tsx
@@ -119,8 +119,7 @@ const WebViewComponent = forwardRef<{}, AndroidWebViewProps>(({
     clearFormData: () => Commands.clearFormData(webViewRef.current),
     clearCache: (includeDiskFiles: boolean) => Commands.clearCache(webViewRef.current, includeDiskFiles),
     clearHistory: () => Commands.clearHistory(webViewRef.current),
-    // @ts-ignore
-    getNativeTag: () => webViewRef.current._nativeTag,
+    getNativeWebView: () => webViewRef.current,
   }), [setViewState, webViewRef]);
 
   const directEventCallbacks = useMemo(() => ({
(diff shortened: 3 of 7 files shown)


craftzdog:
Instead of adding `getNativeTag`, it can be also solved by exposing the ref of the native component.
So, it'd be something like `getNativeRef`?

TheAlmightyBob:
Your finding that "recent webview versions... no longer exposes the ref of the native component" is curious... I _assume_ you're seeing this changed with 11.22.0? But I would have expected the use of `forwardRef` there to preserve the previous behavior..

craftzdog:
It was accessible with 11.16.0, since it was a class-based component and has `webViewRef` property here:
https://github.com/react-native-webview/react-native-webview/blob/v11.16.0/src/WebView.ios.tsx#L75

.eslintrc.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   parser: '@typescript-eslint/parser',
   plugins: ['react', 'react-native', 'import', '@typescript-eslint'],
   rules: {
+    'no-underscore-dangle': 'off',
```
TheAlmightyBob, line 7: Can this just be disabled for the specific lines you need it?

src/WebView.android.tsx
```diff
@@ -119,6 +119,8 @@ const WebViewComponent = forwardRef<{}, AndroidWebViewProps>(({
     clearFormData: () => Commands.clearFormData(webViewRef.current),
     clearCache: (includeDiskFiles: boolean) => Commands.clearCache(webViewRef.current, includeDiskFiles),
     clearHistory: () => Commands.clearHistory(webViewRef.current),
+    // @ts-ignore
+    getNativeTag: () => webViewRef.current._nativeTag,
```
TheAlmightyBob, line 123: It would be nice to have comments about this... e.g., what this is and why you expect it to be there and expose it even though it's not part of the type (and appears undocumented)...

And, along those lines, if it's going to be exposed then it probably should be added to `react-native-webview`'s own types/documentation.

TheAlmightyBob (changes requested):
I understand you're probably not keen on changing an existing working implementation, but wanted to mention that another possible route that we used for loading encrypted content was to give it custom schemes (e.g. `encrypted-file://<path>`) and then extend the native implementation (as per [the](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Custom-Android.md) [docs](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Custom-iOS.md)) to be able to decrypt & load, without using any Javascript injection (though I don't fully know your use case or whether you had a different reason for using injection).

src/WebView.android.tsx
craftzdog: You are right, it's undocumented. So, maybe this would be better instead:

```js
getNativeWebView: () => webViewRef.current
```

or

```js
getNativeWebViewRef: () => webViewRef
```

Which does look better for you?
This way, we no longer need `'no-underscore-dangle': 'off'`.

craftzdog:
Thanks for the tip! Yeah, extending the native implementation for custom schemes sounds nice. I'll look into it.
However, it has to be handled in JS first in order to deal with other related information.
So, it would be so helpful to provide a way to access the native webview ref 🙏
Thanks again for your time!

craftzdog:
Removed the change from `.eslintrc.js`, and changed the method from `getNativeTag` to `getNativeWebView` so that we don't have to directly refer to the undocumented property `_nativeTag` in the library.

src/WebView.ios.tsx
```diff
@@ -214,4 +215,4 @@ const isFileUploadSupported: () => Promise<boolean>
 
 const WebView = Object.assign(WebViewComponent, {isFileUploadSupported});
 
-export default WebView;
+export default WebView;
```
TheAlmightyBob, line 218: weird... I don't see what's different here..

TheAlmightyBob:
What about `WebView.windows.tsx`?

src/WebView.android.tsx
TheAlmightyBob: That seems okay, but I still think it should be added to the types and documented (if only to discourage its use except for advanced scenarios)

src/WebView.ios.tsx
craftzdog: I guess it's a trailing CR

craftzdog:
Oops, added!

craftzdog:
Would you please merge this? Thanks for your time and effort🙏

TheAlmightyBob:
@craftzdog I still think it needs to be added to the [types](https://github.com/react-native-webview/react-native-webview/blob/master/src/WebViewTypes.ts) and [documentation](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md)

craftzdog:
Got it, I'll add them

craftzdog:
added!

docs/Reference.md
```diff
@@ -1674,6 +1675,14 @@ clearHistory();
 
 Tells this WebView to clear its internal back/forward list. [developer.android.com reference](<https://developer.android.com/reference/android/webkit/WebView.html#clearHistory()>)
 
+### `getNativeWebView()`[⬆](#methods-index)<!-- Link generated with jump2header -->
+
+```javascript
+getNativeWebView();
+```
+
+Returns a ref of the native WebView component.
```
TheAlmightyBob, line 1684: Thanks for adding this, I'm just concerned that it could still be confusing.

In particular:
1. It might not be clear what "native WebView component" refers to... e.g., it's not like it really gives you access directly to the iOS/Android `RNCWebView` object. It would perhaps be more accurate to say something like "the underlying `ReactNativeFiberHostComponent`"? (I was even considering whether the method should be renamed, but it's called `NativeWebView` in code, so I guess that's okay)
2. I'd suggest including _some_ note along the lines of "You probably don't need this," something to indicate that it's for advanced scenarios (heck, you could even include a link to [your blog post](https://dev.to/craftzdog/how-to-control-the-existing-react-native-view-instance-from-another-native-module-3doi) to provide an example)

TheAlmightyBob (changes requested):
Do you know if the new React Native architecture (that @Titozzz is working on updating `react-native-webview` to be compatible with in #2686) introduces any complications here? Would that `NativeWebView` ref still work as expected?

TheAlmightyBob:
@craftzdog Alternatively, I wonder if getting the tag from an event would be sufficient? Although the Typescript types don't reflect it, it looks like it's generally available:

image

craftzdog:
@TheAlmightyBob Whoa, that works🤯 I didn't know that! Amazing.
I guess you can close this PR. Thanks for letting me know!

TheAlmightyBob:
@craftzdog Thanks for confirming it worked!

Issue #2654, opened 2022-08-26
ios 16, bounces={false} not working


baesumin:
when using ios 16 beta, bounces={false} doesn't work..

**Environment:**
 - OS: ios
 - OS version: 16 beta
 - react-native version: 0.66
 - react-native-webview version: 11.14.4

baesumin:
https://developer.apple.com/forums/thread/711601

Jim-Bar:
Same problem here, do someone knows how the workaround in the linked bug above can be applied in React Native?

mgscreativa:
Same here, bounces set to false but the webview "bounces" anyway

https://user-images.githubusercontent.com/1286632/205701018-1f91f827-b72f-47a7-bb85-33d30e25924c.MP4

Jim-Bar:
Hi, this issue seems to have been resolved by Apple, see #2692. We do not seem to reproduce it in our app.

squio:
Same here, issue seams to be resolved, please verify and close

albert17:
@Jim-Bar @squio Really? I am still facing this issue.
- react-native: 0.69.3
- react-native-webview: 11.2.4

squio:
@albert17 not sure about your case but for RN > 0.70 and RN Webview > 12 it seems to be fixed, maybe you can try to upgrade?

albert17:
@squio Updated to RN 0.71.7 and RNWebview 11.26.1 and still failing.

I will upgrade to RN Webview > 12 when there is a solution for testing. I tried and it broke my Jest testing.

nati673:
overScrollMode="never"

PR #2692, opened 2022-09-30
Fix webkit issue with `bounce={false}` on iOS 16

xavier-villelegier:
Hey hey !

### Issue
- Since iOS 16, there is a bug where using `bounce={false}` that is related to a [webkit bug](https://developer.apple.com/forums/thread/711601)
- It was also reported here #2654

The fix suggested on the apple thread works fine 👌 

### Preview
|Before|After|
|-----|------|
|Oct-01-2022 00-25-46|Oct-01-2022 00-24-54|

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -426,8 +426,8 @@ RCTAutoInsetsProtocol>
     _webView.scrollView.pagingEnabled = _pagingEnabled;
     //For UIRefreshControl to work correctly, the bounces should always be true
     _webView.scrollView.bounces = _pullToRefreshEnabled || _bounces;
-    _webView.scrollView.alwaysBounceHorizontal = _pullToRefreshEnabled || bounces;
-    _webView.scrollView.alwaysBounceVertical = _pullToRefreshEnabled || bounces;
+    _webView.scrollView.alwaysBounceHorizontal = _pullToRefreshEnabled || _bounces;
+    _webView.scrollView.alwaysBounceVertical = _pullToRefreshEnabled || _bounces;
     _webView.scrollView.showsHorizontalScrollIndicator = _showsHorizontalScrollIndicator;
     _webView.scrollView.showsVerticalScrollIndicator = _showsVerticalScrollIndicator;
     _webView.scrollView.directionalLockEnabled = _directionalLockEnabled;
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -426,6 +426,8 @@ RCTAutoInsetsProtocol>
     _webView.scrollView.pagingEnabled = _pagingEnabled;
     //For UIRefreshControl to work correctly, the bounces should always be true
     _webView.scrollView.bounces = _pullToRefreshEnabled || _bounces;
+    _webView.scrollView.alwaysBounceHorizontal = _pullToRefreshEnabled || bounces;
+    _webView.scrollView.alwaysBounceVertical = _pullToRefreshEnabled || bounces;
     _webView.scrollView.showsHorizontalScrollIndicator = _showsHorizontalScrollIndicator;
     _webView.scrollView.showsVerticalScrollIndicator = _showsVerticalScrollIndicator;
     _webView.scrollView.directionalLockEnabled = _directionalLockEnabled;
@@ -1423,6 +1425,8 @@ didFinishNavigation:(WKNavigation *)navigation
   _bounces = bounces;
   //For UIRefreshControl to work correctly, the bounces should always be true
   _webView.scrollView.bounces = _pullToRefreshEnabled || bounces;
+  _webView.scrollView.alwaysBounceHorizontal = _pullToRefreshEnabled || bounces;
+  _webView.scrollView.alwaysBounceVertical = _pullToRefreshEnabled || bounces;
 }
 #endif // !TARGET_OS_OSX
 


apple/RNCWebView.m
```diff
@@ -426,6 +426,8 @@ - (void)didMoveToWindow
     _webView.scrollView.pagingEnabled = _pagingEnabled;
     //For UIRefreshControl to work correctly, the bounces should always be true
     _webView.scrollView.bounces = _pullToRefreshEnabled || _bounces;
+    _webView.scrollView.alwaysBounceHorizontal = _pullToRefreshEnabled || bounces;
+    _webView.scrollView.alwaysBounceVertical = _pullToRefreshEnabled || bounces;
```
TheAlmightyBob, line 430: Does this compile? Shouldn't it be using `_bounces` instead of `bounces`?

TheAlmightyBob (changes requested):
Please update the description to explain what problem this fixes

xavier-villelegier:
> Please update the description to explain what problem this fixes

Yep just a branch to base my local branch on it. I updated the description, it works fine locally

apple/RNCWebView.m
xavier-villelegier: Yeah sure I didn't mean to open a PR yet, it was supposed to be in draft. I just fixed it 👌

TheAlmightyBob:
Thanks for including the before/after, but they're a little confusing... I'm guessing the entire UI, including the tabs, is all within a WebView? And that's some inner frame that's scrolling/bouncing on the right?

xavier-villelegier:
> Thanks for including the before/after, but they're a little confusing... I'm guessing the entire UI, including the tabs, is all within a WebView? And that's some inner frame that's scrolling/bouncing on the right?


Yep exactly ! Basically a full screen html page will still be bouncing on iOS 16 even with `bounce={false}` before, but not after the fix, the bounce is now really disabled again

apple/RNCWebView.m
```diff
@@ -426,6 +426,8 @@ - (void)didMoveToWindow
     _webView.scrollView.pagingEnabled = _pagingEnabled;
     //For UIRefreshControl to work correctly, the bounces should always be true
     _webView.scrollView.bounces = _pullToRefreshEnabled || _bounces;
+    _webView.scrollView.alwaysBounceHorizontal = _pullToRefreshEnabled || _bounces;
+    _webView.scrollView.alwaysBounceVertical = _pullToRefreshEnabled || _bounces;
```
TheAlmightyBob, line 430: While I trust that setting these to `false` works around the issue of setting `bounces` to `false`, I am concerned about side-effects, such as behavior on other iOS versions and setting these to `true` when `bounces` isn't `false` (e.g., according to Apple's documentation, the default value of `alwaysBounceVertical` is already supposed to be `false`)

TheAlmightyBob (changes requested):
A quick googling turned up the related issue: https://github.com/apache/cordova-ios/issues/1244

and the actual fix in WKWebKit: https://github.com/WebKit/WebKit/pull/2794

So
1. Since the "real" fix has been merged to WKWebKit, any workaround in `react-native-webview` should be pretty tightly scoped... assuming it's even still necessary?
2. It looks like the issue is related to iOS16 supporting the CSS `overscroll-behavior` property. Would setting that in your web content be an acceptable workaround?

xavier-villelegier:
> and the actual fix in WKWebKit: [WebKit/WebKit#2794](https://github.com/WebKit/WebKit/pull/2794)

Ah nice, didn't know they merged it already ! Thanks for the link


> 1. Since the "real" fix has been merged to WKWebKit, any workaround in `react-native-webview` should be pretty tightly scoped... assuming it's even still necessary?

I'm just concerned that this will take months for Apple to bump webkit on iOS, and in the meantime the user experience on full screen pages is pretty bad, once you started to bounce you have to wait until the bounce is really over otherwise you can't even normally scroll inside your web content 😕
How do we even know the version that iOS 16 is currently running ? The versioning isn't clear on https://github.com/WebKit/WebKit/tags, and the user agent doesn't contain the same kind of versions.


> 2. It looks like the issue is related to iOS16 supporting the CSS `overscroll-behavior` property. Would setting that in your web content be an acceptable workaround?

Just tried, and it doesn't work pretty well. You have to apply it on a lot of different layers in your DOM depending on the pages, and overall it's not a scalable solution anyway

TheAlmightyBob:
> How do we even know the webkit version that iOS 16 is currently running ?

Well, my thinking was that, since the fix has been merged, it might be reasonable to assume that it will be in iOS 16.1, so this workaround only needs to be in effect for 16.0.*...

xavier-villelegier:
> Well, my thinking was that, since the fix has been merged, it might be reasonable to assume that it will be in iOS 16.1, so this workaround only needs to be in effect for 16.0.*...

From what I tested, iOS 16.1 beta 3 that came out 4 days ago still has the bug. And I doubt it's going to come in a patch update, so we'll probably have to wait for 16.2 that won't came out before mid-december, and we're not even sure webkit will be bumped 😕

As we don't have any clear ETA, is there any big issue temporarily fixing this, and remove it once it's fixed on iOS directly ? If you want to limit the side effects I could expose these directly as props like `alwaysBounceVertical` and `alwaysBounceHorizontal`, so that everyone will control them and there won't be any breaking change ?

TheAlmightyBob:
> iOS 16.1 beta 3 that came out 4 days ago still has the bug

Ah. Dang. Thanks for testing.

> temporarily fixing this, and remove it once it's fixed on iOS directly ?

In theory, no, but I was a little worried that it wouldn't get removed... either due to forgetting, or due to wanting to support users that haven't upgraded to the fixed version.

> I could expose these directly as props like alwaysBounceVertical and alwaysBounceHorizontal, so that everyone can control them and there won't be any breaking change

That would indeed seem safe.

Though the current approach could probably be made safe enough also... my two main concerns are:
1. Limiting the scope to affected versions. While it sounds like we can't narrow to an exact affected version, it could at least be limited to >=16.
2. Better understanding how the "normal" case should behave. According to Apple's docs, the default value for these properties is already `false`. Is that incorrect? Did it change with iOS 16? Or is explicitly setting them to `false` somehow triggering a fix even if that was already the default value? More specifically, if the `bounces` property isn't set to `false`, is it appropriate to set `alwaysBounceVertical`/`alwaysBounceHorizontal` to `true`, or should they _still_ be set to `false` to preserve the normal behavior? Or should vertical/horizontal be set differently?

moughxyz:
What about exposing `alwaysBounceHorizontal` and `alwaysBounceVertical` as properties that can be customized via the react component? Would allow a fix for this issue + expanding the API to be more complete, without the concern raised here: https://github.com/react-native-webview/react-native-webview/pull/2692/files#r985015210

TheAlmightyBob:
@moughxyz 
> What about exposing alwaysBounceHorizontal and alwaysBounceVertical as properties that can be customized via the react component?

@xavier-villelegier also suggested that above. I think it's a reasonable option, but there's still the question of what the default values should be.

moughxyz:
How about null?

xavier-villelegier:
@TheAlmightyBob Closing this as it's fixed in iOS 16.2 beta, it should be release around December

mgscreativa:
Hi! @xavier-villelegier ! I thing the workaround didn't completely solved the issue, do you know if in iOS 16.2 the same old behavior will be restored? Here are my results on iOS 15.4 emulator. In the video I scroll the page all the way down and I don't get any bump (as expected), the same happens if I try to pull the page after reaching the top. Thanks!

| iOS 15.4  | iOS 16.1 with workaround|
| ------------- | ------------- |
|iOS 15.4 | iOS 16.1 with workaround |

Issue #2172, opened 2021-09-23
Cannot set Android cookies


t0x1c123:
**Bug description:**
Cannot set cookies on Android

**To Reproduce:**
```
           <WebView
                thirdPartyCookiesEnabled
                useWebKit
                source={{
                    uri: "http://10.0.2.2:3000",
                    headers: {
                        Cookie: "cookie1=asdf",
                    },
                }}
            />
```
**Expected behavior:**
to have cookies in document.cookie

**Environment:**
 - OS: Android
 - OS version: API 30
 - react-native version: 0.61.5
 - react-native-webview version: 11.13.0

Issue #1352, opened 2020-04-29
Cookies passed in custom header not working


dquessenberry:
**Bug description:**
I am trying to pass a cookie via a custom header, but it does not seem to work at all. My site looks for a specific cookie on certain pages and it doesn't seem to find it. I have also tried a super simple test example, exactly like the documentation, except the url is hitting a test page on a local server that creates an alert with all cookies. The documentation I am following is here: [https://github.com/react-native-community/react-native-webview/blob/master/docs/Guide.md#working-with-custom-headers-sessions-and-cookies](https://github.com/react-native-community/react-native-webview/blob/master/docs/Guide.md#working-with-custom-headers-sessions-and-cookies)

Am I missing something???

**To Reproduce:**
1) Spin up a local web server with a simple web page that uses JS to show an alert with all the document's cookies. JS example: 
```
<script type="text/javascript">
  alert('document.cookie='+document.cookie);
</script>
```
2) Create a simple RN app with the latest react-native-webview module, and add the following code:
```
const App = () => {
  return (
    <WebView
      source={{
        uri: 'http://localhost:8083', // location of the web page of step 1
        headers: {
          Cookie: 'cookie1=asdf; cookie2=dfasdfdas',
        },
      }}
      sharedCookiesEnabled={true}
    />
  );
};
```
3) Notice that none of the cookies set in the custom header are not found.

**Expected behavior:**
Cookies passed in a custom header can be read by the receiving web page.

**Environment:**
 - OS: macOS Catalina
 - OS version: 10.15.4
 - react-native version: 0.62.2
 - react-native-webview version: 9.4.0

jenriquejr:
Maybe it is not the same thing, but I was able to workaround this by setting incognito=true in the WebView.

In my case the android implementation was overriding the cookie header I was setting.
I need to debug further to see why this happens.

Hope it helps.

dharminderraja:
It was happening same with my implementation also. It was rewriting the cookies. After setting `incognito=true ` it seems to be working fine for now.

dwxw:
I've also found this, but only on Android, iOS works fine.  Adding incognito mode made Android work, but sadly I can't use incognito mode for other reasons.

TommyLeong:
@dwxw You mean you managed to send out custom header from iOS on all subsequent pages? I'm stuck here with this issue too. For now, my issue with iOS.

dwxw:
[...]

PR #2711, opened 2022-10-19
(Fix Android): Adding cookie header with case insensitive header key to android webview

ccreonopoulos:
Hello!

According to our tests and to a couple of issues on the issue tracker, passing the header **Cookie** from react-native, did not work on android. We were able to figure out that this is due to a case sensitivity of the header key. By adding this check of the header and adding it to the existing **CookieManager**, cookies passed in as headers started working fine on Android.

Issues found to reference this issue (there might be more): [2172](https://github.com/react-native-webview/react-native-webview/issues/2172), [1352](https://github.com/react-native-webview/react-native-webview/issues/1352)

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -565,6 +565,10 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
                 view.getSettings().setUserAgentString(headers.getString(key));
               }
             } else {
+              if (key.equalsIgnoreCase("Cookie")) {
+                CookieManager.getInstance().setCookie(url, headers.getString(key));
+              }
+              
               headerMap.put(key, headers.getString(key));
             }
           }
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -566,7 +566,9 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
               }
             } else {
               if (key.equalsIgnoreCase("Cookie")) {
-                CookieManager.getInstance().setCookie(url, headers.getString(key));
+                URI uri = new URI(url);
+                String domain = uri.getHost();
+                CookieManager.getInstance().setCookie(domain, headers.getString(key));
               }
               
               headerMap.put(key, headers.getString(key));


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -565,6 +565,10 @@ public void setSource(WebView view, @Nullable ReadableMap source) {
                 view.getSettings().setUserAgentString(headers.getString(key));
               }
             } else {
+              if (key.equalsIgnoreCase("Cookie")) {
+                CookieManager.getInstance().setCookie(url, headers.getString(key));
```
timbrownsf, line 569: Does this specifically narrows the cookie to the exact `uri` passed into the webview as the domain? Would it make more sense to actually set it to the host instead of the full uri?

ccreonopoulos: @timbrownsf Good point. We only needed it for 1 specific url in our project so it didn't occur to me to set it for a whole domain. PR updated.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -565,6 +565,12 @@ public void setSource(WebView view, @Nullable ReadableMap source) {
                 view.getSettings().setUserAgentString(headers.getString(key));
               }
             } else {
+              if (key.equalsIgnoreCase("Cookie")) {
+                URI uri = new URI(url);
+                String domain = uri.getHost();
+                CookieManager.getInstance().setCookie(domain, headers.getString(key));
```
TheAlmightyBob, line 571: According to the [documentation](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Guide.md#managing-cookies), cookies added via the `source` `headers` prop "will only be sent on the first request"

Setting the cookies via `CookieManager` seems like it would affect more than just the first request?

ccreonopoulos: Good point but this feature didn't work at all without setting the cookie in the CookieManager. Plus, that note has a "set custom headers" link too, which seems to include a "workaround" to exactly this, implying that this is a problem already.

Maybe we should modify the docs to only include **iOS** to that statement? And say on android you have to clear the cookie if you don't want it to carry over to consecutive calls to that domain? But I would like direction on this, as I am not an Android guy :)

TheAlmightyBob: @ccreonopoulos I get what you're saying, I just would rather not trade one confusing behavior for another. In any case it does seem like the documentation should be updated. That section of the guide in general seems to be iOS-specific, quite possibly written before Android support was even really a thing.

If the `Cookie` `headers` property is going to behave differently from the other `headers` properties, and not just apply to the initial `source` request, that suggests to me that it should be a separate property/method, not part of `source`.

Which then makes me wonder what the difference is between this approach and using [@react-native-community/cookies](https://github.com/react-native-community/cookies) as suggested in the guide?

ccreonopoulos: Thank you for your suggestions @TheAlmightyBob! If you think a separate property/method would be something we could accommodate, it would be much more preferable if this PR is not something you'd like to consider further. 

The reasons we didn't use the cookie library are:

1. We needed custom cookie usage, which although trying it the exact way the documentation suggested `If you'd like to send custom cookies in the WebView itself`, the "Cookie" header wasn't working (hence the the original reason for this PR - case sensitivity)

2. It would be a ton of overhead to add a new library and since our application is highly security related, we didn't want to have to increase our attack surface.

TheAlmightyBob: > We needed custom cookie usage, which although trying it the exact way the documentation suggested `If you'd like to send custom cookies in the WebView itself...`

So, I didn't write that of course, but the way I interpret that documentation is by emphasizing the _in the WebView itself_ part, to set cookies for only a specific web request without adding them to a shared cookie manager. But your code here is just adding them to the cookie manager anyway... which is [the same thing react-native-cookies does](https://github.com/react-native-cookies/cookies/blob/master/android/src/main/java/com/reactnativecommunity/cookies/CookieManagerModule.java#L181)?

> a ton of overhead to add a new library

I understand that pain, but it's a slippery slope to just cramming every feature into a single giant library when specialized libraries for those features already exist.

PR #2724, opened 2022-11-02
feat: add loadUrl and support "about:blank" on iOS

sunnylqm:


TheAlmightyBob:
What is the purpose of this?

sunnylqm:
1. To make iOS and android consistent (android already supports this)
2. A controlled webview is easy to use but hard to control. sometimes you need to use it imperatively instead of props which would cause other side effects (re-renders, duplicated events, re-callbacks)

TheAlmightyBob:
> android already supports loadUrl

Interesting... I don't think that was actually intentional, it looks like it was accidentally exposed during refactor... I wonder if there are side-effects to using that, e.g. whether all the events still work as expected...

sunnylqm:
I am writing a basic browser and `loadUrl` works fine on both ends

TheAlmightyBob:
Okay, I played with this briefly and didn't see any obvious issue (which still surprises me... since Android invokes `loadUrl` in the callback of `onShouldStartLoadWithRequest`, I thought that meant invoking `loadUrl` directly would bypass that event... but it still seemed to work).

Please update the docs to include the new method (though preferably pointing out that the `source` property is still the primary/recommended/supported way to set the URL), otherwise this looks okay to me.

TheAlmightyBob:
Er, did you mean to merge #2640 into this PR?

sunnylqm:
sorry i'll make a new pr

sunnylqm:
https://github.com/react-native-webview/react-native-webview/pull/2736

Issue #1990, opened 2021-05-18
granting RESOURCE_PROTECTED_MEDIA_ID


cstranex:
**Bug description:**
Trying to use EME results in an exception that no supported configurations are available. Investigation leads to PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID not being granted. I can make a slight change to RNCWebViewManager.js:1142 for the permission to be granted:
```
      for (int i = 0; i < permissions.size(); i++) {
        if (permissions.get(i).equals(PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID)) {
          grantedPermissions.add(PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID);
        }
        if (ContextCompat.checkSelfPermission(mReactContext, permissions.get(i)) != PackageManager.PERMISSION_GRANTED) {
          continue;
        }
```

I am not sure how the permission checking is supposed to work but I can't see where else to grant the `RESOURCE_PROTECTED_MEDIA_ID` permission without moving the code as above.

**To Reproduce:**
- Install react-native-webview
- Create a page that makes use of EME
- See exception

**Expected behavior:**
EME should be initialized

**Environment:**
 - OS: Android
 - react-native version: 0.63.3
 - react-native-webview version: 11.4.4

tijs:
I'm also having this issue. I can do a patch like @cstranex but there seems to be no, obvious, way to actually set the permissions for a webview?

spector32:
Same here

PR #2732, opened 2022-11-07
feat(Android): Handle RESOURCE_PROTECTED_MEDIA_ID permission

enzo-macri:
Add a new prop : allowsProtectedMedia to grant `android.webkit.resource.RESOURCE_PROTECTED_MEDIA_ID` permission to webview if set to true.

With the current implementation, Android WebViews can't play any media protected by DRM because the permission is not handled properly.

I propose an implementation where the developer can decide whether or not to allow it using a react prop `allowsProtectedMedia`. The default value is false.
The end user is not warned about this, I leave this responsibility to the developer. I believe this is acceptable as even applications such as Chrome Android do not ask for user permission, they are granted by default.

N.B: I haven't tackled the `RESOURCE_MIDI_SYSEX` permission as I don't have the proper setup to test it.

Also if this approach were to be accepted, should I leave the PR as is, or open a new one off the new arch branch?

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -161,6 +161,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
   protected RNCWebChromeClient mWebChromeClient = null;
   protected boolean mAllowsFullscreenVideo = false;
+  protected boolean mAllowsProtectedMedia = false;
   protected @Nullable String mUserAgent = null;
   protected @Nullable String mUserAgentWithApplicationName = null;
   protected @Nullable String mDownloadingMessage = null;
@@ -670,6 +671,20 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
     view.getSettings().setMinimumFontSize(fontSize);
   }
 
+  @ReactProp(name = "allowsProtectedMedia")
+  public void setAllowsProtectedMedia(WebView view, boolean enabled) {
+    // This variable is used to keep consistency
+    // in case a new WebChromeClient is created
+    // (eg. when mAllowsFullScreenVideo changes)
+    mAllowsProtectedMedia = enabled;
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+        WebChromeClient client = view.getWebChromeClient();
+        if (client != null && client instanceof RNCWebChromeClient) {
+            ((RNCWebChromeClient) client).setAllowsProtectedMedia(enabled);
+        }
+    }
+  }
+
   @Override
   protected void addEventEmitters(ThemedReactContext reactContext, WebView view) {
     // Do not register default touch emitter and let WebView implementation handle touches
@@ -875,8 +890,6 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
           mReactContext.removeLifecycleEventListener(this);
         }
       };
-
-      webView.setWebChromeClient(mWebChromeClient);
     } else {
       if (mWebChromeClient != null) {
         mWebChromeClient.onHideCustomView();
@@ -888,9 +901,9 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
           return Bitmap.createBitmap(50, 50, Bitmap.Config.ARGB_8888);
         }
       };
-
-      webView.setWebChromeClient(mWebChromeClient);
     }
+    mWebChromeClient.setAllowsProtectedMedia(mAllowsProtectedMedia);
+    webView.setWebChromeClient(mWebChromeClient);
   }
 
   protected static class RNCWebViewClient extends WebViewClient {
@@ -1227,6 +1240,9 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
     protected RNCWebView.ProgressChangedFilter progressChangedFilter = null;
 
+    // True if protected media should be allowed, false otherwise
+    protected boolean mAllowsProtectedMedia = false;
+
     public RNCWebChromeClient(ReactContext reactContext, WebView webView) {
       this.mReactContext = reactContext;
       this.mWebView = webView;
@@ -1288,9 +1304,11 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
         } else if (requestedResource.equals(PermissionRequest.RESOURCE_VIDEO_CAPTURE)) {
           androidPermission = Manifest.permission.CAMERA;
         } else if(requestedResource.equals(PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID)) {
-          androidPermission = PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID;
+          if (mAllowsProtectedMedia) {
+              grantedPermissions.add(requestedResource);
+          }
         }
-        // TODO: RESOURCE_MIDI_SYSEX, RESOURCE_PROTECTED_MEDIA_ID.
+        // TODO: RESOURCE_MIDI_SYSEX.
 
         if (androidPermission != null) {
           if (ContextCompat.checkSelfPermission(mReactContext, androidPermission) == PackageManager.PERMISSION_GRANTED) {
@@ -1479,6 +1497,15 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
 
     public void setProgressChangedFilter(RNCWebView.ProgressChangedFilter filter) {
       progressChangedFilter = filter;
+    }
+
+    /**
+     * Set whether or not protected media should be allowed
+     * /!\ Setting this to false won't revoke permission already granted to the current webpage.
+     * In order to do so, you'd need to reload the page /!\
+     */
+    public void setAllowsProtectedMedia(boolean enabled) {
+        mAllowsProtectedMedia = enabled;
     }
   }
 
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -1306,6 +1306,15 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
         } else if(requestedResource.equals(PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID)) {
           if (mAllowsProtectedMedia) {
               grantedPermissions.add(requestedResource);
+          } else {
+              /**
+               * Legacy handling (Kept in case it was working under some conditions (given Android version or something))
+               *
+               * Try to ask user to grant permission using Activity.requestPermissions
+               *
+               * Find more details here: https://github.com/react-native-webview/react-native-webview/pull/2732
+               */
+              androidPermission = PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID;
           }
         }
         // TODO: RESOURCE_MIDI_SYSEX.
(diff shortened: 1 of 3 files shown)


TheAlmightyBob:
Can you explain what's improper about the current handling? Although I wasn't involved with #1862, I would like to think that @crasyboy42 tested it... and it looks like the same logic is used for other permissions...

enzo-macri:
@TheAlmightyBob First thanks for your quick answer/support.

I opened this PR because when I tried to use a webview on a website where DRM are required, the permission request toast never shows up. Maybe I was mislead by the remaining TODO regarding PROTECTED_MEDIA_HANDLED and thought there were still some work to be done here.

Now if you look at the [current code](https://github.com/react-native-webview/react-native-webview/blob/master/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L1291), onPermissionRequest will call requestPermission and ask the user to accept permissions. To do so, webview permissions (android.webkit.resource) or mapped to manifest permissions (android.Manifest.permissions). Except for `RESOURCE_PROTECTED_MEDIA_ID` which is not mapped and does not seem to work. (I believe it would be the same for the `RESOURCE_MIDI_SYSEX` permission).

From what I understand Applications permissions (android permissions) and permissions you grant to the webview (webkit permissions) are not the same and therefore might not be handled the same way.
In my understanding android.webkit.resource.RESOURCE_PROTECTED_MEDIA_ID does not correspond to any android permission, whereas to grant webview access to the camera you should first grant your application access to the camera.

I might be mistaken here but it seems I'm not the only one facing this issue cf. https://github.com/react-native-webview/react-native-webview/issues/1990

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -1480,6 +1493,14 @@ protected ViewGroup getRootView() {
     public void setProgressChangedFilter(RNCWebView.ProgressChangedFilter filter) {
       progressChangedFilter = filter;
     }
+
+    /**
+     * Set whether or not protected media should be allowed
+     * /!\ Setting this to fault won't revoke permission already granted to the current webview /!\
```
TheAlmightyBob, line 1499: nit: 
```suggestion
     * /!\ Setting this to false won't revoke permission already granted to the current webview /!\ 
```

Also, this warning might be worth adding to the documentation?

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -1288,9 +1299,11 @@ public void onPermissionRequest(final PermissionRequest request) {
         } else if (requestedResource.equals(PermissionRequest.RESOURCE_VIDEO_CAPTURE)) {
           androidPermission = Manifest.permission.CAMERA;
         } else if(requestedResource.equals(PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID)) {
-          androidPermission = PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID;
+          if (mAllowsProtectedMedia) {
+              grantedPermissions.add(requestedResource);
+          }
```
TheAlmightyBob, line 1304: Do you think it would hurt to keep the previous code if `mAllowsProtectedMedia` is `false`, just in case there was some scenario (e.g. certain Android/WebView version??) where it actually did something useful?

TheAlmightyBob (changes requested):

TheAlmightyBob:
> Maybe I was mislead by the remaining TODO regarding PROTECTED_MEDIA_ID handling and thought there were still some work to be done here.

That TODO was written _before_ the existing support was added... they just didn't update the TODO at the time.

Thanks for the explanation! It's strange to me that there would be "permissions" that don't require going through the normal Android permission flow, but I haven't found anything that contradicts your thinking. Indeed, [this code snippet](https://chromium.googlesource.com/chromium/src/+/cf9457edb7c734f509dd12149acd78410c518570/android_webview/tools/system_webview_shell/apk/src/org/chromium/webview_shell/WebViewBrowserActivity.java#86) does seem to suggest that these particular permissions are special in that they don't map to a normal Android permission...

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
enzo-macri: Good catch thanks, and sorry for that typo!

I've added that to the documentation, and also I've edited this warning to be more accurate. As a matter of fact, I've performed more tests and tried to toggle the allowsProtectedMedia value and it appears that reloading the page would actually result in a new permission request and thus, take the allowsProtectedMedia value change into account.

Let me know if you think the explanation is clear enough.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
enzo-macri: A downside I see to doing that would be that it might hurt the code readability, and result in undocumented/unexpected behavior for developers ending up that case.

Now, I don't think it could do any harm since it would ask for the end-user approval anyway.

That being said, I'm gonna try to make the change this afternoon, I'll let you know once it's done.

TheAlmightyBob (approved):

TheAlmightyBob:
I'm really baffled as to why the Android job is still failing on this PR. It was rebased onto `master`, which is now succeeding, and there's nothing in your change that would reintroduce the previous issue.

My only guess is that something was cached from the broken state and not being refreshed.

Would you mind creating a new branch/PR to see if that fixes it?

enzo-macri:
> I'm really baffled as to why the Android job is still failing on this PR. It was rebased onto `master`, which is now succeeding, and there's nothing in your change that would reintroduce the previous issue.
> 
> My only guess is that something was cached from the broken state and not being refreshed.
> 
> Would you mind creating a new branch/PR to see if that fixes it?

I've created a new branch based on this one (without adding any other commit) and It seems it won't work as GitHub considered the CI has already ran for 40e5a778e9511d131c3b669e7a8e667bde37ff81.

TheAlmightyBob:
@enzo-macri Actually the error I'm seeing now is related to your change.

```
/home/runner/work/react-native-webview/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java:675: Error: Call requires API level 26 (current min is 23): android.webkit.WebView#getWebChromeClient [NewApi]
    WebChromeClient client = view.getWebChromeClient();
                                  ~~~~~~~~~~~~~~~~~~
```

I think you probably need something like `if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.O)`, as in #2640

enzo-macri:
> @enzo-macri Actually the error I'm seeing now is related to your change.
> 
> ```
> /home/runner/work/react-native-webview/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java:675: Error: Call requires API level 26 (current min is 23): android.webkit.WebView#getWebChromeClient [NewApi]
>     WebChromeClient client = view.getWebChromeClient();
>                                   ~~~~~~~~~~~~~~~~~~
> ```
> 
> I think you probably need something like `if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.O)`, as in #2640

Sorry, I totally forgot to run gradle check on my end. This should be fixed :)

TheAlmightyBob:
@enzo-macri Sorry, that issue/fix reminded me of the related problem https://github.com/react-native-webview/react-native-webview/pull/2640 ran into, where `setupWebChromeClient` may re-create the `RNCWebChromeClient` and lose the previously-applied setting. You haven't run into this in your testing? Have you tried setting the `AllowsFullscreenVideo` prop? (I don't think actually setting that was necessary to repro the problem, but I would expect doing so to make the problem _more likely_...)

enzo-macri:
> @enzo-macri Sorry, that issue/fix reminded me of the related problem #2640 ran into, where `setupWebChromeClient` may re-create the `RNCWebChromeClient` and lose the previously-applied setting. You haven't run into this in your testing? Have you tried setting the `AllowsFullscreenVideo` prop? (I don't think actually setting that was necessary to repro the problem, but I would expect doing so to make the problem _more likely_...)

I did not run into this during my testing, but I had a simple test case: A webview open on Spotify web player, and two buttons : one to toggle the `allowsProtectedMedia` prop, and another to reload the content of the webview.

I hadn't used the `allowsFullScreenVideo` prop so far, but you're right, it seems it would cause the loss of the setting since it creates a new instance of RCNWebChromeClient. After testing it, it indeed loose the `mAllowsProtectedMedia` setting as expected.

Also I haven't ran into the case where `onDropViewInstance` is called and then `mWebChromeClient` is set to null. Maybe this is due to the fact I'm not using directly the `mWebChromeClient` view, but `view.getWebChromeClient()` instead (based on the discussion you had in #2640). However, if you have any idea on how to reproduce this, let me know.

I see two immediate solutions (They are actually more workarounds than proper solutions) to this:
- in `setupWebChromeClient` checking if `webView.getWebChromeClient()` is not null and an instance of `RNCWebChromeClient`, and if so reuse its settings in the new instance we are creating here.
- Saving mAllowsProtectedMedia state in `RNCWebViewManager` in addition to `RNCWebChromeClient` to avoid losing the state.

Of course, a downside of these is that every setting impacting the WebChromeClient should be reapplied when creating the new instance which is a bit tedious.

Given this, another option would be for `mAllowsProtectedMedia` to be a member of `RNCWebView` so that it would be accessible from both `RNCWebChromeClient` and `RNCWebViewManager` (therefore no need to store this state there) and wouldn't be reset in `setupWebChromeClient`. However, if there's also an issue with `onDropViewInstance`, it probably wouldn't solve it. Moreover, this just doesn't feel right to store this property in the View class.

Do you have any other ideas?

TheAlmightyBob:
> I haven't ran into the case where onDropViewInstance is called and then mWebChromeClient is set to null. Maybe this is due to the fact I'm not using directly the mWebChromeClient view, but view.getWebChromeClient() instead (based on the discussion you had in https://github.com/react-native-webview/react-native-webview/pull/2640). However, if you have any idea on how to reproduce this, let me know.

I never repro'd that issue, my suggestion related to it was based purely on what @Ldoppea said they were observing.

I agree that it "doesn't feel right to store this property in the View class," though that is an interesting thought.

Both of your "immediate" solutions sound reasonable to me. I lean towards the second only because that's what #2640 is currently doing and would have some slight consistency with `mAllowsProtectedMedia`. But if it becomes more common to expose properties that apply to the WebChromeClient, I could imagine that the "reuse settings" approach might be more scalable.

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -165,6 +165,7 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
   protected @Nullable String mUserAgentWithApplicationName = null;
   protected @Nullable String mDownloadingMessage = null;
   protected @Nullable String mLackPermissionToDownloadMessage = null;
+  protected boolean mAllowsProtectedMedia = false;
```
TheAlmightyBob, line 168: nit: how about moving this up to where `mAllowsFullscreenVideo` is declared?

src/WebViewTypes.ts
```diff
@@ -1119,6 +1120,13 @@ export interface AndroidWebViewProps extends WebViewSharedProps {
    * @platform android
    */
   lackPermissionToDownloadMessage?: string;
+
+  /**
+   * Boolean value to control whether webview can control media protected by DRM.
```
TheAlmightyBob, line 1125: nit: "control" media, not "play" it?

TheAlmightyBob (changes requested):
Just a couple small things, otherwise looks good

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
enzo-macri: Done :)

TheAlmightyBob (approved):

Issue #2633, opened 2022-08-11
File picker doesn't work on macOS


Ava7788:
**Bug description:**
I am using react-native-macos to build a react native macOS app. It is an all-webview app that RN is just a wrapper to render pages on the app.
There is an upload documents function on the website that expects to open the mac file finder(system folder) and allow the user to choose one file to upload. However, no action after clicking the button. Users can only drag or drop files to there
This function works on the website and ios platform. 

**Expected behavior:**

After clicking the button, expect to popup the mac file finder and allow users to select a file to upload.

**Screenshots/Videos:**

<img width="511" alt="Screen Shot 2022-08-11 at 3 57 42 pm" src="https://user-images.githubusercontent.com/110361291/184072607-885b2f19-ff6e-4551-a6c5-2e2611dbaa4f.png">

**Environment:**
 - OS:macOS
 - OS version:12.5
 - react-native version:"0.69.1"
 -  react-native-macos version:"^0.66.0-0"
 - react-native-webview version:"^11.22.7"

banwarCoeus:
@Ava7788 Were you able to find a solution for it? I am facing this same issue.

Ava7788:
> @Ava7788 Were you able to find a solution for it? I am facing this same issue.

No, I don't find any solution about it...

ninohoglund:
This worked for me:
https://bitcoden.com/answers/how-do-i-get-the-file-open-dialog-to-work-when-called-through-a-swiftui-wkwebview-inside-a-macos-native-app

amilich:
@ninohoglund can you explain how you implemented the solution above?

amilich:
Resolved!

PR #2733, opened 2022-11-08
Support File Input On Macos

17Amir17:
Support File Input On MacOS #2633, by implementing https://developer.apple.com/documentation/webkit/wkopenpanelparameters?language=objc

This requires `User Selected File` read permission otherwise will crash when attempting to open file input

Screen Shot 2022-11-08 at 12 28 32

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -315,6 +315,23 @@ RCTAutoInsetsProtocol>
   return nil;
 }
 
+/**
+ * Enables file input on macos, see https://developer.apple.com/documentation/webkit/wkuidelegate/1641952-webview
+ */
+#if TARGET_OS_OSX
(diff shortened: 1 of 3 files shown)


amilich:
Hey all! Would love to get this merged.

17Amir17:
@Titozzz You can test this by changing the url in the example app to `imgur` or some other website with a file input and check if it opens.

amilich:
Any way to update this and get it in?

amilich:
@Titozzz @jamonholmgren any way we can merge this?

amilich:
Is this actively maintained?

TheAlmightyBob:
> You can test this by changing the url in the example app to imgur or some other website with a file input and check if it opens

There is an existing file upload page in the example app, but it is currently only enabled for Android (via `Platform.OS` check in `App.tsx`). Can that be used for MacOS with this change?

17Amir17:
> There is an existing file upload page in the example app, but it is currently only enabled for Android (via `Platform.OS` check in `App.tsx`). Can that be used for MacOS with this change?

Yup! I'll update it

Screenshot 2022-12-07 at 17 43 33

TheAlmightyBob (approved):

PR #2736, opened 2022-11-09
feat: add loadUrl and support "about:blank" on iOS

sunnylqm:


Titozzz:
why would you want to imperatively do that instead of using the prop ?

sunnylqm:
it's a choice for developers, no harm anyway.

TheAlmightyBob:
@Titozzz FWIW, I added a test component to the example app that logs the lifecycle events (might try to clean that up and PR it at some point, seems like a useful test to have) and, to my surprise, didn't see any difference between setting the `source` property and invoking `loadUrl`. That includes scenarios like returning `false` from `onShouldLoadWithRequest`, which I had been particularly concerned about. Of course, it's possible that I missed something.

I'm not quite sure what you mean by
> this kind of feature could easily be implemented with a forwardRef, useState and useImperativeHandle without any side effect like rerendering

Isn't that what this PR is doing? Adding to the existing `useImperativeHandle`?

Titozzz:
> it's a choice for developers, no harm anyway.

To be honest, WebView is already very complicated to maintain with tons of props so if there is already a prop to achieve a behaviour, I'm quite unhappy to add more complexity. Another examples of such cases of overly complicated props are renderLoading, renderError, that could be implemented in userland, also postMessage is a duplicate of injectJavascript.

Titozzz:
> @Titozzz FWIW, I added a test component to the example app that logs the lifecycle events (might try to clean that up and PR it at some point, seems like a useful test to have) and, to my surprise, didn't see any difference between setting the `source` property and invoking `loadUrl`. That includes scenarios like returning `false` from `onShouldLoadWithRequest`, which I had been particularly concerned about. Of course, it's possible that I missed something.
> 
> I'm not quite sure what you mean by
> > this kind of feature could easily be implemented with a forwardRef, useState and useImperativeHandle without any side effect like rerendering
> 
> Isn't that what this PR is doing? Adding to the existing `useImperativeHandle`?

What I meant is that I'd rather someone needing this implementing the feature outside the lib than inside.

As long as WebView give developpers all the tools they need, I would refrain from implementing too much user logic within the component. There is already so much that doesn't belong inside WebView

TheAlmightyBob:
I do agree that WebView is complicated to maintain, and admit I still don't fully understand why setting the `source` property would be insufficient.

And, it's a bit off-topic, but I've wondered why `postMessage` was never removed as the Readme still says it will be...

I think the "about:blank" change is safe though.

sunnylqm:
We can think `webview` as a controlled component via setting `source` and `onNavigationStateChange` both as data bindings right? And if I control the `source` prop in redux and pass it to a textinput, it should work as a browser address bar  , either I input an url manually or click a link with the page, the url in the textinput and webview should perfectly synchronize right?  
The problem is that it's not working as simple as a textinput. There are fundamental inconsistencies between the webview internal state and the react state due to the complex network condition. Say if you are loading an unreponding page (server down, timeout, ...), the `source` prop would be overwrite repeatedly with `onNavigationStateChange` event. You simply can not control the webview in a controlled component way in such situation.

Titozzz:
@sunnylqm (I think) I understand your needs but why can't you do something like this ?

```tsx
const ControlledWebview = forwardRef(
  (
    props: ComponentProps<typeof WebView>,
    ref: ForwardedRef<{
      updateSource: (source: string) => void;
    }>,
  ) => {
    const [uri, setUri] = useState('about:blank');

    useImperativeHandle(ref, () => ({
      updateSource: (newUri: string) => {
        setUri(newUri);
      },
    }));

    return <WebView {...props} source={{ uri }} />;
  },
);

```

sunnylqm:
No. There are many other props and events attached to the webview instance, the way you suggested would still trigger rerender (it's still `setState` anyway) and duplicated events and cause dead loop sometimes

sunnylqm:
Sorry I am not a native speaker. Let me try to rephrase:

```
<Webview source={source} onNavigationStateChange={{ /* set url back to source */ }}
```

The above is a perfect typical controlled component right? When user clicks a link -> `onNavigationStateChange` sets url back to `source` prop -> if the url is broken and hang on loading -> you try to set the source prop to 'about:blank' via a textinput -> but `onNavigationStateChange` continues to revert it back to a broken url

Titozzz:
if changing the source prop to 'about:blank' and still receiving onNavigationStateChange for the old broken url, that's an issue with onNavigationStateChange. Rerendering the webview should never be an issue, even more with the new arch

Issue #2522, opened 2022-06-01
Implement custom MenuItems feature for Android to get feature parity to iOS


MrRio:
The MenuItems feature is a way to add custom menu items to the context menu after text selection. It's [described in the API documentation](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#menuitems) and doesn't currently mention that it's iOS-only. We've erroneously built an app which needs this functionality, assuming it was available in both platforms. At the very least, the documentation will need updating to indicate that it's iOS-only to prevent anyone else running into the same issue.

<img width="476" alt="126560575-1712b4f3-366b-4ff6-95c3-ff0256dfd1b2" src="https://user-images.githubusercontent.com/47539/171347934-a21d932d-bf70-41a9-bc3e-db68e6f94789.png">

There have been previous tickets mentioning this:

- https://github.com/react-native-webview/react-native-webview/issues/2244

Including the tickets that the original author of the iOS version linked to:

- https://github.com/react-native-webview/react-native-webview/issues/647
- https://github.com/facebook/react-native/issues/3414
- https://github.com/react-native-webview/react-native-webview/issues/1117#issuecomment-698020935

### Possible Solutions

There are some possible solutions linked in various places, so I will try to collate them here to make it easier to talk through what a good implementation might look like.

**[Stack Overflow: Use a custom contextual action bar for WebView text selection](https://stackoverflow.com/questions/22336903/use-a-custom-contextual-action-bar-for-webview-text-selection)**

This solution overrides the onActionModeStarted, and defines a 'click' handler. 

**[React Native Selectable Text](https://github.com/Astrocoders/react-native-selectable-text)**

This project implements the functionality needed, but outside of a WebView, it could potentially be possible to port some of the code from here?

Flutter InAppWebView has a context menu item implementation:  
https://github.com/pichillilorenzo/flutter_inappwebview/blob/442b51db0f0d9995f8f0fe24c9e26b8d9f06bba1/android/src/main/java/com/pichillilorenzo/flutter_inappwebview/in_app_webview/InAppWebView.java#L1344

---

Getting feature parity where possibles makes a much better experience IMHO. We're going to have a look to see if we can implement this with our in-house skills, or potentially find someone to help out, unless there's someone in the community that thinks they could tackle this?

MrRio:
[...]

PR #2610, opened 2022-07-21
(iOS & Android) Custom Menu Items - Android implementation & removes default menu in iOS 

naveennazimudeen:
### What's in this PR
---
- Implements Custom Context Menu in Android
- Removes the default menu items, showing only the custom menu items in both iOS and Android
- Added a CustomMenu example 

This adds to the work done in #2101 in iOS, and resolves issues #2244, #2522 for Android.

| iOS  | Android |
| ------------- | ------------- |
|  |  |

### How to test
---

Add WebView as shown below:
```
<WebView
  source={{html: HTML}}
  menuItems={[{ label: 'Highlight', key: 'highlight' }, { label: 'Strikethrough', key: 'strikethrough' }]}
  onCustomMenuSelection={(webViewEvent) => {
    const { label, key, selectedText } = webViewEvent.nativeEvent;
    console.log('Custom Menu Item Clicked: ', label, ' :: ', key, ' :: ', selectedText)
  }}
/>
```

A demo has also been added to examples. Custom Menus as set in the WebView prop `menuItems` will be shown on highlighting text. And then on tapping a menu item, `onCustomMenuSelection` will be called. 

If no `menuItems` prop, then the default menu is shown.

PR #2748, opened 2022-11-20
fix(docs): Add information that custom menu items are only available for iOS

Handschrift:
This PR adds two new information tables to the following sections in the Reference.portuguese.md and in the Reference.md:

- menuItems
- onCustomMenuItemSelection

These features are currently only available on iOS which was discussed in some issues and pull requests already:

- #2523 
- #2522 

Currently there is no documentation that says that this feature is only available on iOS. As long there is no feature parity this should be stated in the documentation.

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -1508,6 +1508,10 @@ Example:
 
 An array of custom menu item objects that will be appended to the UIMenu that appears when selecting text (will appear after 'Copy' and 'Share...').  Used in tandem with `onCustomMenuSelection`
 
+| Type                                                               | Required | Platform |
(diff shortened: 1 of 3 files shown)


TheAlmightyBob:
There's also #2610, but for now this is a good catch... I could have sworn this was already documented!

TheAlmightyBob (approved):

TheAlmightyBob:
@Handschrift Would you mind updating the types also? It looks like those could use corresponding `@platform ios` annotations

Handschrift:
Yes, I added the @platform annotations to the types

TheAlmightyBob (approved):

PR #2766, opened 2022-12-02
fix(Android): Apply WebView settings before loading content

### LINKED ISSUE #2723: Loading local files broken since `11.22.0` on Android (`net::ERR_ACCESS_DENIED`)
**Bug description:**
Since version `11.22.0`, it is not possible anymore to load a uri pointing to the file system on Android. The WebView shows an error message.

This 

```
source={{uri: "file:///..."}}`) 
```

leads to this

```
net::ERR_ACCESS_DENIED
```

I am aware of the `allowFileAccess` setting and have set it to true.

It works fine on version `11.21.2`.

**Expected behavior:**
Uris pointing to file system should load without error.

**Environment:**
 - OS: Android
 - OS version: 12
 - react-native version: 0.69.6
 - react-native-webview version: 11.22.0

TheAlmightyBob:
Fixes #2723. As mentioned in that issue, the problem was that the `11.22.0` refactor moved the `source` property before others when rendering `NativeWebView` on Android, which meant that `allowFileAccess` was not set until after the WebView was already trying (and failing) to load a local file. Other platforms were unaffected because the `source` property was not moved in those files.

Added example from @jacobpenny. Modified to use [react-native-blob-util](https://github.com/RonRadtke/react-native-blob-util) instead of [react-native-file-access](https://github.com/alpha0010/react-native-file-access) to populate the local filesystem because the latter broke the macOS build. The former does not support macOS or Windows, so the example is disabled on those platforms, but at least it doesn't break the build (it **did** require a tweak to the `build.gradle` file for Android, related to [this issue](https://github.com/facebook/react-native/issues/35210) again... I'm _guessing_ `react-native-blob-util` needs to update its React Native dependency, but I'm not 100% certain...)

(The `podfile.lock` change is just because that's been getting updated any time I run `pod install` on the example project for a long time now so I finally decided maybe I should actually commit the updated version)

To be extra-clear: The actual fix here is just a tiny change in `WebView.android.tsx`. Everything else is just about adding the example.

### DIFF
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -174,11 +174,15 @@ export default class App extends Component<Props, State> {
             title="ImportHtmlAsset"
             onPress={() => this._changeTest('ImportHtmlAsset')}
           />
-          <Button
-            testID="testType_localHtmlFile"
-            title="LocalHtmlFile"
-            onPress={() => this._changeTest('LocalHtmlFile')}
-          />
+          {/* Loading Local files should be supported on other platforms, but the test page
+              uses react-native-blob-util, which only supports iOS/Android */}
+          {(Platform.OS === 'ios' || Platform.OS === 'android') && (
+            <Button
+              testID="testType_localHtmlFile"
+              title="LocalHtmlFile"
+              onPress={() => this._changeTest('LocalHtmlFile')}
+            />
+          )}
           <Button
             testID="testType_downloads"
             title="Downloads"
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -17,6 +17,7 @@ import Downloads from './examples/Downloads';
 import Uploads from './examples/Uploads';
 import Injection from './examples/Injection';
 import ImportHtmlAsset from './examples/ImportHtmlAsset';
+import { LocalHtmlFile } from './examples/LocalHtmlFile';
 import Messaging from './examples/Messaging';
 import NativeWebpage from './examples/NativeWebpage';
 import ApplePay from './examples/ApplePay';
@@ -84,6 +85,14 @@ const TESTS = {
     description: 'Import HTML Asset test',
     render() {
       return <ImportHtmlAsset />;
+    },
+  },
+  LocalHtmlFile: {
+    title: 'LocalHtmlFile',
+    testId: 'LocalHtmlFile',
+    description: 'Local HTML File test',
+    render() {
+      return <LocalHtmlFile />;
     },
   },
   NativeWebpage: {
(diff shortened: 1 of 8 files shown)


TheAlmightyBob:
@Titozzz @jamonholmgren Could I get a quick approval on this small bug fix? Can't approve my own PRs.

(as noted in the description, the actual fix is just moving one line in one file.. the rest is about adding an example... though if you have concerns about that I welcome you to share them)

Titozzz:
Do you have discord ? I'd like to chat about the changes 🙂

TheAlmightyBob:
@Titozzz I don't. I'm in the Infinite Red React Native Slack though? And I'm not opposed to Discord, I just don't currently use it.

TheAlmightyBob:
Hello bot yes that's a mistake please keep this open

TheAlmightyBob:
Hm. The Windows CI failure is:

```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v160\Microsoft.Cpp.WindowsSDK.targets(46,5): error MSB8036: The Windows SDK version 10.0.18362.0 was not found. Install the required version of Windows SDK or change the SDK version in the project property pages or by right-clicking the solution and selecting "Retarget solution". [D:\a\react-native-webview\react-native-webview\node_modules\react-native-blob-util\windows\ReactNativeBlobUtil\ReactNativeBlobUtil.vcxproj]
Done Building Project "D:\a\react-native-webview\react-native-webview\node_modules\react-native-blob-util\windows\ReactNativeBlobUtil\ReactNativeBlobUtil.vcxproj" (default targets) -- FAILED.
```

It worked before, so I'm _guessing_ this is because the Windows CI was updated to target newer versions as part of the new architecture changes, and `react-native-blob-util` is depending on an older version...

TheAlmightyBob:
Hi bot please keep open... for now...

ArGeoph:
Can you please reopen this one? We're still facing this bug with Fabric enabled with the current latest version of `react-native-webview` (**13.6.2**), so it's blocking us from enabling the new architecture in production.

TheAlmightyBob:
@ArGeoph I'll reopen, but your comment that you're seeing a bug "with Fabric enabled" tells me that this PR won't help, at least in its current state. This bug predates the new architecture work (so is not specific to Fabric being enabled), and my last update was that the changes here no longer seem to help after rebasing on the new architecture changes. Unfortunately I think there's still a lot of work to do here...

ArGeoph:
> @ArGeoph I'll reopen, but your comment that you're seeing a bug "with Fabric enabled" tells me that this PR won't help, at least in its current state. This bug predates the new architecture work (so is not specific to Fabric being enabled), and my last update was that the changes here no longer seem to help after rebasing on the new architecture changes. Unfortunately I think there's still a lot of work to do here...

Thanks for the reply! I'll see if I can do anything to fix this issue and open a PR if I find a solution.

Issue #2353, opened 2022-01-28
File download on Android doesn't understand "filename*=UTF-8''" nor basic auth


djibarian:
**Bug description:**
Downloading a file in Android where the filename is defined with this alternate format `Content-Disposition: attachment; filename*=UTF-8''...` makes the filename being ignored, using the URL pathname instead. Additionally, basic auth is ignored.

**To Reproduce:**

**Expected behavior:**
Filename and basic auth should be properly handled, as it is in the standalone browser.

**Screenshots/Videos:**

**Environment:**
 - OS: Android
 - OS version:
 - react-native version:
 - react-native-webview version:

piechoo:
Just ran into this bug, still an issue

Andre-at-LaCapitale:
Same for us :-/ still an issue

UNIDY2002:
I made a fix in #2767 which has been included in version 11.25.1. You may try it out and leave feedbacks if the problem remains.

PR #2767, opened 2022-12-02
[Android] Respect "filename*" parameter in the field Content-Disposition when detecting filenames for downloading.

UNIDY2002:
According to [Section 4.3 of RFC6266](https://www.rfc-editor.org/rfc/rfc6266#section-4.3), filenames with non-ASCII characters may be specified with a `"filename*"` parameter.

This library uses `android.webkit.URLUtil.guessFileName` to detect filenames. However, the API is a simplistic implementation of the RFC, and thus cannot understand the `"filename*"` parameter, which leads to issue #2353.

I modified the implementation of `android.webkit.URLUtil` and added a naive support for the `"filename*"` parameter. Besides, I added a test case in `example/examples/Downloads.tsx` to cover this case.

NOTE that my modified implementation is still too simple, and may fail to cover [every corner case](http://test.greenbytes.de/tech/tc2231/) of the RFC. Nevertheless, since filename detection is not the focus of this library, the modification shall be valuable as long as it understands the `"filename*"` parameter in most cases.

Looking forward to your revision.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/URLUtil.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/URLUtil.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright (C) 2006 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.reactnativecommunity.webview;
+
+import android.net.Uri;
+import android.webkit.MimeTypeMap;
+import androidx.annotation.Nullable;
+
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class URLUtil {
+    /**
+     * Guesses canonical filename that a download would have, using
+     * the URL and contentDisposition. File extension, if not defined,
+     * is added based on the mimetype
+     * @param url Url to the content
+     * @param contentDisposition Content-Disposition HTTP header or {@code null}
+     * @param mimeType Mime-type of the content or {@code null}
+     *
+     * @return suggested filename
+     */
+    public static final String guessFileName(
+            String url,
+            @Nullable String contentDisposition,
+            @Nullable String mimeType) {
+        String filename = null;
+        String extension = null;
+
+        // If we couldn't do anything with the hint, move toward the content disposition
+        if (filename == null && contentDisposition != null) {
+            filename = parseContentDisposition(contentDisposition);
+            if (filename != null) {
+                int index = filename.lastIndexOf('/') + 1;
+                if (index > 0) {
+                    filename = filename.substring(index);
+                }
+            }
+        }
+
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/reactnativecommunity/webview/URLUtil.java
```diff
@@ -0,0 +1,157 @@
+/*
+ * Copyright (C) 2006 The Android Open Source Project
```
TheAlmightyBob, line 2: Please clarify what exactly is modified in this file.

example/examples/Downloads.tsx
```diff
@@ -21,6 +21,8 @@ const HTML = `
   </head>
   <body>
     <a href="https://www.7-zip.org/a/7za920.zip">Example zip file download</a>
+    <br>
+    <a href="http://test.greenbytes.de/tech/tc2231/attwithisofn2231iso.asis">Download file with non-ascii filename</a>
```
TheAlmightyBob, line 25: Would be helpful to say in this text (or at the very least a code comment) what the expected filename is

TheAlmightyBob (changes requested):

android/src/main/java/com/reactnativecommunity/webview/URLUtil.java
```diff
@@ -0,0 +1,168 @@
...
+            Matcher m = CONTENT_DISPOSITION_PATTERN.matcher(contentDisposition);
+            if (m.find()) {
+                if (m.group(3) != null && m.group(4) != null) {
+                    try {
+                        return URLDecoder.decode(m.group(4), m.group(3));
+                    } catch (UnsupportedEncodingException e) {
+                        // Skip the ext-parameter as the encoding is unsupported
+                    }
+                }
+                return m.group(2);
```
TheAlmightyBob, line 161: I like the summary change list comment you put at the top, but could you also add something briefly here to say what this is doing? (e.g., what's the significance of these groups, how does this relate to "filename*", why is decoding required..)

TheAlmightyBob (changes requested):
Much better, thanks, just one more thing..

android/src/main/java/com/reactnativecommunity/webview/URLUtil.java
UNIDY2002: Thanks for your feedback.

According to the RFC, `m.group(4)` refers to the percent-encoded filename, and the charset is specified in `m.group(3)`; `m.group(2)` is the fallback filename.

I have also left some comments in the source code.

android/src/main/java/com/reactnativecommunity/webview/URLUtil.java
```diff
@@ -135,25 +135,32 @@ public static final String guessFileName(
...
-     * Unfortunately some servers do not quote the value so to maintain
-     * consistent behaviour with other browsers, we allow unquoted values too.
      */
     static String parseContentDisposition(String contentDisposition) {
         try {
+            // The regex attempts to match the following pattern:
+            //     attachment; filename="(Group 2)"; filename*=(Group 3)'(lang)'(Group 4)
+            // (Group 1 refers to the quotation marks around Group 2)
+            //
+            // Test cases can be found at http://test.greenbytes.de/tech/tc2231/
```
TheAlmightyBob, line 156: Thanks for including that! Seems to have other cases that are not supported, such as `filename=foo.html` (no quotes) or breaking things up with `filename*1`/`filename*2`...

But it doesn't look like your changes break support for a previously-working case, so that should be good.

TheAlmightyBob (approved):

android/src/main/java/com/reactnativecommunity/webview/URLUtil.java
UNIDY2002: Yes, that's the case. If there remain issues in the future regarding filename guessing, I would like to switch to a more robust implementation.

Thank you!

PR #2772, opened 2022-12-06
fix(Android): Apply WebView settings before loading content - take 2

TheAlmightyBob:
Same as #2766 but using `react-native-blob-util` instead of `react-native-file-access`, because I couldn't get the latter to build with macOS.

I'm not certain yet if this will work, mainly creating a PR to test it with the CI.


TheAlmightyBob:
Since the CI passed, I've pushed this change to #2766 and am closing this PR in favor of that one

PR #2809, opened 2023-01-16
Fixes for webview manager

ghalibt:


TheAlmightyBob:
Your branch is many versions behind. The change you're proposing was already made a few releases ago.

PR #2814, opened 2023-01-19
Fix clearCache link in index

Reisclef:
The id for the clearCache method is clearCachebool, but the index was incorrectly only referencing clearCache.


TheAlmightyBob (changes requested):
Thanks! Would you mind updating the Portuguese version while you're at it? Looks like it has the same issue

Reisclef:
Sure thing! Let me know if there's anything else. :)

TheAlmightyBob (approved):

TheAlmightyBob:
Argh. Obviously the CI failure is unrelated to your change... I'm _guessing_ it's a break from using the latest XCode version... (possibly combined with not using the latest version of react-native-macos)

Reisclef:
Thanks for merging back in the most recent changes @TheAlmightyBob . 👍

Issue #1070, opened 2019-12-09
WebView doesn't have momentum scrolling - IOS


avnerI:
The webView doesn't have momentum scrolling for IOS devices.
No matter if the destination URL has ` -webkit-overflow-scrolling: touch` css property set.

To Reproduce:
1. Any fresh installation of react-native
2. Install react-native-webview
3. Run on IOS device when the url is directing to a website that provides momentum scrolling in regular browser (Safari)

<img width="884" alt="Screen Shot 2019-12-09 at 16 46 22" src="https://user-images.githubusercontent.com/44140044/70445335-0b042080-1aa4-11ea-8fd9-9eb4f54b81ca.png">

Expecting to be able to scroll with momentum scrolling in IOS devices

Environment:
 - OS: IOS 
 - OS version: 11.0 and above 
 - react-native version: 0.61.5
 - react-native-webview version: 7.6.0

avnerI:
I've found this:
https://github.com/facebook/react-native/pull/5527
just add `<WebView .....
                            decelerationRate='normal'
...
/>
`

The naming and the change in file names on the IOS side made it hard to find

phyng:
@avnerI Thank you, you made my day!

Rutvij07:
The solution worked , lovely

Rajeshkanth:
Hi guys, im facing an issue which is the web view is rendered in ios app (flutter) the web view has heading, p tag, 4 inputs and a submit button which fixed in bottom of the page, when the input gets focus, the keyboard opens and after clicking on the second input the total page is coming down behind the keyboard, this issue is resolved if the 
<section> 
<form>
<h></h>
<p></p>
<div>
4 inputs with labels
</div>
</form>
</section>

now if the section is given as above 60rem for the height it is not coming behind the keyboard, and it is scrollable when keyboard presents, but there is a catch when cliking for the first time, the section is set to 60rem so it is more scrollable to end with white spaces, but i dont want that empty spaces at the end of the page, and the subsequent cliicks on the input makes the whitespaces gone (empty space ), i need help to rectify this.

saeid-elio:
The solution suggested by @avnerI should be added to Webview's documentation.

jslee-tpz:
> After upgrading my `react native` and `react-native-webview` to latest versions (0.78.0 and 13.13.2 respectively) the existing solution of `decelerationRate="normal"` started throwing following error on android phones:
> 
> ```
[...]

PR #2815, opened 2023-01-19
fix(iOS): setting correct default DecelerationRate for iOS to normal

ftlno:
- bugfix setting correct default DecelerationRate for iOS to normal rate if undefined. 
- updating wrong docs.

**Source:** developer.apple.com
https://developer.apple.com/documentation/uikit/uiscrollview/decelerationrate

### DIFF
--- a/src/WebView.ios.tsx
+++ b/src/WebView.ios.tsx
@@ -37,7 +37,7 @@ const processDecelerationRate = (
   decelerationRate: DecelerationRateConstant | number | undefined,
 ) => {
   let newDecelerationRate = decelerationRate;
-  if (newDecelerationRate === 'normal') {
+  if (newDecelerationRate === 'normal' || newDecelerationRate === undefined) {
     newDecelerationRate = 0.998;
   } else if (newDecelerationRate === 'fast') {
     newDecelerationRate = 0.99;
--- a/src/WebView.ios.tsx
+++ b/src/WebView.ios.tsx
@@ -37,7 +37,7 @@ const processDecelerationRate = (
   decelerationRate: DecelerationRateConstant | number | undefined,
 ) => {
   let newDecelerationRate = decelerationRate;
-  if (newDecelerationRate === 'normal' || newDecelerationRate === undefined) {
+  if (newDecelerationRate === 'normal') {
     newDecelerationRate = 0.998;
   } else if (newDecelerationRate === 'fast') {
     newDecelerationRate = 0.99;
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -133,6 +133,11 @@ RCTAutoInsetsProtocol>
     _autoManageStatusBarEnabled = YES;
     _contentInset = UIEdgeInsetsZero;
     _savedKeyboardDisplayRequiresUserAction = YES;
+    
+#if !TARGET_OS_OSX
+    _decelerationRate = UIScrollViewDecelerationRateNormal;
+#endif // !TARGET_OS_OSX
+
 #if !TARGET_OS_OSX
     _savedStatusBarStyle = RCTSharedApplication().statusBarStyle;
     _savedStatusBarHidden = RCTSharedApplication().statusBarHidden;
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -811,8 +811,8 @@ Example:
 
 A floating-point number that determines how quickly the scroll view decelerates after the user lifts their finger. You may also use the string shortcuts `"normal"` and `"fast"` which match the underlying iOS settings for `UIScrollViewDecelerationRateNormal` and `UIScrollViewDecelerationRateFast` respectively:
 
-- normal: 0.998
-- fast: 0.99 (the default for iOS web view)
+- normal: 0.998 (the default for iOS web view)
+- fast: 0.99 
 
 | Type   | Required | Platform |
 | ------ | -------- | -------- |
--- a/src/WebViewTypes.ts
+++ b/src/WebViewTypes.ts
@@ -438,8 +438,8 @@ export interface IOSWebViewProps extends WebViewSharedProps {
    * settings for `UIScrollViewDecelerationRateNormal` and
    * `UIScrollViewDecelerationRateFast` respectively:
    *
-   *   - normal: 0.998
-   *   - fast: 0.99 (the default for iOS web view)
+   *   - normal: 0.998 (the default for iOS web view)
+   *   - fast: 0.99
    * @platform ios
    */
   decelerationRate?: DecelerationRateConstant | number;


src/WebView.ios.tsx
```diff
@@ -37,7 +37,7 @@ const processDecelerationRate = (
   decelerationRate: DecelerationRateConstant | number | undefined,
 ) => {
   let newDecelerationRate = decelerationRate;
-  if (newDecelerationRate === 'normal') {
+  if (newDecelerationRate === 'normal' || newDecelerationRate === undefined) {
```
TheAlmightyBob, line 40: Is this really necessary? Looks like the native iOS code already handles this: https://github.com/react-native-webview/react-native-webview/blob/master/apple/RNCWebViewManager.m#L163

TheAlmightyBob (changes requested):
The Types fix makes sense, and looks like that should also be fixed in the documentation? (the Reference.md files)

src/WebView.ios.tsx
ftlno: It might be a better way to do it. But without it - the package defaults to fast and gets the wrong scroll :( Thank you for looking at it so fast! :)

TheAlmightyBob: > the package defaults to fast and gets the wrong scroll

You've verified this? Do you know how that's happening? Is the iOS code not working?

ftlno: Yes, I have ran the code in react native - iOS simulator and iOS 13 Pro device. 
Native WebView defaults to "normal", but this package defaults to "fast"

TheAlmightyBob: Ah, I see. The code I linked doesn't get called at all by default, and the `decelerationRate` that it is setting is a custom property of `RNCWebView`, which is not initialized to a default value (and thus being treated similar to "fast"). I think it might be better to initialize that variable to the "normal" constant (i.e. in `initWithFrame`, where other such properties are initialized to default values). Regardless of whether the default should be "fast" or "normal," using an uninitialized float just seems like a bug.

That does mean that this is more of a breaking change though... while "normal" does seem more correct and what might be expected by new users, it's possible that existing uses are relying on the "fast" behavior...

ftlno: Good point. I'm setting in upon `initWithFrame` now and its working as expected on Simulator(iPhone 14 iOS 16.2) 👍

And yes.. it could be that someone rely on the wrong default behaviour, but that anti pattern kinda deserve a breaking change. I bet way more people will be happy that their apps scroll work as native apps now.

TheAlmightyBob (approved):

TheAlmightyBob:
@Titozzz This "fix" makes sense, but thoughts on dealing with the "breaking change" aspect (changing the default, thus likely leading to different scrolling behavior for existing apps)? Would something like this justify a major version bump?

Titozzz:
> @Titozzz This "fix" makes sense, but thoughts on dealing with the "breaking change" aspect (changing the default, thus likely leading to different scrolling behavior for existing apps)? Would something like this justify a major version bump?

It does and when I merge it, it will mark as breaking

@ftlno I've been hard working on the new arch pr but hopelly u can go through all open PRs soon 😉

eirikluka:
It would still be great to see this merged

Titozzz:
Hello, after going through as many stack overflow and apple threads I could: 

From my understanding, WebView, even if it's using a scrollview has it's own default. See (https://stackoverflow.com/questions/14763255/why-does-scrolling-a-uiwebview-feel-so-much-different-than-scrolling-any-other#comment25190463_14763634)

Understanding this, I'm not sure that change make sense as the idea is to stick to native behavior as much as possible. Also users are free to set decelerationRate to normal if they want to. What we could probably improve is the documentation, explaining this.

WDYT?

TheAlmightyBob:
@Titozzz The thing is, right now we're not really using the WebView default I don't think. The code is explicitly setting the deceleration rate to an uninitialized float (which _maybe_ is treated the same as if we didn't set it at all? but I wouldn't bet on that..).

Point is valid though that maybe it's worth checking what the default value really is for the webview, in case it's different from a standalone scrollview (and update the documentation either way).

ewan-m:
It was my expectation that a standard web wrapper app such as this

```
export default function App() {
  return (
    <WebView
      style={styles.container}
      source={{ uri: "https://expo.dev" }}
    />
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    marginTop: Constants.statusBarHeight,
  },
});
```
would exactly match the scroll behaviour in the safari or chrome iOS apps. The fact it didn't felt janky and to me it's a hard fix to find that the `decelerationRate={0.998}` needs to be explicitly set. 

I made a similar starter app in SwiftUI using the same underlying WebView component and the scroll behaviour matches safari/chrome apps.

Titozzz:
React native WebView should match whatever the default of WebView is when building a native app with ObjC/Swift. Someone should create a test app to find out the default even though I expect it will be "deceleration: fast"

ewan-m:
> 

It's set to `UIScrollViewDecelerationRate(rawValue: 0.998)`

ewan-m:
> React native WebView should match whatever the default of WebView is when building a native app with ObjC/Swift. Someone should create a test app to find out the default even though I expect it will be "deceleration: fast"

It's set to 0.998 which corresponds to normal

Here's the logs and code from inside a swift app
```
UIScrollView.DecelerationRate.fast // = 0.99
UIScrollView.DecelerationRate.normal // = 0.998
WKWebView().scrollView.decelerationRate // = 0.998
```

eirikluka:
> I expect it will be "deceleration: fast"
"Fast" decelerates too fast compared to Safari. We had to override to "normal" in our app to get the expected scroll desceleration behaviour

TheAlmightyBob:
Has anybody tried just commenting out https://github.com/react-native-webview/react-native-webview/blob/891e595bcf5cafe4a903d5aa643e8361d0c3d467/apple/RNCWebViewImpl.m#L909 in a patch (and maybe having it log the current value instead of setting it), for science?

(I'm not ignoring the science @ewan-m already did above, I'm just curious to further double-check the "default" value in the context of `react-native-webview`)

TheAlmightyBob:
> that's a good suggestion quite curious too - but is there anything else to do to help getting this in?

It seemed to me like the main blocker was disagreement/confusion over what the correct default behavior would be. My suggestion was intended to help resolve that debate.

ewan-m:
> > that's a good suggestion quite curious too - but is there anything else to do to help getting this in?
> 
> It seemed to me like the main blocker was disagreement/confusion over what the correct default behavior would be. My suggestion was intended to help resolve that debate.

[https://github.com/ewan-m/webview-scroll-speeds](https://github.com/ewan-m/webview-scroll-speeds)

Here's a repo with a very simple webview app built in both expo and swift. Running on an iOS device you can noticeably feel the difference in scrolling behaviour. The swift app prints out the defaults too.

```
UIScrollView.DecelerationRate.fast // = 0.99
UIScrollView.DecelerationRate.normal // = 0.998
WKWebView().scrollView.decelerationRate // = 0.998
```

In order to make the expo app match the swift app you need to add `decelerationRate="normal"` or `decelerationRate={0.998}` to the props of the webview.

Funnily enough if you're using Apple Silicon (M1/M2) you can't notice the bug at all because there's a bug in the scroll behaviour of V14 Simulator itself [https://developer.apple.com/forums/thread/668488](https://developer.apple.com/forums/thread/668488) - I imagine this is where a lot of confusion must comes from!

Would you like a screen recording of both apps in my phone with examples of what changing the deceleration rate prop does?

ftlno:
I'm looking forward to seeing this getting merged. 

It seems like a anti pattern to make every developer writing webviews to understand that they should change this to **normal** to get the correct default behaviour. And understand that it breaks the UX of those who are used to using the wrong default, but its way better for future apps and developers to getting the right default.

eirikluka:
Is there anything else we can do to get this merged?

eirikluka:
@Titozzz @TheAlmightyBob – Please consider merging this fix

eirikluka:
This is very much a bug still. Are there any reasons why you won't implement this fix, @Titozzz @TheAlmightyBob ?

ewan-m:
hello! @Titozzz @TheAlmightyBob 

Would be great if this would be merged. I think it contributes greatly to the perception of react native having a poor performance because a lot of people starting a migration project from a web app will be likely to begin with a Webview. They'll then notice it "seems to struggle to scroll the webpage" and think hmm maybe react native has poor performance - but that's not true! just this default is wrong 😁

eirikluka:
The default is still wrong. Could we please get a response on this? @TheAlmightyBob @Titozzz

ftlno:
Any news?

ftlno:
Does this need any changes to be merged?

jamonholmgren:
I don't generally work on WebView anymore but will see what I can do here.

jamonholmgren:
I tested on Swift and Objective-C and it does indeed look like it's set to 0.998, which corresponds to `UIScrollViewDecelerationRateNormal`, aka `normal`.




Thibault and I discussed and will include it in the next major release, which will be the rewrite of `react-native-webview` that drops the old architecture.

In the meantime, just use the prop:

```tsx
<WebView
  // ...
  decelerationRate="normal"
  // ...
/>
```

There is a report that it [can throw errors on Android](https://github.com/react-native-webview/react-native-webview/issues/1070#issuecomment-2696784926), and this should work:

```tsx
<WebView
  // ...
  decelerationRate={0.998}
  // ...
/>
```

Thanks for your patience.

ewan-m:
Open sesame!

PR #2832, opened 2023-01-31
fix(ios): add missing `ignoreSilentHardwareSwitch`

wjaykim:
Added missing RCT_EXPORT_VIEW_PROPERTY declaration for `ignoreSilentHardwareSwitch` in `RNCWebViewManager.m`.

`ignoreSilentHardwareSwitch` functionality was not working because of the missing prop declaration in view manager.


TheAlmightyBob (approved):

TheAlmightyBob:
I think this is still valid, but will need to be updated because the target file was renamed from `RNCWebViewManager.m` -> `RNCWebViewManager.mm`

PR #2845, opened 2023-02-02
feat(ios): Added allowTextStyleOptions

Arjan-Zuidema:
This option makes it possible to enabled/disable the text style options on selected text.
This is useful when working with rich text editor (like Lexical) to disable certain styling using the native menu. We are currently using a `patch-package` in our app with this feature.

### DIFF
--- a/apple/RNCWebView.m
+++ b/apple/RNCWebView.m
@@ -64,6 +64,29 @@ NSString *const CUSTOM_SELECTOR = @"_CUSTOM_SELECTOR_";
 @end
 #endif // TARGET_OS_OSX
 
+#if !TARGET_OS_OSX
+@interface RNCWKWebView : WKWebView
+@end
+@implementation RNCWKWebView
+- (BOOL) canPerformAction:(SEL)action withSender:(id)sender
+{
+  NSString *sel = NSStringFromSelector(action);
+  RNCWebView *rncWebView = (RNCWebView *)[self superview];
+  RCTAssert([rncWebView isKindOfClass:[rncWebView class]], @"superview must be an RNCWebView");
+  
+  if(![rncWebView allowTextStyleOptions]) {
+    NSArray* textStyleOptions = @[@"toggleBoldface:", @"toggleItalics:", @"toggleUnderline:"];
+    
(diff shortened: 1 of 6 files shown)


TheAlmightyBob:
These options aren't already handled by the [custom menu](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#menuitems) support?

Arjan-Zuidema:
No, that adds options. I want to disable the structure menu item (Bold, Italic, Underline)

TheAlmightyBob:
My understanding is that it doesn't just add menu items anymore after https://github.com/react-native-webview/react-native-webview/pull/2610 ?

Arjan-Zuidema:
I just need the native menu (with copy/paste/select), but without "some" items like bold, italic, underline. I don't want a custom menu

apple/RNCWebView.m
```diff
@@ -64,6 +64,29 @@ - (void)scrollWheel:(NSEvent *)theEvent {
...
+@end
+@implementation RNCWKWebView
+- (BOOL) canPerformAction:(SEL)action withSender:(id)sender
+{
+  NSString *sel = NSStringFromSelector(action);
+  RNCWebView *rncWebView = (RNCWebView *)[self superview];
+  RCTAssert([rncWebView isKindOfClass:[rncWebView class]], @"superview must be an RNCWebView");
+  
+  if(![rncWebView allowTextStyleOptions]) {
+    NSArray* textStyleOptions = @[@"toggleBoldface:", @"toggleItalics:", @"toggleUnderline:"];
```
TheAlmightyBob, line 78: Perhaps this could be made more flexible by taking the array of options to remove as a property? So instead of `allowTextStyleOptions` it would be something like `suppressMenuItems={['toggleBoldFace', 'toggleItalics']}` (or `removeMenuItems`... or `menuItemsToRemove`... whatever) ?

Arjan-Zuidema: Yeah, sure! Looks like a good option. I will just close this PR and create a new one, since this is really outdated now en the feature changes a lot.

PR #2861, opened 2023-02-17
fix(Android): Don't crash while downloading file with % in filename

valentinaporch:
We ran into an issue where react-native-webview crashed on Android 13 while downloading a file with the "%" character in its name. It is similar to https://github.com/react-native-webview/react-native-webview/issues/2428, but the fix in v11.22.2 does not prevent the crash because the exception being thrown is different. Here is the stack trace:

`Fatal Exception: java.lang.SecurityException: java.io.IOException: Invalid file path
       at android.os.Parcel.createExceptionOrNull(Parcel.java:3011)
       at android.os.Parcel.createException(Parcel.java:2995)
       at android.os.Parcel.readException(Parcel.java:2978)
       at android.database.DatabaseUtils.readExceptionFromParcel(DatabaseUtils.java:190)
       at android.database.DatabaseUtils.readExceptionFromParcel(DatabaseUtils.java:142)
       at android.content.ContentProviderProxy.insert(ContentProviderNative.java:557)
       at android.content.ContentResolver.insert(ContentResolver.java:2193)
       at android.content.ContentResolver.insert(ContentResolver.java:2155)
       at android.app.DownloadManager.enqueue(DownloadManager.java:1119)
       at com.reactnativecommunity.webview.RNCWebViewModule.downloadFile(RNCWebViewModule.java:313)
       at com.reactnativecommunity.webview.RNCWebViewManager$2.onDownloadStart(RNCWebViewManager.java:252)
       at Ea.handleMessage(chromium-TrichromeWebViewGoogle6432.aab-stable-541411734:522)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:201)
       at android.os.Looper.loop(Looper.java:288)
       at android.app.ActivityThread.main(ActivityThread.java:7872)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)`

This fix replaces % and a few other invalid characters in the file name with "_". This fixes the issue in our app, resulting in a successful file download.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
@@ -99,7 +99,7 @@ class RNCWebViewManagerImpl {
                 Log.w(TAG, "Unsupported URI, aborting download", e)
                 return@DownloadListener
             }
-            val fileName = URLUtil.guessFileName(url, contentDisposition, mimetype)
+            var fileName = URLUtil.guessFileName(url, contentDisposition, mimetype)
 
             // Sanitize filename by replacing invalid characters with "_"
             fileName = fileName.replace(invalidCharRegex, "_")
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
@@ -29,6 +29,8 @@ import java.net.MalformedURLException
 import java.net.URL
 import java.util.*
 
+val invalidCharRegex = "[\\\\/%\"]".toRegex()
+
 
 class RNCWebViewManagerImpl {
     companion object {
@@ -98,6 +100,10 @@ class RNCWebViewManagerImpl {
                 return@DownloadListener
             }
             val fileName = URLUtil.guessFileName(url, contentDisposition, mimetype)
+
+            // Sanitize filename by replacing invalid characters with "_"
+            fileName = fileName.replace(invalidCharRegex, "_")
+
             val downloadMessage = "Downloading $fileName"
 
             //Attempt to add cookie, if it exists
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModuleImpl.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModuleImpl.java
@@ -36,6 +36,7 @@ import com.facebook.react.modules.core.PermissionListener;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.SecurityException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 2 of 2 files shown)


TheAlmightyBob:
What about catching the exception?

android/src/main/java/com/reactnativecommunity/webview/URLUtil.java
```diff
@@ -84,9 +84,12 @@ public static final String guessFileName(
             }
         }
 
-        // Finally, if couldn't get filename from URI, get a generic filename
         if (filename == null) {
+            // If couldn't get filename from URI, get a generic filename
             filename = "downloadfile";
+        } else {
+            // Sanitize filename by replacing invalid characters with "_"
+            filename = filename.replaceAll("[\\\\/%\"]", "_");
```
TheAlmightyBob, line 92: Would it make sense to move this logic to where the `Request` is configured in `RNCWebViewManager`?

As the comments at the top of this file indicate, it is taken from https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/webkit/URLUtil.java and mostly unmodified.

There is one exception, but that is explicitly noted in the comments and would not have worked outside of this file... whereas this sanitization change seems like it would be safe to relocate and run on the string returned by `guessFileName`?

(if not, then I'd ask that it at least be added to the comments calling out modifications to the file)

TheAlmightyBob: Yep, if `URLUtil` was an original file of `react-native-webview` I would totally agree with your logic, I'm just biased towards keeping it a "pure" copy of the Android WebKit version as much as possible...

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -230,7 +230,7 @@ public void onDownloadStart(String url, String userAgent, String contentDisposit
           return;
         }
 
-        String fileName = URLUtil.guessFileName(url, contentDisposition, mimetype);
+        String fileName = URLUtil.guessFileName(url, contentDisposition, mimetype).replaceAll("[\\\\/%\"]", "_");
```
TheAlmightyBob, line 233: Could you add back the comment explaining why the `replaceAll` is being done?

TheAlmightyBob: I do actually think a separate line would be more readable if you don't mind 🙂

TheAlmightyBob (approved):

TheAlmightyBob:
@valentinaporch While this looks good and I would consider "in the queue" to be merged (and I would like this fix also, so appreciate your finding and fixing it!), we've been pausing some merges to minimize conflicts with #2686, so I'm not yet certain when this will get in.

TheAlmightyBob:
@valentinaporch What happened to this PR? The "new architecture" PR finally went in, so others are no longer blocked on that

TheAlmightyBob (approved):

TheAlmightyBob:
Thanks @valentinaporch! Sorry for the delay, busy times.

PR #2888, opened 2023-03-14
fix: update `userAgent` reference

TheKohan:
I was trying to set a `user-agent` with version `"react-native-webview": "11.26.1"` and was mislead by the `Reference.md`. Seems like it only applies to `Android`.

Code:
image

Docs:
image

@Edit 

It seems like i was wrong, only type comment needs to be updated.


TheAlmightyBob:
@TheKohan The `userAgent` properties _are_ available on iOS/macOS as well. It looks like it's the comments on the types that need updating. You can see [here](https://github.com/react-native-webview/react-native-webview/blob/1707328bda42bd24bfd0a3d01d79c816b20948a5/src/WebViewTypes.ts#L564) that it's part of the iOS props type, it's just missing the `@platform ios` annotation.

TheKohan:
You're right, sorry for confusion, gonna update it.

TheAlmightyBob (approved):

TheKohan:
Hi @Titozzz , @jamonholmgren, can we merge this ? Thanks.

PR #2896, opened 2023-03-16
feat(android): add androidDebuggingEnabled prop

Tom-Bury:
Hi!

Proposal PR to add a prop to the `Webview` to allow enabling/disabling Android debugging. Makes it a bit easier instead of having to patch the native code as currently described in the [Debugging](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Debugging.md#android--chrome) docs.

Debugging is already enabled by default in debug builds. This change should allows to have it also in release builds, as well as being able to toggle it on the fly.

Thanks for reviewing!

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -684,6 +684,13 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
     }
   }
 
+  @ReactProp(name = "androidDebuggingEnabled")
+  public void setAndroidDebuggingEnabled(WebView view, boolean enabled) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
+      view.setWebContentsDebuggingEnabled(enabled);
+    }
+  }
+
   @Override
   protected void addEventEmitters(ThemedReactContext reactContext, WebView view) {
     // Do not register default touch emitter and let WebView implementation handle touches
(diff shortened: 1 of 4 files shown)


TheAlmightyBob:
Would setting this to `false` disable debugging for debug builds?

Technically the current documentation is not suggesting patching anything, just a one-line change to your application code... but perhaps it's more complicated if you're using Expo?

Tom-Bury:
Seems like I misunderstood the docs on the patch, sorry! I read it as needing to put it as a patch in the webview Android code, but indeed you can just put it in your app's native code. Not sure about whether that's more difficult to do on Expo, as I haven't used that yet. It is however not easily changeable at runtime + requires rebuilding the app.

On the behaviour in debug builds, there seems to be different behaviour on emulators versus real devices:

- Emulators always allow debugging for both debug & release builds, regardless of how the prop is set. I haven't found any sources confirming, but I assume it's because it's an emulator. https://user-images.githubusercontent.com/22749794/226468469-6d361cfb-e37c-4b18-bb33-3df83223ece0.mov

- Devices + debug builds will allow debugging initially when the prop is not set / undefined, because it is enabled in `RNCWebViewManager.createViewInstance@216` https://user-images.githubusercontent.com/22749794/226468657-59d1508a-932f-4322-9f13-fdad4b6adb47.mov

- Devices + release builds will not allow debugging initially when the prop is not set https://user-images.githubusercontent.com/22749794/226468773-2ce05151-5dcc-4759-b2a3-986bc794a612.mov

I've tested this quickly in a sample project on RN71, (source code here)[https://github.com/Tom-Bury/rn-example-app], you can also find the debug & release APKs in there

Tom-Bury:
If you want I can try adding it as well. Am not an iOS dev but seems straightforward enough

PR #2897, opened 2023-03-17
[windows] Support headers and cookies in source prop

asvahagnas:
Supporting passing headers in **source** property for WebView2.
You also can pass cookies in headers. The passed cookie will be treated same way as http response  **Set-Cookie** header.
It will be persisted on WebView2.

Example of passing cookies via headers.

`<WebView
  useWebView2={true}
  source={{uri: 'https://fb.workplace.com/app-embed/canvas/211801378174677?mh_p_pixel_ratio=1&js_debug=1&gk_debug=1&sr_sourcemaps=1',
  headers: {
    cookie: "cookie1=value1; Domain=.example.com; Expires=Wed, 13 Mar 2024 17:16:14 GMT;  Path=/; Secure;, cookie2=value2; Domain=.example.com; Expires=Wed, 13 Mar 2024 17:16:14 GMT;  Path=/; Secure;"
  }}}
  style={styles.webView}
/>`


chiaramooney (approved):
Look good to me!

PR #2899, opened 2023-03-20
[NewArchBranch] Allow passing a custom RNCWebView subclass to RNCWebViewManager's createViewInstance (old architecture only)

TheAlmightyBob:
@Titozzz 

(as I said on Slack, I'm open to better ideas, but this seemed like the simplest way to get things working again)

### DIFF
--- a/android/src/oldarch/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/oldarch/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -47,6 +47,10 @@ public class RNCWebViewManager extends SimpleViewManager<RNCWebView> {
         return mRNCWebViewManagerImpl.createViewInstance(context);
     }
 
+    public RNCWebView createViewInstance(ThemedReactContext context, RNCWebView webView) {
+      return mRNCWebViewManagerImpl.createViewInstance(context, webView);
+    }
+
     @ReactProp(name = "allowFileAccess")
     public void setAllowFileAccess(RNCWebView view, boolean value) {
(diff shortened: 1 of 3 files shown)


android/src/oldarch/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -47,6 +47,10 @@ public RNCWebView createViewInstance(ThemedReactContext context) {
         return mRNCWebViewManagerImpl.createViewInstance(context);
     }
 
+    public RNCWebView createViewInstance(ThemedReactContext context, RNCWebView webView) {
+      return mRNCWebViewManagerImpl.createViewInstance(context, webView);
+    }
+
```
Titozzz, line 53: Is this necessary? wouldn't just people overriding things replace the first one with a custom one?

Titozzz (approved):

android/src/oldarch/com/reactnativecommunity/webview/RNCWebViewManager.java
TheAlmightyBob: Sorry I didn't see this sooner.

The reason for this was that `mRNCWebViewManagerImpl` is declared `private final`, so subclasses don't have access to call `createViewInstance` on it. This way allows overriding the single-parameter version and then using `super` to invoke the two-parameter version.

Making it `protected` instead of `private` would also have been an option. I'm not sure if there's much advantage to one way or the other. I think I just initially assumed that the use of `private` was an intentional choice and so this would be a less-controversial approach.

PR #2930, opened 2023-04-18
[iOS] Fix webview not loading in background tab

peterlazar1993:
This is the same fix as #2626 on top of master.

Fixes: 
- #2431 

Reference: 
- #2626 

Credits to [sjransom](https://github.com/sjransom)


Santhosh-Umapathi:
`"react-native-webview": "^13.10.0"`

Still the same issue, has to be reopened

LoadJulz:
`"react-native-webview": "^13.10.3" `

Same issue

badalsaibo:
It says merged but the master still has the old code.

```
- (void)didMoveToWindow
{
```

https://github.com/react-native-webview/react-native-webview/blob/1ddfe70521725c365cf8accf2a1bdf82eb4db80f/apple/RNCWebViewImpl.m#L515

Issue #2942, opened 2023-04-26
Motion sensor permissions are asked repetitively 


ThorstenBux:
Hi there, 
I'm new to RN and react-native-webview. However, I've worked on several project in RectJS. 
**Bug description:**
I'm reporting this issue cause I'm including a ReactJS project into a ReactNative app using the react-native-webview.
The Web page will use the motion sensor so I've setup the ReactNative App to request permissions and expect the permissions to be available to the WebView, however, this doesn't seem to be the case.

**To Reproduce:**

- Create a ReactNative App
- Install react-native-permissions (https://github.com/zoontek/react-native-permissions)
- Request access to the motion sensor using the react-native-permissions library
```
          RNPermissions.request(PERMISSIONS.IOS.MOTION)
            .then(status => {
              console.log('request(MOTION)', status);
            })
            .catch(error => {
              console.error(error);
            });
```
- Allow access to the sensor 
- Launch the WebView and trigger sensor readings

**Expected behavior:**
Expected: No permission dialog to come up because I have just granted access to the motion sensor.

**Screenshots/Videos:**

https://user-images.githubusercontent.com/16772796/234722782-f76aa2cc-ab05-4dcc-ad1a-7a5e5a410e16.mp4

GitHub repo with demo: https://github.com/ThorstenBux/reactNativeMotion

**Environment:**
 - OS: iOS Iphone 14 Pro (actual device not simulator)
 - OS version: 16.2.1
 - react-native version: 0.71.7
 - react-native-webview version: 12.0.2

ThorstenBux:
Hi guys, was anyone able to reproduce? Any pointers into what is going on?

priyanshi-vindaloo:
yes I am also facing this issue.

PR #2944, opened 2023-05-04
adding proper usage of Motion and Orientation permissions for iOS.

ThorstenBux:
see https://github.com/react-native-webview/react-native-webview/issues/2942

The solution is a bit crude as it relies on the 'mediaCapturePermissionGrantType' prop it should probably get it's own prop. Asking for guidance on how to clean up the implementation as I'm not (yet) an iOS developer


ThorstenBux:
@Titozzz anything you need on this one?

PR #2957, opened 2023-05-10
implement injectJavaScriptBeforeContentLoaded method

kevinvangelder:
Supports iOS, Android, and macOS.

Use case: We're injecting a token into the webview, but when the token nears it's expiration we refresh it and reload the page. `injectJavaScript` is insufficient as the code is lost when the page is reloaded. `injectedJavaScriptBeforeContentLoaded` prop changes appear to be ignored. This new method allows us to set the new token in the webview before reloading, ensuring it will be available the instant the page finishes loading.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
@@ -273,6 +273,7 @@ class RNCWebViewManagerImpl {
     val COMMAND_INJECT_JAVASCRIPT = 6
     val COMMAND_LOAD_URL = 7
     val COMMAND_FOCUS = 8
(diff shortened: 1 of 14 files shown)


jamonholmgren:
Thanks Kevin!

Would it be better to change `injectedJavaScriptBeforeContentLoaded` to properly apply those changes when the prop changes?

There's a subtle but important difference between the `injectJavaScript` method and this new `injectJavaScriptBeforeContentLoaded` method; in the first, the JS is executed immediately, but in the second, you're essentially just making sure that `injectedJavaScriptBeforeContentLoaded` gets changed to a different value internally, so it's not executed immediately.

It seems like making the `injectedJavaScriptBeforeContentLoaded` prop changes apply to subsequent page loads would fix your issue and wouldn't require adding a new method.

jamonholmgren:
Closing in favor of #2958.

PR #2958, opened 2023-05-11
fix iOS injectedJavaScriptBeforeContentLoaded

kevinvangelder:
This PR fixes iOS injectedJavaScriptBeforeContentLoaded to correctly update the injected javascript when the provided props change.

### DIFF
--- a/.github/workflows/android-ci.yml
+++ b/.github/workflows/android-ci.yml
@@ -9,31 +9,28 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
+        newArchEnabled: [false, true]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Set up Node.js
-        uses: actions/setup-node@v2.5.1
+        uses: actions/setup-node@v2
         with:
-          node-version: 16
+          node-version: 18
+          cache: 'yarn'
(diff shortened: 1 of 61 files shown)


jamonholmgren:
I think this is headed the right direction!

Talking with @Titozzz about it right now.

Couple things we'll want to see for sure:

1. Update [Guide](https://github.com/react-native-webview/react-native-webview/blob/681aac4fe7e35834264daa01b67a9893c4a9ebe7/docs/Guide.md?plain=1#L339), including a good solid example of how to use the JS object
2. Update [other docs](https://github.com/react-native-webview/react-native-webview/blob/681aac4fe7e35834264daa01b67a9893c4a9ebe7/docs/Reference.md?plain=1#LL193C17-L193C17) as necessary, including adding the JS object thing

(These were just 2 I found with a quick search.)

kevinvangelder:
@Titozzz any feedback on this PR? I need to incorporate it in an app later this week and would love to move forward or update it with any feedback ASAP.

Titozzz:
@kevinvangelder if it is due this week, you should use patch package for your needs. I do not have a lot of bandwidth this week

kevinvangelder:
@Titozzz Thanks for letting me know.

kevinvangelder:
@Titozzz It does appear that it should, but in my tests changing the value of `injectedJavaScriptBeforeContentLoaded` and then reloading the page does not result in the updated code being run on the webview. After my changes, it does.

PR #2960, opened 2023-05-12
Android injectJavaScriptObject

kevinvangelder:
I've added an injectJavaScriptObject prop for Android that allows passing any object from the JSX/TSX through the native code and into the browser. In my testing it was consistently available before the webpage loaded, unlike injectedJavaScriptBeforeContentLoaded on Android. I'm open to adjusting the naming, etc.

Usage:
```tsx
<WebView
    injectJavaScriptObject={{authToken: 'myValue'}}
/>
```

```js
<script>
    window.onload = (event) => {
        JSON.parse(injectedObject.toString()).authToken;
    }
</script>
```

My Java-foo is not all that fantastic, so I was not able to figure out how to bypass the `.toString()` requirement and just have the contents be available directly, but open to suggestions.

Thanks to @morganick for assisting.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
@@ -219,7 +219,7 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
         }
     }
 
-    public void injectJSObject(String obj) {
+    public void injectJavaScriptObject(String obj) {
         if (getSettings().getJavaScriptEnabled()) {
             WebView webView = this;
             webView.addJavascriptInterface(new JsObject(obj), "injectedObject");
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
@@ -37,21 +37,13 @@ import com.reactnativecommunity.webview.events.TopMessageEvent;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 
- class JsObject {
-    protected String value;
-    public JsObject(String newValue) {
-        value = newValue;
-    }
-    @JavascriptInterface
-    public String toString() { return value; }
- }
-
 public class RNCWebView extends WebView implements LifecycleEventListener {
     protected @Nullable
     String injectedJS;
     protected @Nullable
     String injectedJSBeforeContentLoaded;
     protected static final String JAVASCRIPT_INTERFACE = "ReactNativeWebView";
+    RNCWebViewBridge bridge;
 
     /**
      * android.webkit.WebChromeClient fundamentally does not support JS injection into frames other
@@ -172,8 +164,11 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
         return mRNCWebViewClient;
     }
 
-    protected RNCWebViewBridge createRNCWebViewBridge(RNCWebView webView) {
-        return new RNCWebViewBridge(webView);
+    protected RNCWebViewBridge getRNCWebViewBridge(RNCWebView webView) {
+        if (!bridge) {
+            bridge = new RNCWebViewBridge(webView);
+        }
+        return bridge;
     }
 
     protected void createCatalystInstance() {
@@ -193,7 +188,7 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
         messagingEnabled = enabled;
 
         if (enabled) {
-            addJavascriptInterface(createRNCWebViewBridge(this), JAVASCRIPT_INTERFACE);
+            addJavascriptInterface(getRNCWebViewBridge(this), JAVASCRIPT_INTERFACE);
         } else {
             removeJavascriptInterface(JAVASCRIPT_INTERFACE);
         }
@@ -221,8 +216,9 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
 
     public void injectJavaScriptObject(String obj) {
         if (getSettings().getJavaScriptEnabled()) {
-            WebView webView = this;
-            webView.addJavascriptInterface(new JsObject(obj), "injectedObject");
+            RNCWebViewBridge b = getRNCWebViewBridge(this);
+            b.setInjectedObject(obj)
+            addJavascriptInterface(b, JAVASCRIPT_INTERFACE);
         }
     }
 
@@ -323,9 +319,14 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
 
   protected class RNCWebViewBridge {
         RNCWebView mWebView;
+        String injectedObject;
 
         RNCWebViewBridge(RNCWebView c) {
           mWebView = c;
+        }
+
+        setInjectedObject(String s) {
+            injectedObject = s;
         }
 
         /**
@@ -336,6 +337,9 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
         public void postMessage(String message) {
           mWebView.onMessage(message);
         }
+
+        @JavascriptInterface
+        public String injectedObject() { return value; }
     }
 
 
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
@@ -37,6 +37,15 @@ import com.reactnativecommunity.webview.events.TopMessageEvent;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 
+ class JsObject {
+    protected String value;
+    public JsObject(String newValue) {
+        value = newValue;
+    }
+    @JavascriptInterface
+    public String toString() { return value; }
+ }
+
 public class RNCWebView extends WebView implements LifecycleEventListener {
     protected @Nullable
     String injectedJS;
@@ -207,6 +216,13 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
                 injectedJSBeforeContentLoaded != null &&
                 !TextUtils.isEmpty(injectedJSBeforeContentLoaded)) {
             evaluateJavascriptWithFallback("(function() {\n" + injectedJSBeforeContentLoaded + ";\n})();");
+        }
+    }
+
+    public void injectJSObject(String obj) {
+        if (getSettings().getJavaScriptEnabled()) {
+            WebView webView = this;
+            webView.addJavascriptInterface(new JsObject(obj), "injectedObject");
         }
     }
 
--- a/android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -181,6 +181,11 @@ public class RNCWebViewManager extends SimpleViewManager<RNCWebView>
 
     }
 
+    @ReactProp(name = "injectJSObject")
+    public void setInjectJSObject(RNCWebView view, @Nullable String value) {
+        mRNCWebViewManagerImpl.setInjectJSObject(view, value);
+    }
+
     @Override
     @ReactProp(name = "javaScriptCanOpenWindowsAutomatically")
     public void setJavaScriptCanOpenWindowsAutomatically(RNCWebView view, boolean value) {
--- a/android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -182,8 +182,8 @@ public class RNCWebViewManager extends SimpleViewManager<RNCWebView>
     }
 
     @ReactProp(name = "injectJavaScriptObject")
-    public void setInjectJavaScriptObject(RNCWebView view, @Nullable String value) {
-        mRNCWebViewManagerImpl.setInjectJavaScriptObject(view, value);
+    public void setInjectedJavaScriptObject(RNCWebView view, @Nullable String value) {
+        mRNCWebViewManagerImpl.setInjectedJavaScriptObject(view, value);
     }
 
     @Override
--- a/android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -181,9 +181,9 @@ public class RNCWebViewManager extends SimpleViewManager<RNCWebView>
 
     }
 
-    @ReactProp(name = "injectJSObject")
-    public void setInjectJSObject(RNCWebView view, @Nullable String value) {
-        mRNCWebViewManagerImpl.setInjectJSObject(view, value);
+    @ReactProp(name = "injectJavaScriptObject")
+    public void setInjectJavaScriptObject(RNCWebView view, @Nullable String value) {
+        mRNCWebViewManagerImpl.setInjectJavaScriptObject(view, value);
     }
 
     @Override
--- a/docs/Guide.md
+++ b/docs/Guide.md
@@ -375,16 +375,16 @@ By setting `injectedJavaScriptBeforeContentLoadedForMainFrameOnly: false`, the J
 > Note on Android Compatibility: For applications targeting `Build.VERSION_CODES.N` or later, JavaScript state from an empty WebView is no longer persisted across navigations like `loadUrl(java.lang.String)`. For example, global variables and functions defined before calling `loadUrl(java.lang.String)` will not exist in the loaded page. Applications should use the Android Native API `addJavascriptInterface(Object, String)` instead to persist JavaScript objects across navigations.
 
 
-#### the `injectJavaScriptObject` prop (Android Only)
+#### The `injectJavaScriptObject` prop (Android Only)
 
-Due to the Android race condition mentioned above, this more reliable prop was added. While you cannot execute arbitrary JavaScript, you can make any JS object available to the JS run in the webview prior to the page load completing.
+Due to the Android race condition mentioned above, this more reliable prop was added. While you cannot execute arbitrary JavaScript, you can make an arbitrary JS object available to the JS run in the webview prior to the page load completing.
 
 ```html
 <html>
   <head>
     <script>
       window.onload = (event) => {
-        document.getElementById('output').innerHTML = JSON.parse(injectedObject.toString()).authToken;
+        document.getElementById('output').innerHTML = JSON.parse(window.RNCWebViewBridge.injectedObject()).customValue;
       }
     </script>
   </head>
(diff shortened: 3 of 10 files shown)


android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -210,6 +219,13 @@ public void callInjectedJavaScriptBeforeContentLoaded() {
         }
     }
 
+    public void injectJavaScriptObject(String obj) {
+        if (getSettings().getJavaScriptEnabled()) {
+            WebView webView = this;
+            webView.addJavascriptInterface(new JsObject(obj), "injectedObject");
```
TheAlmightyBob, line 225: `injectedObject` may be a _little_ general, considering folks might not fully control the web content being loaded in their apps' WebViews.... might be worth calling it something like `ReactNativeWebViewInjectedObject` (or something shorter but specific...)

I actually wonder if it would be possible to leverage the existing `RNCWebViewBridge` instead of creating a new `JsObject` interface... then instead of `injectedObject.toString()` you could be calling `window.ReactNativeWebView.injectedObject()`...?

android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -181,6 +181,11 @@ public void setInjectedJavaScriptBeforeContentLoadedForMainFrameOnly(RNCWebView
 
     }
 
+    @ReactProp(name = "injectJavaScriptObject")
+    public void setInjectJavaScriptObject(RNCWebView view, @Nullable String value) {
```
TheAlmightyBob, line 185: Probably should be _injected_ rather than _inject_, since this is a property, not a function

docs/Guide.md
```diff
@@ -366,14 +366,57 @@ export default class App extends Component {
...
 > On Android, `injectedJavaScript` runs a method on the Android WebView called `evaluateJavascriptWithFallback`
 > Note on Android Compatibility: For applications targeting `Build.VERSION_CODES.N` or later, JavaScript state from an empty WebView is no longer persisted across navigations like `loadUrl(java.lang.String)`. For example, global variables and functions defined before calling `loadUrl(java.lang.String)` will not exist in the loaded page. Applications should use the Android Native API `addJavascriptInterface(Object, String)` instead to persist JavaScript objects across navigations.
 
+
+#### the `injectJavaScriptObject` prop (Android Only)
```
TheAlmightyBob, line 378:
```suggestion
#### The `injectJavaScriptObject` prop (Android Only)
```

docs/Guide.md
```diff
@@ -366,14 +366,57 @@ export default class App extends Component {
...
+
+Due to the Android race condition mentioned above, this more reliable prop was added. While you cannot execute arbitrary JavaScript, you can make any JS object available to the JS run in the webview prior to the page load completing.
```
TheAlmightyBob, line 380: "_any_ JS object" is probably a stretch... e.g., if the original JS object had functions, those wouldn't work...

docs/Reference.md
```diff
@@ -247,6 +247,40 @@ If `false`, (only supported on iOS and macOS), loads it into all frames (e.g. if
...
+
+Inject any JavaScript object into the webview so it is available to the JS running on the page.
+
+| Type | Required | Platform                                          |
+| ---- | -------- | ------------------------------------------------- |
+| obj | No       | Android only |
+
+Example:
+
+Set an auth token to be used in JavaScript.
```
TheAlmightyBob, line 260: In general the docs should warn that this object will be accessible to all frames of the web page.... this auth token scenario is a good example of where that might be a security risk for some apps. (and thus _might_ be a bad example...)

TheAlmightyBob (changes requested):

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -184,7 +188,7 @@ public void setMessagingEnabled(boolean enabled) {
         messagingEnabled = enabled;
 
         if (enabled) {
-            addJavascriptInterface(createRNCWebViewBridge(this), JAVASCRIPT_INTERFACE);
+            addJavascriptInterface(getRNCWebViewBridge(this), JAVASCRIPT_INTERFACE);
         } else {
             removeJavascriptInterface(JAVASCRIPT_INTERFACE);
         }
```
kevinvangelder, line 194: With this updated implementation, I'm concerned about the possibility of RNCWebViewBridge being removed from the webview if messaging is disabled, and I'm interested in suggestions for a preferred way to address them.

TheAlmightyBob: Ah, yes. That is a valid concern I had not considered.

I could imagine perhaps refactoring this to have a separate `javascriptInterfaceEnabled` (or `webViewBridgeEnabled`) type of member variable that's used to manage the `addJavascriptInterface`/`removeJavascriptInterface` calls, then that would be set according to the combination of the `messagingEnabled`/`injectedObject` variables....?

(and the actual implementations would need to be guarded by their respective properties... e.g., `onMessage` shouldn't do anything if called when `messagingEnabled` is `false`...)

I'm still open to considering separate interfaces, I just thought the single interface would be cleaner.

TheAlmightyBob:
What the heck, bot.

kevinvangelder:
@TheAlmightyBob Thanks for reopening, I still need to work through this with @jamonholmgren to see if we can come up with a good solution.

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
kevinvangelder: @TheAlmightyBob @jamonholmgren and I worked together to come up with a solution to this dilemma. Now, rather than removing the javascript interface, we check if messaging is enabled when a message comes through. There is a very narrow window where this might break existing code if a website were relying on checks for the existence of `window.RNCWebViewBridge` to determine whether or not it's expecting the app to catch messages, but in all other scenarios we could think of it shouldn't cause any issues.

TheAlmightyBob: > There is a very narrow window where this might break existing code if a website were relying on checks for the existence of window.RNCWebViewBridge to determine whether or not it's expecting the app to catch messages

That does seem like a potentially breaking change, which I was trying to avoid with the idea of "still remove the interface if _neither_ `messagingEnabled` _nor_ `injectedObject` are set" (so the behavior would be unchanged if people don't touch `injectedObject`).

There's also the small possibility of this being a security concern... Google themselves [warn](https://developer.android.com/develop/ui/views/layout/webapps/webview#BindingJavaScript) that `addJavascriptInterface`
> can be a very useful feature or a dangerous security issue

(And I previously had to review `react-native-webview`'s use of it with my company's security team)

I don't think what you're proposing changes the _actual_ security risk, I just wonder if people would be concerned about `addJavascriptInterface` being invoked even if they're not enabling features that require it...


If @jamonholmgren / @Titozzz think that's all fine and are comfortable merging this as a breaking change, I won't object. But I'm not personally able to merge breaking changes.

kevinvangelder: @TheAlmightyBob The way the new code works, if neither `messagingEnabled` (aka `onMessage`) nor `injectedObject` are set, the interface does not get added to the webview at all, so in that regard the behavior remains unchanged. It's only if `onMessage` is set and then subsequently unset where the behavior is different (an extreme edge case, IMO), other than using `injectedObject` obviously.

TheAlmightyBob: Oh. Okay, I'll take a look at the new code later, my comment above was only responding to my interpretation of your comment, not the actual changes.

jamonholmgren: Yeah it's a very edge-case situation where behavior would change (adding and then removing an existing `onMessage` prop, PLUS relying on the existence of `window.RNCWebVewBridge` to tell them ... something?). If a single user reports an issue I'd be extremely surprised. :joy:

We also protect against messages being sent in the code by checking `messagingEnabled` which is helpful for security concerns.

jamonholmgren:
This is feeling mergable, to me. And I don't think it requires a breaking change label. No documented behavior has changed.

Titozzz:
What's 'annoying' to me with that PR is that it's been created because InjectBeforeLoaded is bugged..

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -55,6 +57,7 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
     protected @Nullable
     String injectedJSBeforeContentLoaded;
     protected static final String JAVASCRIPT_INTERFACE = "ReactNativeWebView";
+    RNCWebViewBridge bridge;
```
TheAlmightyBob, line 60: nit: Shouldn't this include `protected @Nullable`?

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -264,11 +275,7 @@ public void setMessagingEnabled(boolean enabled) {
 
         messagingEnabled = enabled;
 
-        if (enabled) {
-            addJavascriptInterface(createRNCWebViewBridge(this), JAVASCRIPT_INTERFACE);
-        } else {
-            removeJavascriptInterface(JAVASCRIPT_INTERFACE);
-        }
+        createRNCWebViewBridge(this);
```
TheAlmightyBob, line 278: Seems like this should be gated with `if (enabled)`?

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -291,6 +298,13 @@ public void callInjectedJavaScriptBeforeContentLoaded() {
         }
     }
 
+    public void injectedJavaScriptObject(String obj) {
```
TheAlmightyBob, line 301: nit: should be `setInjectedJavaScriptObject`?

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -387,20 +401,33 @@ public ThemedReactContext getThemedReactContext() {
   }
 
   protected class RNCWebViewBridge {
+        private String TAG = "RNCWebViewBridge";
         RNCWebView mWebView;
+        String injectedObject;
```
TheAlmightyBob, line 406: ultra-nit: I wonder if it would be worth calling this `injectedObjectJson`? It seems potentially confusing that it's called an "object" but is a string...

TheAlmightyBob (changes requested):

TheAlmightyBob:
> I don't think it requires a breaking change label.

Yeah my suggestion there was for if the change is going to be to _always_ use `addJavascriptInterface`, even if no relevant features have been enabled.

(although my read of the current code in the PR suggests that it _will_ do that... but it sounds like that wasn't the intent)

If it only adds the interface when a relevant feature has been "activated," and the change is just that it doesn't remove it upon deactivation, then I agree that seems like much more of an edge case.

kevinvangelder:
@Titozzz That's fair, but the InjectBeforeLoaded issue is an Android issue, not an RNWV issue.

@TheAlmightyBob Thanks for the feedback.

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
jamonholmgren: Good idea.

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -387,20 +404,33 @@ public ThemedReactContext getThemedReactContext() {
...
-          mWebView.onMessage(message);
+            if (mWebView.getMessagingEnabled()) {
+                mWebView.onMessage(message);
+            } else {
+                FLog.w(TAG, "ReactNativeWebView.postMessage method was called but messaging is disabled. Pass an onMessage handler to the WebView.");
+            }
         }
+
+        @JavascriptInterface
+        public String injectedObject() { return injectedObjectJson; }
```
TheAlmightyBob, line 433: I would vote for the getter/setter to also use the `Json` suffix (especially the getter, since that's what's exposed to the Javascript in the WebView, which needs to know that this is JSON to be parsed?)

TheAlmightyBob: nit: I also wonder if it would be useful for this to use the `get` prefix, just to make clear when consuming in the Javascript side that it's a function, not a property? I don't feel strongly about that though

docs/Reference.md
```diff
@@ -249,6 +249,42 @@ If `false`, (only supported on iOS and macOS), loads it into all frames (e.g. if
...
+
+| Type | Required | Platform                                          |
+| ---- | -------- | ------------------------------------------------- |
+| obj | No       | Android only |
+
+Example:
+
+Set a value to be used in JavaScript.
+
+Note: Any value in the object will be accessible to *all* frames of the webpage. If sensitive values are present please ensure that you have a strict Content Security Policy set up to avoid data leaking.
```
TheAlmightyBob, line 264: Good note, thanks!

TheAlmightyBob (changes requested):
Allllllmost there!

I'm sorry I didn't mention this earlier, but: it would be _realllllly_ nice to update the example app to include this. There's already an `Injection.tsx` screen that could probably be added to...?

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
kevinvangelder: `window.ReactNativeWebView.injectedObjectJson()` is already pretty long, so I think I'll skip the `get` prefix unless you feel really strongly about it.

kevinvangelder:
@TheAlmightyBob I added to the Injection screen, however to really demonstrate this would require changes to the website being loaded, as the injectedJavaScriptBeforeContentLoaded does not consistently inject before content is loaded, and would result in inconsistent behavior of the demo.

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -387,20 +404,33 @@ public ThemedReactContext getThemedReactContext() {
...
   protected class RNCWebViewBridge {
+        private String TAG = "RNCWebViewBridge";
         RNCWebView mWebView;
+        String injectedObjectJson;
 
         RNCWebViewBridge(RNCWebView c) {
           mWebView = c;
         }
 
+        public void setInjectedObjectJsonJson(String s) {
```
TheAlmightyBob, line 415: One `Json` is enough 🙂

TheAlmightyBob:
> to really demonstrate this would require changes to the website being loaded, as the injectedJavaScriptBeforeContentLoaded does not consistently inject before content is loaded, and would result in inconsistent behavior of the demo

Huh, I saw the HTML defined at the top of the file... didn't realize that it's not actually being used.

Couldn't you put something in the normal `injectedJavaScript` script that checks for the `injectedObjectJson`? (i.e., don't you want the script to run _after_ load to verify that `injectedObjectJson` is available?)

If your concern is that such a demo wouldn't prove that the object is available before the content is loaded... I get that would be more complex, but think it's still useful to demonstrate how the feature _generally_ works and the object can be accessed.

example/examples/Injection.tsx
```diff
@@ -102,6 +102,11 @@ export default class Injection extends Component<Props, State> {
                 window.self.document.body.style.backgroundColor = "cyan";
               }
 
+              // Example usage of injectedJavaScriptObject({hello: 'world'})
+              const injectedObjectJson = window.ReactNativeWebView.injectedObjectJson();
+              const injectedObject = JSON.parse(injectedObjectJson);
+              console.log(injectedObject); // Should be: { hello: 'world' }
```
TheAlmightyBob, line 108: At the very least, I think this should include the text "injectedObject" or "injectedJavaScriptObject"

But I think it would be _ideal_ if you could actually put the content into the webview using the `appendChild` technique from further down in this script, like with `namedFramesAtAfterContentLoadedEle`?

kevinvangelder:
@TheAlmightyBob I just returned from two weeks off. I'll try to look at making those final changes today.

kevinvangelder:
@TheAlmightyBob Any final thoughts?

TheAlmightyBob:
@kevinvangelder I found an issue where the example broke if the object wasn't available, because `JSON.parse` throws an error if given `undefined` rather than just returning `undefined`. I fixed the example app, but perhaps should update the docs also?

I also made the change I suggested to actually display the object in the example app, just to bring more visibility to this feature!

Feel free to adjust if you think there's something wrong with these changes, I was just trying to make up for some of the delays 🙂

kevinvangelder:
@TheAlmightyBob Looks good to me.

TheAlmightyBob:
Updated the docs according to my last comment

TheAlmightyBob (approved):

TheAlmightyBob:
@kevinvangelder Thanks for your patience/persistence! 🙂

Titozzz:
@TheAlmightyBob I appreciate the follow up on the PR but we just introduced a new API to work around a platform bug instead of fixing the bug, making the webview API surface even bigger. We might want to document that this is a bandage, until the real issue gets solved, that way we can remove it later on. What do you think?

birdofpreyru:
My 5 cents (passing by to see what the new, still undocumented, prop does) &mdash; if this indeed solves the bug with `injectedJavaScriptBeforeContentLoaded`, instead of adding a new prop, probably this hack could be used just to alter and fix the internal implementation of `injectedJavaScriptBeforeContentLoaded` prop in the library, so that works correctly.

TheAlmightyBob:
@birdofpreyru 
re:
> passing by to see what the new, still undocumented, prop does

What documentation was missed? The Guide, Reference, and JSDocs were all updated as part of this PR.

@Titozzz 
I think that's fine... in my mind that was already covered by the updates to Guide & Reference docs. What else did you have in mind? Are you just thinking of a big "TODO" in code, or like a warning in the Reference doc that the feature might be removed in the future?

Back to @birdofpreyru (Sorry for the mixed order but wanted to address documentation questions first): I don't think that this can be used to fix `injectedJavaScriptBeforeContentLoaded`. This approach only works if the web page's existing Javascript knows to look for the injected object. You can't use it to just inject arbitrary Javascript to a page that knows nothing about React Native. Even if we augmented this to support passing functions instead of just data, it would have the limitation that the web page needs to know to look for and execute that function.

For that reason I don't even really see this as a real workaround since it's of limited capability, but since `injectedJavaScriptBeforeContentLoaded` is so unreliable on Android (I myself tried to use it for a project and had to abandon it), this solution seemed like it could at least unblock some use cases.

kevinvangelder:
@birdofpreyru It's in the API Reference, but I failed to add a link in the index section, so I've opened a follow-up PR to do so. https://github.com/react-native-webview/react-native-webview/pull/3139

PR #3011, opened 2023-06-19
feat(android): 🌟 add support for React Native 0.73

MasonLe2497:
# Summary
Starting from React Native v0.73 , all libraries will need to be updated with these two one-liners due to Android Gradle Plugin upgrade
[discussions-671](https://github.com/react-native-community/discussions-and-proposals/issues/671)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -44,6 +44,7 @@ if (isNewArchitectureEnabled()) {
 apply plugin: 'kotlin-android'
 
 android {
+    namespace = "com.reactnativecommunity.webview"
     compileSdkVersion getExtOrIntegerDefault('compileSdkVersion')
 
     defaultConfig {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -43,8 +43,31 @@ if (isNewArchitectureEnabled()) {
 }
 apply plugin: 'kotlin-android'
 
+def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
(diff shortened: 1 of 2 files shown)


android/build.gradle
```diff
@@ -43,7 +43,31 @@ if (isNewArchitectureEnabled()) {
...
+} else {
+    if(!manifestContent.contains("$PACKAGE_PROP")){
+        manifestContent = manifestContent.replace(
+            '<manifest',
+            "<manifest $PACKAGE_PROP "
+        )
+    }
+}
+manifestContent.replaceAll("  ", " ")
+manifestOutFile.write(manifestContent)
```
TheAlmightyBob, line 65: This is a lot more than the "two one-liners" mentioned in the description. Could you add a comment explaining the purpose of this? (I expect at some point we'll be able to remove it as we drop support for older React Native versions)

alexhernandez: This might be a leaner fix:
https://github.com/react-native-community/discussions-and-proposals/issues/671#issuecomment-1598892444

TheAlmightyBob: Indeed. I did read the thread though and understand that the point of the more complicated version is to reduce compiler warnings... which isn't a bad idea, if it can be done in an understandable way.

MasonLe2497: @TheAlmightyBob if user can accept the warning message, we can keep package in AndroidManifest and add namspace in build.gradle. Then we can keep going

PR #2937, opened 2023-04-24
Feature/webview debugging enabled prop

Tom-Bury:
## Changes
- Adds the `webviewDebuggingEnabled` prop to enable/disable webview debugging on Android & iOS 16.4+
- Clean up `Reference.md` linking

---

Closed https://github.com/react-native-webview/react-native-webview/pull/2896) to address new architecture & add the iOS functionality (issue https://github.com/react-native-webview/react-native-webview/issues/2914).

Change is similar to https://github.com/react-native-webview/react-native-webview/pull/2933, except that  it sits behing a prop flag.

Currently tested on a [fresh RN starter app](https://github.com/Tom-Bury/rn-example-app) in simulators. Will also try to test on an older version of RN on actual builds & on new architecture.

PR #3029, opened 2023-06-30
Update debugging instructions

TuurDutoit:
A new `webviewDebuggingEnabled` prop has been added in https://github.com/react-native-webview/react-native-webview/pull/2937. However, the debugging instructions were still mentioning the old ways:
- Android: `setWebContentsDebuggingEnabled`
- iOS: not working on iOS 16.4+

I updated the relevant parts of the instructions to mention the new prop.

### DIFF
--- a/docs/Debugging.md
+++ b/docs/Debugging.md
@@ -51,6 +51,7 @@ It's possible to debug WebView contents in the iOS simulator or on a device usin
 
 #### Steps:
 
+1. Mark the WebView as being "inspectable" using the [`webviewDebuggingEnabled`](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#webviewDebuggingEnabled) prop
 1. Open Safari Preferences -> "Advanced" tab -> enable checkbox "Show Develop menu in menu bar"
 2. Start app with React Native WebView in iOS simulator or iOS device
 3. Safari -> Develop -> [device name] -> [app name] -> [url - title]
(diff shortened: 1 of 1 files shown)


o-alexandrov (commented):
Steps for iOS are in the wrong order: 1 after 1.

docs/Debugging.md
```diff
@@ -51,6 +51,7 @@ It's possible to debug WebView contents in the iOS simulator or on a device usin
 
 #### Steps:
 
+1. Mark the WebView as being "inspectable" using the [`webviewDebuggingEnabled`](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#webviewDebuggingEnabled) prop
 1. Open Safari Preferences -> "Advanced" tab -> enable checkbox "Show Develop menu in menu bar"
```
o-alexandrov, line 55: We probably should fix the invalid order here

TuurDutoit: You mean the numbering? Or is something else wrong?

o-alexandrov: just the numbering

TheAlmightyBob: @jamonholmgren pointed out in https://github.com/react-native-webview/react-native-webview/pull/3081#discussion_r1282326026 that Markdown will still render the numbering correctly, so while it looks weird in the raw view, I guess it doesn't actually hurt the user-facing documentation

TheAlmightyBob (approved):

PR #3033, opened 2023-07-03
feat: add support for onUrlRejected prop

kennethlynne:
Make it possible to override how rejected urls are handled. 
Example use case: When a link is opened in a webview and it should open a native screen, be downloaded or silently rejected.


TheAlmightyBob (changes requested):
What's the advantage of this over using `onShouldStartLoadWithRequest` to do custom handling?

(also, please update `Reference.md`)

kennethlynne:
I actually think I can work around it using that prop yes! Closing, thanks for your feedback

Issue #3024, opened 2023-06-26
Kotlin version issue ,react native build is failing


ashirkhan94:
After upgrade gradle version from 7.3.3 to 7.5
we getting below issue
```
The Android Gradle plugin supports only kotlin-android-extensions Gradle plugin version 1.6.20 and higher.
The following dependencies do not satisfy the required version:
project ':react-native-webview' -> org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.0
```
The project `build.gradle` contain `kotlin_version = '1.6.20'`
but in `react-native-webview` contain `ReactNativeWebView_kotlinVersion=1.6.0`

<img width="1041" alt="Screenshot 2023-06-26 at 12 24 32 PM" src="https://github.com/react-native-webview/react-native-webview/assets/52521541/7b08f192-1f26-4c03-bf7b-38abe4044f80">

 - OS: macOS
 - OS version:13.2
 - react-native version: 66.3
 - react-native-webview version: 13.2.2

I saw some same issues but those are closed

Garguy:
Same issue here with React Native 71 and WebView 13.2.2

salimkt:
Same issue here also

mohzameer:
I forked and updated ReactNativeWebView_kotlinVersion to my project version.
https://github.com/react-native-webview/react-native-webview/blob/master/android/gradle.properties

I sent a PR https://github.com/react-native-webview/react-native-webview/pull/3035

esalaza:
I went with a [patch-package](https://www.npmjs.com/package/patch-package) in the meantime (for `12.4.0`):

```patch
diff --git a/node_modules/react-native-webview/android/build.gradle b/node_modules/react-native-webview/android/build.gradle
index b38e813..55f8031 100644
--- a/node_modules/react-native-webview/android/build.gradle
+++ b/node_modules/react-native-webview/android/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         gradlePluginPortal()
     }
     dependencies {
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${safeExtGet('kotlinVersion')}")
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.20"
         classpath("com.android.tools.build:gradle:7.0.4")
     }
 }
```

markholland:
[...]

PR #3035, opened 2023-07-04
Update gradle.properties

mohzameer:
Updated gradle version


omonk:
@Titozzz is it possible to merge this? from what I can see (and am experiencing) this is broken on latest versions of gradle https://github.com/react-native-webview/react-native-webview/issues/3024

Issue #3003, opened 2023-06-14
Javascript in the Webview is not executed on iOS with new RN architecture from 13.0.0


louiszawadzki:
**Bug description:**
JS is not executed inside Webviews on iOS from `react-native-webview` 13.0.0.
Anything in `<script>` tags or `onclick` callbacks are not executed.

Version `12.4.0` or using the old RN architecture works fine.

**To Reproduce:**
- Create new RN app
- Run `yarn add react-native-webview` 
- Run `(cd ios && RCT_NEW_ARCH_ENABLED=1 pod install)`
- Paste this component into your App.tsx:

```ts
import WebView from 'react-native-webview';
import React from 'react';
import {
  View,
} from 'react-native';

function App(): JSX.Element {
  return (
      <View
        style={{
          marginTop: 20,
          height: 560,
          width: 300,
        }}>
        <WebView
          source={{
            html: `
            <html lang="en">
  <head>
    <script>alert('Loading WebView')</script>
  </head>

  <body>
        <button type="button" onclick="alert('clicked button')">Trigger alert</button>
  </body>
</html>`,
          }}
        />
      </View>
  );
}

export default App;

```

**Expected behavior:**
- An alert showing "Loading Webview" shows up
- When clicking on the button, "clicked button" shows up

**Screenshots/Videos:**

No alert show up.

**Environment:**
 - OS: iOS
 - OS version: tested in 16.4, 16.3
 - react-native version: 0.71.8
 - react-native-webview version: 13.0.0

SvenNussgruber:
@louiszawadzki I ran into a similar issue on an older version (v11). Javascript events were just not being triggered anymore. It still worked on other dev's systems, so what fixed it in the end for me was recloning the repo 🤷‍♂️ hope this helps

agent-0028:
I can confirm, this is a regression in 13.x.

We have been working towards integrating the WebView into our app for a while, I tried to upgrade to the latest and seems JS is not being evaluated at all.

Going back to 12.4.0 is working properly in this regard.

billnbell:
Same for me - v13+ does not have Javascript while 12.4 does.

Who can look into this?

TheAlmightyBob:
Looked into and published #3041. Thank you all for the report! The change is simple enough that I encourage you to try patching the library in the meantime.

TheAlmightyBob:
[...]

Issue #3025, opened 2023-06-28
[iOS] Crash EXC_BAD_ACCESS (KERN_INVALID_ADDRESS) on webview unmounting


Minishlink:
**Bug description:**

Hello, we have these crash reports on Crashlytics after an upgrade from 11.23.1 to 13.2.1

```stacktrace
EXC_BAD_ACCESS (KERN_INVALID_ADDRESS) 0x000000000000001e
Crashed: com.apple.main-thread
0  libobjc.A.dylib                0x9af0 object_isClass + 16
1  Foundation                     0x8090 KVO_IS_RETAINING_ALL_OBSERVERS_OF_THIS_OBJECT_IF_IT_CRASHES_AN_OBSERVER_WAS_OVERRELEASED_OR_SMASHED + 72
2  Foundation                     0x3f9f8 NSKeyValueWillChangeWithPerThreadPendingNotifications + 296
3  WebKit                         0x430b6c WebKit::PageClientImplCocoa::gpuProcessDidExit() + 44
4  WebKit                         0x658748 WebKit::WebProcessPool::gpuProcessExited(int, WebKit::ProcessTerminationReason) + 488
5  WebKit                         0x6e4f54 WebKit::GPUProcessProxy::gpuProcessExited(WebKit::ProcessTerminationReason) + 216
6  WebKit                         0x6e50cc WebKit::GPUProcessProxy::processIsReadyToExit() + 136
7  WebKit                         0x1d6e98 WebKit::GPUProcessProxy::didReceiveMessage(IPC::Connection&, IPC::Decoder&) + 200
8  WebKit                         0x415f10 IPC::Connection::dispatchMessage(std::__1::unique_ptr<IPC::Decoder, std::__1::default_delete<IPC::Decoder> >) + 260
9  WebKit                         0x418c1c WTF::Detail::CallableWrapper<IPC::Connection::enqueueIncomingMessage(std::__1::unique_ptr<IPC::Decoder, std::__1::default_delete<IPC::Decoder> >)::$_13, void>::call() + 188
10 JavaScriptCore                 0x51b58 WTF::RunLoop::performWork() + 200
11 JavaScriptCore                 0x52890 WTF::RunLoop::performWork(void*) + 36
12 CoreFoundation                 0xd5f54 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 28
[...]

Issue #3020, opened 2023-06-25
Crash when having multiple webviews after version 13.0.0


ororsatti:
**Bug description:**
Theres a crash when webview on multiple tabs.
I suspect the issue got in on PR #2930 ,since using earlier versions works.

I am not sure what the solution is here, maybe mark it as unsafe that you have to pass a prop to get it ?
**To Reproduce:**
Here's a repo I created: [webview-memory-leak](https://github.com/ororsatti/webview-memory-leak)

**Screenshots/Videos:**

https://github.com/react-native-webview/react-native-webview/assets/109456266/42682df5-940f-42a4-92eb-541d8df90ae5

**Environment:**
 - OS: iOS
 - OS version: doesnt matter
 - react-native version: 0.72
 - react-native-webview version: 13.0.0

TheAlmightyBob:
Are you using something to detect an actual memory leak here? The repo you linked just mentions a crash. I've experienced an occasional crash with version 13 but I don't _think_ it's a leak... in my experience it happens "randomly," including sometimes on only the second WebView (or perhaps cleaning up the first? I'm still not certain of the cause)...

ororsatti:
> Are you using something to detect an actual memory leak here? The repo you linked just mentions a crash. I've experienced an occasional crash with version 13 but I don't _think_ it's a leak... in my experience it happens "randomly," including sometimes on only the second WebView (or perhaps cleaning up the first? I'm still not certain of the cause)...

Good point, I didn't use anything to identify that its a memory leak, the error logs and the stack trace of iOS led me to believe it was, but I am not sure for certain it is.

I'd be happy to help debugging this.
Also changing the title.

PR #3041, opened 2023-07-08
fix: Revert to didMoveToWindow #2930

TheAlmightyBob:
Fixes multiple issues with the iOS new architecture (WebViews failing to render, #3003, possibly more) and occasional crash with the iOS old architecture.

Re: the crashes on old architecture - I don't know the exact reason for this, but I saw the test app crash with EXC_BAD_ACCESS frequently before this change and am unable to repro the crash after the change.

Re: new architecture issues - I don't know the details, but the new architecture rarely calls `didMoveToSuperview`, so initialization code was simply never getting executed. It still calls `didMoveToWindow`. An example of the "WebViews failing to render" is that, if you run the test app, the WebView will initially display, but then if you switch to a different test, or tap "simulate restart," or refresh React Native via cmd+R/etc, the WebView will disappear and never come back until you kill and restart the app. (I repro'd this outside of the test app also)

I'm not opposed to the spirit of #2930 , but I think these issues are serious enough that it's best to revert for now and reevaluate the solution. (perhaps there's a "smart" way to use both methods...)

@Titozzz

### DIFF
--- a/apple/RNCWebViewImpl.m
+++ b/apple/RNCWebViewImpl.m
@@ -436,16 +436,9 @@ RCTAutoInsetsProtocol>
   return wkWebViewConfig;
 }
 
-// react-native-mac os does not support didMoveToSuperView https://github.com/microsoft/react-native-macos/blob/main/React/Base/RCTUIKit.h#L388
-#if !TARGET_OS_OSX
-- (void)didMoveToSuperview
-{
-  if (_webView == nil) {
-#else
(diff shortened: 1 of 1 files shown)


Titozzz:
Thanks for investigating, I'll take a look tomorrow ( and probably revert )

TheAlmightyBob:
Some other open issues that this should address: #3020, #3025

PR #3046, opened 2023-07-09
Reference documentation updates for menuItems

naveennazimudeen:
### What's in this PR
---
- Update reference documentation for `menuItems` and `onCustomMenuSelection` props, to reflect the changes made to it in Release [v12.3.0](https://github.com/react-native-webview/react-native-webview/releases/tag/v12.3.0). <br><br>The changes are described in #2610

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -1478,11 +1478,11 @@ Example:
 
 ### `menuItems`[⬆](#props-index)
 
-An array of custom menu item objects that will be appended to the UIMenu that appears when selecting text (will appear after 'Copy' and 'Share...').  Used in tandem with `onCustomMenuSelection`
+An array of custom menu item objects that will be shown when selecting text. An empty array will suppress the menu.  Used in tandem with `onCustomMenuSelection`
 
 | Type                                                               | Required | Platform |
 | ------------------------------------------------------------------ | -------- | -------- |
-| array of objects: {label: string, key: string}                     | No       | iOS      |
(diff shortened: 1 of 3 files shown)


TheAlmightyBob (changes requested):
Thanks for following-up on this. It looks like `WebViewTypes.ts` needs corresponding updates?

Also, I admit I hadn't closely followed the original MR... is it no longer possible to have both custom and default menu items? Why? (not a use case I have, I'm just curious)

Crash--:
@naveennazimudeen Thanks for this PR. @TheAlmightyBob I made a follow up https://github.com/react-native-webview/react-native-webview/pull/3066 with the updated types. (I can't answer for the why...)

naveennazimudeen:
@TheAlmightyBob Thanks for reviewing the PR. I have updated `WebViewTypes.ts`. 

The requirement was to show only the menu items passed replacing the default ones. 

To add on to default menu items, there is a way to do in iOS by changing to `YES` or calling the parent `canPerformAction` [here](https://github.com/react-native-webview/react-native-webview/blob/197593c62a200aa5b8f27be2c75a354dc45dba71/apple/RNCWebViewImpl.m#L65). I didn't see (or couldn't figure out yet) a way to do it in Android. 

Since default context menu can sometimes be a long list, as the requirement comes up, we will have to see if there is a need to keep some and not all the default menu items, and if there is a need to order them.

TheAlmightyBob (approved):

PR #3047, opened 2023-07-10
Update new architecture config

j-piasecki:
- when building on new architecture on RN 0.72 I've got a few build errors about some methods not being available on iOS 9.0, so I bumped the minimum iOS version on the new arch to 10.0
  - I've also replaced the manually listed dependencies with `install_modules_dependencies` (see https://reactnative.dev/docs/new-architecture-library-ios#add-folly-and-other-dependencies, introduced in 0.71).
- I've removed the `react` block from `build.gradle`, since paths are set by default in RNGP (see https://github.com/facebook/react-native/blob/main/packages/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/ReactExtension.kt), and the rest is also declared in `codegenConfig` in `package.json`

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -83,13 +83,3 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib:${safeExtGet('kotlinVersion')}"
     implementation "androidx.webkit:webkit:${safeExtGet('webkitVersion')}"
 }
-
-if (isNewArchitectureEnabled()) {
-    react {
-        jsRootDir = file("../src/")
-        libraryName = "rncwebview"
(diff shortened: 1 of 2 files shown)


WoLewicki (approved):
🚀

WoLewicki:
cc @Titozzz do we want to merge it?

Titozzz:
> any plan to merge that soon @Titozzz 

When I get back from holidays ✌️

Titozzz:
The iOS part has been updated by #3231 , will merge for the andorid part

WoLewicki:
Yeah, should be enough now 🚀

PR #3049, opened 2023-07-11
Update Guide.md to link a working static server package

qwertynik:
The static server package does not seem to work in the later version of React Native. Or, there's a way to make it work but not discussed.  A [fork](https://github.com/birdofpreyru/react-native-static-server) of the library works as expected.  While it may not have many 'stars' the fact that it works and is also maintained, makes it a more suitable candidate.


qwertynik:
Thanks @Titozzz 🙏. Much appreciated.

PR #3075, opened 2023-07-28
add inspectable property for iOS/macOS

kevinvangelder:
As far as I can tell this *should* work, but does not.

https://developer.apple.com/documentation/webkit/wkwebview/4111163-inspectable/
https://webkit.org/blog/13936/enabling-the-inspection-of-web-content-in-apps/


TheAlmightyBob:
What's the difference between this and [`webViewDebuggingEnabled`](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#webviewDebuggingEnabled)?

example/ios/.xcode.env
```diff
@@ -1 +1 @@
-export NODE_BINARY=/usr/local/bin/node
+export NODE_BINARY=/Users/kvg/.asdf/installs/nodejs/18.16.1/bin/node
```
o-alexandrov, line 1: You probably want to remove this leftover

o-alexandrov (changes requested):
Example is broken

kevinvangelder:
@TheAlmightyBob I looked for an existing implementation and could not find one, thanks for pointing it out.

PR #3082, opened 2023-08-01
Suppress menu items

Arjan-Zuidema:
This is a followup on https://github.com/react-native-webview/react-native-webview/pull/2845

Implemented the `suppressMenuItems` prop (iOS only) to suppress menu items from the long press context menu.

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -77,6 +77,7 @@ This document lays out the current public properties and methods for the React N
 - [`onFileDownload`](Reference.md#onFileDownload)
 - [`limitsNavigationsToAppBoundDomains`](Reference.md#limitsNavigationsToAppBoundDomains)
 - [`textInteractionEnabled`](Reference.md#textInteractionEnabled)
+- [`suppressMenuItems`](Reference.md#suppressMenuItems)
 - [`mediaCapturePermissionGrantType`](Reference.md#mediaCapturePermissionGrantType)
 - [`autoManageStatusBarEnabled`](Reference.md#autoManageStatusBarEnabled)
 - [`setSupportMultipleWindows`](Reference.md#setSupportMultipleWindows)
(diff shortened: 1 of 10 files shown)


TheAlmightyBob (changes requested):
Please update `Reference.md` (the previous PR did, but seems to have been missed here)

Also, has this been tested with the new architecture on iOS? I'm curious if anything needs to be added to `RNCWebView.mm` (it's possible it doesn't, I'm not sure)

Otherwise looks good

Arjan-Zuidema:
Hi @TheAlmightyBob I was on vacation for the past couple of weeks.

> Please update `Reference.md` (the previous PR did, but seems to have been missed here)
Yes will do this!

> 
> Also, has this been tested with the new architecture on iOS? I'm curious if anything needs to be added to `RNCWebView.mm` (it's possible it doesn't, I'm not sure)

I can't remember, will test it again at the end of the week. It seems the checks all pass on CI. Which should be sufficient?

TheAlmightyBob:
> It seems the checks all pass on CI. Which should be sufficient?

Unfortunately not, this project's CI is not quite that sophisticated. 😞

Arjan-Zuidema:
Seems like I did not test the new arch. I can't seem to `install pod` with `RCT_NEW_ARCH_ENABLED=1` Any idea how to do this? Can't seem to find anything in [Contributing.md](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Contributing.md) about this

image

Arjan-Zuidema:
I got a build running with fabric enabled:
- `cd example/ios`
- `RCT_NEW_ARCH_ENABLED=1 pod install`
- `cd ../..`
- `yarn ios`

But it seems the properties are not working with fabric enabled, but without fabric it's working just fine. How do I fix this?

TheAlmightyBob:
Glad you figured out the new architecture build and good point that the Contributing doc should include instructions for that!

I can't tell you how to fix things without debugging and knowing details of what's failing (also I'm not an expert on the new architecture), but as I suggested in my previous comment, `RNCWebView.mm` may need to be updated in some form. That file is specific to the new architecture.

Arjan-Zuidema:
> Glad you figured out the new architecture build and good point that the Contributing doc should include instructions for that!

I'll make a separate pull request to update the docs, next week.
 
> I can't tell you how to fix things without debugging and knowing details of what's failing (also I'm not an expert on the new architecture), but as I suggested in my previous comment, `RNCWebView.mm` may need to be updated in some form. That file is specific to the new architecture.

I added support for fabric, it's working now. Hope this is good, I also am not expert on fabric or objective-c for that matter 😛

TheAlmightyBob (approved):

docs/Reference.md
```diff
@@ -1420,6 +1421,30 @@ Example:
 
 ---
 
+### `suppressMenuItems`[⬆](#props-index)
+
+Allowes to suppress menu item from the default context menu.
```
TheAlmightyBob, line 1426: typo
```suggestion
Allows to suppress menu item from the default context menu.
```

TheAlmightyBob (changes requested):

TheAlmightyBob (approved):

PR #3105, opened 2023-08-17
Fix mistake in doc for Custom-iOS (RNCWebView -> RNCWebViewImpl)

fredikey:
There is a mistake in "Setting Client Certificate Authentication Credential" and "Allowing custom CAs"

import of "RNCWebView" doesn't work ("Use of undeclared identifier 'RNCWebView") and only "RNCWebViewImpl" works correctly

### DIFF
--- a/docs/Custom-iOS.italian.md
+++ b/docs/Custom-iOS.italian.md
@@ -137,7 +137,7 @@ Una volta esposti i metodi, puoi farvi riferimento nella tua classe di web view 
 Se si aprono pagine web che richiedono un certificato client per l'autenticazione, è possibile creare una credenziale e passarla alla webview:
 
 ```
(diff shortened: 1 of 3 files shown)


TheAlmightyBob (approved):
Thanks! A lot of related updates are still needed in this doc, but this is a start!

TheAlmightyBob:
Hmm... Circle CI seems unhappy.

Oh damn. It looks like this is likely because you're submitting from Russia. 😞 

CircleCI has enacted restrictions against this in accordance with U.S. sanctions: https://support.circleci.com/hc/en-us/articles/360043679453-CircleCI-Terms-of-Service-Violation-Sanctioned-Country

fredikey:
@TheAlmightyBob ohh, what should we do to fix this?

Issue #2949, opened 2023-05-05
Android: userAgent disables goBack to first url


jadfls:
**Bug description:**
On Android: When userAgent is set then goBack does not work anymore to the first url. It does work on subsequent pages but you can never go back to the very first.

**To Reproduce:**
```
<WebView
  ref={(ref) => (this.webview = ref)}
  userAgent={'custom-string'}
  source={{ uri: 'https://infinite.red'}}
/>
```

1. Create a button to trigger this.webview.goBack()
2. Now have the first webpage opened (https://infinite.red)
3. Click a link and try to go back => does not work
4. Click a link and another link and try to go back => does work once. but cannot go back to very first page

Note: if you omit userAgent or if you set userAgent to empty string '', then it works as expected.

**Expected behavior:**
With custom userAgent set, that you can navigate and goBack as without custom userAgent set. You can go back to the very first page as well

**Screenshots/Videos:**

**Environment:**
 - OS: Android
 - OS version: Android 12
 - react-native version: 0.66.x
 - react-native-webview version: 12.0.2

Yangnnnn:
same issue, looking for a solution

marcelo-soares-souza:
same issue here!

ooyo:
same issue here! works on emulator, but not on physical Android devices

Asad520:
I'm facing the same issue for `applicationNameForUserAgent` prop. Is any fix available?

alexanderblackh:
Can confirm this is an issue still.

sudesh-poojari:
Facing same issue with custom userAgent. Where history back doesn't work for first url.

Is there any work around for this.

sparkertime:
I am so relieved to find this Github issue. I thought I was going crazy!

I am seeing similar behavior to what @SteffenSr noted - there seems to be some interaction in our app between `applicationNameForUserAgent ` and `injectedJavascript` that causes this. If I remove either entirely, no problem.

However, initial testing seems to be that we can switch `applicationNameForUserAgent` to `userAgent` and the problem goes away. I need to do more testing to have much confidence in that though.

gronxb:
Hello. I solved the problem and made a pull request.

I need your attention.

asbaghel08:
I am not able to go back in webview after a location access pop appears. Although not using userAgent

Mookiies:
This issue also affects navigation done via `window.history.back()`. It is not limited to the webview's `goBack()` API.

[...]

PR #3106, opened 2023-08-17
fix(android): userAgent disables goBack to first url

gronxb:
https://github.com/react-native-webview/react-native-webview/issues/2949

When set with `injectedJavaScript` and `userAgent`, `history.back()` or `history.go(-1)` does not work in webview.

It seems that there is a conflict as it is set up at the same time.
I kept the user agent and set the user agent when the page started.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewClient.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewClient.java
@@ -81,6 +81,8 @@ public class RNCWebViewClient extends WebViewClient {
       mLastLoadFailed = false;
 
       RNCWebView reactWebView = (RNCWebView) webView;
+      
+      reactWebView.getSettings().setUserAgentString(reactWebView.getUserAgentString());
       reactWebView.callInjectedJavaScriptBeforeContentLoaded();
     }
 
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewClient.java
```diff
@@ -81,6 +81,8 @@ public void onPageStarted(WebView webView, String url, Bitmap favicon) {
       mLastLoadFailed = false;
 
       RNCWebView reactWebView = (RNCWebView) webView;
+      
+      reactWebView.getSettings().setUserAgentString(reactWebView.getUserAgentString());
       reactWebView.callInjectedJavaScriptBeforeContentLoaded();
```
TheAlmightyBob, line 86: Considering that `injectedJavaScriptBeforeContentLoaded` does not work reliably on Android, I'm a bit nervous that this would make setting the user agent not work reliably anymore either...

gronxb: Isn't the reason why `injectedJavaScriptBeforeContentLoaded` didn't work reliably because of `evaluationJavaScript()`?

Is `onPageStarted` also relevant to this issue?

TheAlmightyBob: I don't think the reason has ever been fully understood, but there was at one point a theory that it was related to the lifecycle events: https://github.com/react-native-webview/react-native-webview/pull/1099#issuecomment-645678919

gronxb: After some research, we come up with a different solution.

https://github.com/react-native-webview/react-native-webview/pull/3133

PR #1465, opened 2020-06-16
Incognito bugfix, clearCache on iOS and clear cookies on both platforms

pdandradeb:
# Summary

Problems related to cache and cookies are common and these commits aim to improve the code parity between the platforms and add a easy way to remove cookies from the Webview;

- Fixes an incognito bug in Android that clears all data when incognito={false} as was reported in a few issues (Closes #1143, Closes #1296, Suggest a new approach for #1447)
- Adds iOS support for `clearCache` (which was previously supported only in Android);
- Creates a method `clearCookies` working in both iOS and Android;

### What's required for testing (prerequisites)?

1. Pull the changes;
2. Reinstall Pods. Easy way: cd example/ios && rm -rf Pods && pod install
3. Rebuild JS. Easy way: rm -rf lib; yarn build

### What are the steps to reproduce (after prerequisites)?

1. Change the uri to a page that use cookies and reload to make sure cookies are being used.
2. Change the `incognito` property to `true`; reload;
3. Verify that cookies are not being used;
4. Change the `incognito` property back to `false`; reload;
5. Verify on iOS: cookies are being used again; on Android: cookies were clean but can be used again;
6. Click on the 'Clear Cookies' button and verify that cookies were cleared;
7. Click on the 'Clear Cache' button and verify that data was cleared;

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |      ✅     |
| Android |      ✅     |

## Checklist

- [X] I have tested this on a device and a simulator
- [X] I added the documentation in `Reference.md` and `Guide.md`
- [ ] I mentioned this change in `CHANGELOG.md`
- [X] I updated the typed files (TS and Flow)
- [X] I added a sample use of the API in the example project (`example/App.js`)

PR #1270, opened 2020-03-25
Add clearCache for iOS [breaking change]

Minishlink:
# Summary

This adds `webviewInstance.clearCache()` for the iOS platform. This calls `WKWebsiteDataStore removeDataOfTypes:WKWebsiteDataStore.allWebsiteDataTypes`.

I didn't implement the "includeDiskData" boolean parameter because I'm not familiar with this and had not the need for it.

I noticed that this clears cache for all webviews, for android & ios. Maybe we should move this command at the native module level rather than the view level, since it's not really tied to the view... This way we could (at least for ios) return a promise. EDIT: I implemented a promise anyway actually

### What's required for testing (prerequisites)?

- iOS device or simulator
- app with `<WebView />`
- website that has a cache policy

### What are the steps to reproduce (after prerequisites)?

- mount the component
- call webviewRef.pauseLoading() (I do this in order to not have the wrong page cached page loaded completely)
- call await webviewRef.clearCache()
- call webviewRef.reload()

You can try with https://refreshyourcache.com/en/cache-test/ (although it's random between green and red so you might have to try again and again..)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] I have tested this on a device and a simulator
- [x] I added the documentation in `README.md`
- [ ] I mentioned this change in `CHANGELOG.md` => there's no CHANGELOG.md anymore?
- [x] I updated the typed files (TS and Flow)
- [ ] I added a sample use of the API in the example project (`example/App.js`)

PR #3119, opened 2023-08-28
Add clearCache method on iOS

robinheinze:
# Overview

Issues arise where it's necessary to clear web caches and storage. This makes the API complete by adding the ability to clear the cache on iOS (which was previously only supported on Android).

These changes are very similar to #1270 and #1465 (both closed no merge) but updated to account for the recent new arch and codebase structure changes.

## Visuals

Below you can see a demonstration using the example app. 

In this first video, you can see the font assets being loaded from memory in the simple reload, then when cache is cleared, you can see the same assets being loaded over the wire. 

https://share.cleanshot.com/1xBTWLf8

In the second video, you can see I've added an entry in local storage. When I perform the clearCache action, you can see that entry is removed. 

https://share.cleanshot.com/QDk1sXSk 

## Testing Info

- Clone the repo
- Run `yarn` and `cd example/ios && pod install`
- Run `yarn ios`
- From the example screen choose "Clear Data"
- Open Safari developer console for the that webview instance
- View the disk storage and cache and try out the clearCache button

### DIFF
--- a/example/examples/ClearData.tsx
+++ b/example/examples/ClearData.tsx
@@ -1,0 +1,55 @@
+import React, { Component } from 'react';
+import { View, Button } from 'react-native';
+
+import WebView from 'react-native-webview';
+
+type Props = {};
+type State = {};
+
+export default class ClearData extends Component<Props, State> {
+  state = {};
+
+  constructor(props) {
+    super(props);
+    this.webView = React.createRef();
+  }
+
+  // clearCookiesAndReload = () => {
+  //   this.webView.current.clearCookies();
+  //   this.webView.current.reload();
+  // }
+
+  clearCacheAndReload = (includeDiskFiles: boolean) => {
+    this.webView.current.clearCache(includeDiskFiles);
+    this.webView.current.reload();
+  };
+
+  reload = () => {
+    this.webView.current.reload();
+  };
+
+  render() {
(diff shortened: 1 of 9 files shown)


example/examples/ClearData.tsx
```diff
@@ -0,0 +1,55 @@
...
+  constructor(props) {
+    super(props);
+    this.webView = React.createRef();
+  }
+
+  // clearCookiesAndReload = () => {
+  //   this.webView.current.clearCookies();
+  //   this.webView.current.reload();
+  // }
+
```
jamonholmgren, line 21:
```suggestion
```

Let's remove this commented out code.

example/examples/ClearData.tsx
```diff
@@ -0,0 +1,55 @@
...
+    this.webView.current.reload();
+  };
+
+  render() {
+    return (
+      <View style={{ height: 1000 }}>
+        {/* <Button
+          title='Clear cookies'
+          onPress={this.clearCookiesAndReload}
+        /> */}
```
jamonholmgren, line 37:
```suggestion
```

Remove.

jamonholmgren (approved):
Approved with two nits; @Titozzz I think this is good to ship once those are removed.

PR #3133, opened 2023-09-09
fix(ios): Set User-Agent in the headers

gronxb:
Related https://github.com/react-native-webview/react-native-webview/issues/2949

hello. I wrote a PR (https://github.com/react-native-webview/react-native-webview/pull/3106) about [History API](https://developer.mozilla.org/en-US/docs/Web/API/History) not working properly when setting userAgent on Android,

The userAgent value, which is set with the `@ReactProp` annotation, is lazy load. 

So I'm predicting that the userAgent is set during loading, and on reload, the history is broken.

https://developer.android.com/reference/android/webkit/WebSettings#setUserAgentString(java.lang.String)

I found the fact that it reloads in the above link.

**Therefore, we decided that it would be very difficult to set the userAgent before loading the webpage because `@ReactProp` sets the userAgent and source at the same time.**

And I found that when I set user-agent in headers in the source prop, it also sets userAgent.

https://github.com/react-native-webview/react-native-webview/blob/10105bad59c9d9177c371a8d288f7a3b0faf517b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt#L412-L437

In this case, userAgent works fine and history works fine because we are setting userAgent before loading.

**So we need to make it work the same on IOS, to ensure hybridization.**

**I think it would be a good idea to deprecate the existing userAgent prop and recommend setting the userAgent in the headers.**

## Example Code
```tsx
import React, { Component } from 'react';
import { View } from 'react-native';

interface Props {}
interface State {}

export default class NativeWebpage extends Component<Props, State> {
  state = {};

  render() {
    return (
      <View style={{ height: 400 }}>
        <WebView
          source={{
            uri: 'https://infinite.red',
            headers: {
              // This Field !
              'User-Agent': 'Test Header',
            },
          }}
          style={{ width: '100%', height: '100%' }}
          injectedJavaScript={`
              const div = document.createElement('div');
              div.innerHTML = navigator.userAgent;

              document.body.appendChild(div);
              true;

            `}
         />
      </View>
    );
  }
}

```


gronxb:
@Titozzz @jamonholmgren 

[Issues](https://github.com/react-native-webview/react-native-webview/issues?q=is%3Aissue+is%3Aopen+user-agent)

I still believe this feature is necessary. We must acknowledge how messy the code is to support user-agent smoothly on both Android and iOS!

**Current Implementation (as-is):**
```jsx
<WebView
  source={{
    uri: 'https://infinite.red',
    ...(Platform.OS === 'android' && {
      headers: {
        'User-Agent': 'my-custom-user-agent',
      },
    })
  }}
  style={{ width: '100%', height: '100%' }}
  userAgent={Platform.OS === 'ios' ? 'my-custom-user-agent' : undefined}
  injectedJavaScript={`
      const div = document.createElement('div');
      div.innerHTML = navigator.userAgent;
      document.body.appendChild(div);
      true;
  `}
/>
```

**Proposed Implementation (to-be):**
```jsx
<WebView
  source={{
    uri: 'https://infinite.red',
    headers: {
      'User-Agent': 'my-custom-user-agent',
    }
  }}
  style={{ width: '100%', height: '100%' }}
  injectedJavaScript={`
      const div = document.createElement('div');
      div.innerHTML = navigator.userAgent;
      document.body.appendChild(div);
      true;
  `}
/>
````

gronxb:
@Titozzz @jamonholmgren 🙏

Titozzz:
Will do a global pass on PR today

gronxb:
@Titozzz Thank you. Is the PR actually closed?

Titozzz:
Sorry the bot closed it so I missed it

gronxb:
@Titozzz Please remove the stale tag.

krmao:
does this commit can instead of `injectedJavaScriptBeforeContentLoaded  Warning On Android, this may work, but it is not 100% reliable`, such as 'statusbarheight' in user agent

tomimarkus991:
need this 🙏🏼

Aikenwu:
hello @gronxb,

I'm still facing the same issue: when accessing page B from page A using the same WebView, regardless of whether I use `webview goback` or `history.back`, returning from page B to page A causes page A to reload.

This is consistent with the description; this issue doesn't occur on iOS, but it does on Android.

I've also looked at the pull request (https://github.com/react-native-webview/react-native-webview/pull/3133), but I still don't understand how to fix this problem.
image

gronxb:
@Aikenwu Remove `userAgent` or `applicationNameForUserAgent` props.  

And Migrate set  source.headers userAgent

Aikenwu:
> @Aikenwu Remove `userAgent` or `applicationNameForUserAgent` props.移除 `userAgent` 或 `applicationNameForUserAgent` props。
> 
> And Migrate set source.headers userAgent set source.headers userAgent

Could you provide me with sample code? Do I need the patch webview package? Thank you.

PR #3157, opened 2023-09-23
feat(ios): ios injectJavaScriptObject

gronxb:
Inspired by https://github.com/react-native-webview/react-native-webview/pull/2960.
As you may know, see the following usage:
```js
<WebView
    injectJavaScriptObject={{authToken: 'myValue'}}
/>
```
The code above can be accessed like the following way:
```js
<script>
    window.onload = (event) => {
        JSON.parse(injectedObject.toString()).authToken;
    }
</script>
```
The problem here is that we had to check whether injectedObject exists or not because most of webview contents use the same page sources not regarding the platform user uses. I have implemented it for iOS so that the `injectJavaScriptObject` can also be used in iOS.

The *working* screenshot of this implementation is as follows:
스크린샷 2023-09-23 오후 10 40 36


Titozzz (approved):

TheAlmightyBob:
> @TheAlmightyBob If you are not busy, please review the code.

It turned out I **was** busy. Sorry about that.

Thanks @Titozzz for taking over!

PR #3170, opened 2023-10-03
Update escape-string-regexp to 4.x

stianjensen:
This is the last version of escape-string-regexp before it's rewritten to ES modules.

Updating past 3.x requires at least node 10, but that is already a very old node version.

https://github.com/sindresorhus/escape-string-regexp/releases

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "react-native": "*"
   },
   "dependencies": {
-    "escape-string-regexp": "2.0.0",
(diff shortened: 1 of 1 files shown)


stianjensen:
Still relevant!

stianjensen:
Still relevant!

stianjensen:
Not stale

stianjensen:
Same status as last time.

Issue #3194, opened 2023-10-20
[iOS] missing cookies on initial requests


Skizu:
**Bug description:**

`webView` is `nil` when calling `-[RNCWebViewImpl writeCookiesToWebView:completion:]` in `-[RNCWebViewImpl resetupScripts:]`. 

**To Reproduce:**

Have a WebView that makes a request with an auth cookie from ReactNative.

**Expected behavior:**

Request headers include cookies.

**Screenshots/Videos:**

N/A

**Environment:**
 - OS: iOS
 - OS version: 11+
 - react-native version: v0.71+
 - react-native-webview version: v12.0.0+

MikhailMalinin-iOSDeveloper:
Same issue.
Your solution is not working

Asad520:
Facing a similar issue. I'm trying to set cookies using injectJavascript`document.cookie = myJWT`. However, the webview is not able to read the cookie when the next request is sent and refreshes the token which messes up some scenarios. I tried the code in your PR but the same results.

federico-agustin-diaz:
> Facing a similar issue. I'm trying to set cookies using injectJavascript`document.cookie = myJWT`. However, the webview is not able to read the cookie when the next request is sent and refreshes the token which messes up some scenarios. I tried the code in your PR but the same results.

Were you able to solve this issue?

one-god:
Same issue.

PR #3195, opened 2023-10-20
fix(iOS): nil webView during initial resetupScripts

Skizu:
`webView` is `nil` when calling `-[RNCWebViewImpl writeCookiesToWebView:completion:]` in `-[RNCWebViewImpl resetupScripts:]`. 

Initial fix for #3194.


Skizu:
New to the code base, as such only made a minor change for the specific area of fault.

Issue #3014, opened 2023-06-20
Scroll position of ScrollView jumps to WebView when pressed


vigneshrajs:
**Bug description:**
When pressing a WebView inside a scrollable ScrollView, the scroll position jumps down to the WebView.

**To Reproduce:**
 - Install React Native version 0.71.3 and React Native webview Version 11.25.0
 - Place webview inside ScrollView.
 
The first time you press, it will jump down slightly.

**Expected behavior:**
When clicking a WebView inside a ScrollView, it should not change the scroll position of the parent ScrollView.

**Environment:**
 - OS: Android
 - OS version: >11
 - react-native version: 0.71.3
 - react-native-webview version: 11.25.0

knakandakare:
Same issue here

sanduluca:
I have same issue. 
It was also reported here: #2360 #2057 
I also reported this issue in another repo ( here: https://github.com/iou90/react-native-autoheight-webview/issues/259 ) but it seems to be a problem with WebView



**Bug description:**
Content jumps to top on first time press/click

**To Reproduce:**

**To test on real device**
1. Open the screen
2. Press the webview content

**Source (static HTML or url):**
https://snack.expo.dev/VhrGR3bC5

1. Scan qr code and open app on a real android device
2. Press any area below first white section
To reproduce start from 1

**Expected behavior:**
No jump when press webview content

**Screenshots/Videos:**

https://github.com/iou90/react-native-autoheight-webview/assets/56228534/6155f52b-e469-4eaf-a105-fe863665b3b5


**Environment:**
 - OS: Android
 - OS version: 12
 - react-native version: 0.71.7
 - react-native-webview version: ^12.0.2
 - react-native-autoheight-webview version: ^1.6.5

TheAlmightyBob:
It looks like the auto-scrolling is a behavior of the React Native ScrollView.

The only thing I imagine could be done from the `react-native-webview` side would be to set `focusable` to `false`... but I think that only makes sense if your WebView content is completely noninteractive? If that _is_ your scenario, you could try hacking/patching in that change to see if it helps.

sanduluca:
> It looks like the auto-scrolling is a behavior of the React Native ScrollView.
> 
> The only thing I imagine could be done from the `react-native-webview` side would be to set `focusable` to `false`... but I think that only makes sense if your WebView content is completely noninteractive? If that _is_ your scenario, you could try hacking/patching in that change to see if it helps.

[...]

Issue #2360, opened 2022-02-02
Android: Scroll position of ScrollView jumps to WebView when pressed


drager:
Bug description:
When pressing a WebView inside a scrollable ScrollView, the scroll position jumps down to the WebView.

To Reproduce:

1. Install Expo
2. Clone repo: https://github.com/mtnptrsn/scrollview-webview-bug
3. Run yarn start in root
4. Scan QR code with expo
5. Before scrolling, click on the WebView

The first time you press, it will jump down slightly.

Expected behavior:
When clicking a WebView inside a ScrollView, it should not change the scroll position of the parent ScrollView.

Screenshots/Videos:

https://user-images.githubusercontent.com/684179/122516743-6d31d000-d00f-11eb-9814-7574d23c9bc2.mp4

Environment:

    OS: Android
    OS version: 11
    react-native version: 0.64.3
    react-native-webview version: 11.14.1

Duplicate off: https://github.com/react-native-webview/react-native-webview/issues/2057

Issue #2057, opened 2021-06-18
Scroll position of ScrollView jumps to WebView when pressed (Android)


mtnptrsn:
**Bug description:**
When pressing a WebView inside a scrollable ScrollView, the scroll position jumps down to the WebView.

**To Reproduce:**
1. Install Expo
2. Clone repo: https://github.com/mtnptrsn/scrollview-webview-bug
3. Run `yarn start` in root
4. Scan QR code with expo
5. Before scrolling, click on the WebView

The first time you press, it will jump down slightly.

**Expected behavior:**
When clicking a WebView inside a ScrollView, it should not change the scroll position of the parent ScrollView.

**Screenshots/Videos:**
https://user-images.githubusercontent.com/684179/122516743-6d31d000-d00f-11eb-9814-7574d23c9bc2.mp4

**Environment:**
 - OS: Android
 - OS version: 11
 - react-native version: 41.0.0
 - react-native-webview version: 11.2.3

bashen1:
same issue

drager:
Yeah, can confirm that this happens for me as well. Even in the latest release of react-native-webview (11.14.0).

amirkhanzada-iinv:
Same issue

steffenkleinle:
Experiencing the same issue, did anyone find a solution? Still present in latest version (12.0.2).

PR #3201, opened 2023-10-27
fix(android): Jumps with scrollview when pressed

VidocqH:
Jumps to `WebView` top when first time press inside `WebView` on Android.

Related issues: #2360 #2057 #3014 

Before:

https://github.com/react-native-webview/react-native-webview/assets/16725418/d52e3f45-e1a4-4121-80b8-1ad1166b0e7e

After:

https://github.com/react-native-webview/react-native-webview/assets/16725418/e7c5786e-7460-4ffa-a352-96bbc83b376a

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
@@ -101,6 +101,11 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
     }
 
     @Override
+    public boolean requestFocus(int direction, Rect previouslyFocusedRect) {
+      return true;
+    }
+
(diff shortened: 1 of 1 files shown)


TheAlmightyBob:
How does this behave with a web page that has interactive elements like text fields?

VidocqH:
You're right, it's a problematic solution. With <input> of type text, it cannot open keyboard as expected. I tested and confirm this solution can work with <button>, <select>, and even <input> of type checkbox. It's not a good enough solution but for my case that guarantee does not contain <input> seems good though.

TheAlmightyBob:
@VidocqH The effort is still appreciated, perhaps it will inspire a more robust solution in the future 🙂

PR #3219, opened 2023-11-10
fix(android): add buildFeatures.buildConfig true for AGP8+ compat

mikehardy:
## Summary

Upcoming react-native 0.73+ includes android gradle plugin 8+ which needs namespace in build.gradle but also needs buildFeatures.buildConfig enabled as well for modules that use it

It is not sufficient to enable this at the top-level app build.gradle, specific modules that use it (such as those that implement new architecture, it seems, or just directly use buildconfig like this one) must also enable it at the module level

This change was necessary and is in-use in my work app via patch-package as I work through android-gradle-plugin 8+ issues in prep for react-native 0.73 launching for everyone

This change is backwards compatible with android gradle plugin < 7, because it is in the same conditional as the namespace change which has similar backwards compatibility requirements

It is similar to changes I needed to do as react-native-firebase maintainer --> https://github.com/invertase/react-native-firebase/commit/b52d0ce6723c077190618641ce0f33ced9fd4090

With apologies, you have to alter an app that integrates this module to use android gradle plugin 8, it's difficult to do that in repos I'm proposing these changes in because bumping to android gradle plugin 8 requries a large amount of transitive dependency changes in CI

I have integrated this in an app and tested it, and done similar work as maintainer of react-native-firebase, react-native-netinfo and react-native-device-info, and I'm now pushing these out to the repos

run the build for android

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -58,6 +58,9 @@ apply plugin: 'kotlin-android'
 android {
     if (supportsNamespace()) {
         namespace "com.reactnativecommunity.webview"
+        buildFeatures {
+            buildConfig true
+        }
 
         sourceSets {
             main {


UNIDY2002:
Hmm, I've got one question: won't [this commit in react-native](https://github.com/facebook/react-native/commit/db45da71018c9e6e93f6049116d87df7937b9cda) automatically set `buildFeatures.buildConfig = true` for libraries?

mikehardy:
@UNIDY2002 that's a good point! I wasn't aware of that commit. Interesting - then it seems this PR would only be needed for the case where people were still on RN72 but testing AGP8+ (as I am). That is likely a very small audience, so despite being technically correct and *theoretically* needed for a small audience, it may not be worth it

mikehardy:
One immediate thought after re-reading that PR - it's nice the react-gradle handles that for modules, but there is a beauty to knowing the things a module needs are handled by the module

Stated differently - if a module knows it needs buildConfig, seems more coherent for the module to actually take care of it locally in it's own build.gradle. But that's a style / design argument and those are qualitative. Adults can and should disagree about those things, either way does work

Titozzz:
IIRC after discussing with Nico we should be able to remove 
```
        buildConfigField "boolean", "IS_NEW_ARCHITECTURE_ENABLED", isNewArchitectureEnabled().toString()
```
and remove the need for buildConfig altogether
WDYT @mikehardy ?

mikehardy:
Possibly? I recall a commit that is going in to the react-native gradle plugin 0.74 that handles build config variable for modules, if the only reason it is needed is for new architecture then that handles it, for rn74+

Possibly worth mentioning that it is possible for people on rn73 to use agp8 so it won't handle them, and if they need api34 compile target like I do then they'll be affected until rn74 releases and they upgrade rn

Stated differently, I require this patch now at least

UNIDY2002 (approved):

PR #3226, opened 2023-11-16
fix(Android): add capture attribute support for video

ohayou12:
```html
<!-- fine -->
<input name="video-only" type="file" accept="video/*">

<!-- not work -->
<input name="video-only" type="file" accept="video/*" capture="user">
```

fixed it.


Deibi12:
any update

Issue #3189, opened 2023-10-18
[Windows] build error: This project references NuGet package(s) that are missing on this computer.


panthuncia:
**Bug description:**

Attempting to build & run this project in Windows with 'npx react-native run-windows', I get the error:

This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is C:\Users\matth\Sailbot\sailbot_telemetry_reactnative\windows\packages\Microsoft.UI.Xaml.2.8.0\build\native\Microsoft.UI.Xaml.targets.. Check your build configuration.

The console log has:

√ Restoring NuGet packages 
 √ Auto-linking...
in it, so I'm not quite sure what's wrong.

The folder "Microsoft.UI.XAML.2.8.0" is indeed missing in the packages folder- instead, there is "Microsoft.UI.Xaml.2.8.0-prerelease.210927001". It looks like NuGet is getting the wrong version? 

If I rename that folder to Microsoft.UI.XAML.2.8.0, I get a new problem- command ""C:\Users\matth\.nuget\packages\microsoft.windows.cppwinrt\2.0.211028.7\build\native\..\..\bin\"cppwinrt @"x64\Debug\SailbotTelemetry.vcxproj.cppwinrt_ref.rsp"" exited with code 1.

This also happens when building in visual studio.

On android, it runs successfully.
**To Reproduce:**

https://github.com/panthuncia/sailbot_telemetry_reactnative

or:

just attempt to set up the example project from react-native-webview

install npm, windows extension, and try to run the project on Windows

**Expected behavior:**

The project should build successfully, or indicate how to resolve the issue

**Screenshots/Videos:**

**Environment:**
 - OS: Windows 11
 - OS version: 22621.2428
 - react-native version: 0.72.6
 - react-native-webview version: 13.6.2
 - react-native-windows version: 0.72.14

panthuncia:
Update: I got it to work by opening the visual studio solution and re targeting the NuGet packages to the versions that the error said were missing. Perhaps the defaults should be updated?

vinaykumardetrix:
can you describe more about it, which helps me out in a detailed way to follow.

rain9155:
[...]

Issue #1861, opened 2021-02-03
React-Native-Windows not build


maribeiroleya:
I am creating a new react-native app. I only install the react-native, react-native-windows and react-native-webview. When i run the command npx react-native run-windows i receive the error:

Error This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is C:\....\windows\packages\Microsoft.Windows.CppWinRT.2.0.200729.8\build\native\Microsoft.Windows.CppWinRT.props.    ReactNativeWebView    C:\...\node_modules\react-native-webview\windows\WebViewBridgeComponent\WebViewBridgeComponent.vcxproj 

I need the solution for this problem, please.

Environment:
 - OS: windows 10
 - OS version: 19042.746
 - react-native version:0.63.4
 - react-native-webview version:11.2.1

true-hamid:
We are facing the same issue as well

aalbaharnah:
I was able to build my app after changing the solution platform in visual studio from `x64` to `x86`

** Environment**
 - OS: windows 10 pro
 - OS version: 18363.1316
 - react-native version: 0.63.4
 - react-native-webview version: ^11.2.3
 - react-native-windows: ^0.63.0-0
 - react-native-windows language: cs

dd-apt:
I was only able to get this working by adding the WebViewBridgeComponent project to the main Windows app solution. At that point, the Nuget package required will be restored at build.

Note: even if you remove the project, locally it will still build after this (since nuget package is downloaded). However, it will not build on a different machine / CI build.

harrisrobin:
Also facing this issue on windows:

![Error-message_mmr-huv](https://user-images.githubusercontent.com/3880008/108807346-0583e080-7559-11eb-832b-7b8b47e29f58.png)

This can be reproduced by simply trying to build the project with `11.2.3`. 
here's an example repo here:
https://github.com/harrisrobin/windows-react-native-webview-repro

maribeiroleya:
The solution of @dd-apt works! @harrisrobin  if you are debugging you have to build x86.

harrisrobin:
I can confirm @dd-apt 's solution works.

VMBindraban:
This seems to be still the issue after 2 years with version 13.6.3. Can we document this somewhere?

PR #3242, opened 2023-11-29
Fix NuGet package reference

VMBindraban:
Fixes https://github.com/react-native-webview/react-native-webview/issues/3189 and https://github.com/react-native-webview/react-native-webview/issues/1861


VMBindraban:
@jamonholmgren Can you review this?

Issue #3289, opened 2024-01-15
Android: `onError()` implementation is deficient


birdofpreyru:
**Bug description:**

The `onError()` implementation for Android (see below), in its native layer, overrides [the version of `onReceivedError()` method deprecated in Android API 23](https://developer.android.com/reference/android/webkit/WebViewClient#onReceivedError(android.webkit.WebView,%20int,%20java.lang.String,%20java.lang.String)) (thus, since 2015 &mdash; this library is really well-maintained :rofl: )
https://github.com/react-native-webview/react-native-webview/blob/6960a1903c3c32e1e77be11e6370f1c384ed7ee4/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewClient.java#L211-L216

and unlike [the new version of the method](https://developer.android.com/reference/android/webkit/WebViewClient#onReceivedError(android.webkit.WebView,%20android.webkit.WebResourceRequest,%20android.webkit.WebResourceError)) the old one is fired for the main resource errors only &mdash; you get the error only if it failed to load the URI passed into the `source` itself, if it failed to load assets it depends upon (like JS scripts, CSS sheets, and other stuff required by the main resource) you won't get any error, even if it breaks the page and deserves an attempt to reload the page, or some other error-handling in your host app.

**P.S.:** As this is one is a sensitive problem for myself, I'll fix it and create PR shortly. If anybody is willing to say thanks for that, I am accepting donations via [GitHub.Sponsors](https://github.com/sponsors/birdofpreyru) :wink: 

**To Reproduce:**

**Expected behavior:**

**Screenshots/Videos:**

**Environment:**
 - OS:
 - OS version:
 - react-native version:
 - react-native-webview version:

birdofpreyru:
Hmm... having looked around a bit more, I see that `onHttpError` implementation actually actively ignores errors for assets loading... which is kind of inconsistent with what I want to do with `onError` :(
https://github.com/react-native-webview/react-native-webview/blob/6960a1903c3c32e1e77be11e6370f1c384ed7ee4/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewClient.java#L247-L263

Another inconsistency is with the current logic assuming that `onLoadEnd` should be fired prior to `onError`; while if I do fire `onError` for each asset may fail to load for the main resource, it makes sense to emit `onError` for each failing asset, and then `onLoadEnd` after it.

birdofpreyru:
I guess, my suggestion is to:
[...]

PR #3290, opened 2024-01-16
[#3289] Android: Trigger `onError` callback on asset loading errors

birdofpreyru:
Here is my PR for #3289:
- Avoids using [the deprecated version of `onReceivedError`](https://developer.android.com/reference/android/webkit/WebViewClient#onReceivedError(android.webkit.WebView,%20int,%20java.lang.String,%20java.lang.String)) for `onError` implementation on Android.
- Fires `onError` callback, if provided, for asset loading errors, in addition to the primary source loading errors.
- Does not alter the logic of other related callbacks (`onLoad`, `onLoadEnd`, _etc._); _i.e._ `onError` signals triggered for page assets do not prevent `onLoad` and `onLoadEnd` to be fired for the page itself, if that was loaded &mdash; thus, the page is signaled to have been loaded successfully, even if its assets failed to load, and that is just the original behavior of those signals before my PR.

Can we merge it, please? :)

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewClient.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewClient.java
@@ -3,12 +3,14 @@ package com.reactnativecommunity.webview;
 import android.annotation.TargetApi;
 import android.graphics.Bitmap;
 import android.net.http.SslError;
+import android.net.Uri;
 import android.os.Build;
 import android.os.SystemClock;
 import android.util.Log;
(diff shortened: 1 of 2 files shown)


Titozzz (approved):

Titozzz:
Note to self: Maybe release as breaking as we will trigger more errors

gmartpad:
When will these changes be added to a new version of react-native-webview. This might actually solve a problem where my page eternally loads because of assets failing to load, and onLoadEnd never triggers.

PR #3302, opened 2024-01-25
feat: move RNCWebViewImpl from objective-c to objective c++

### LINKED ISSUE #3298: NSInvalidArgumentException: -[RNCWebViewImpl destroyWebView]: unrecognized selector sent to instance 0x10b1e9a00
Hello,
i get this error when i navigate back from the webview screen

"""""""""""""""
NSInvalidArgumentException: -[RNCWebViewImpl destroyWebView]: unrecognized selector sent to instance 0x10b1e9a00
"""""""""""""""

The app crashes automatically. Thanks

Francesco-Voto:
Close #3298

### DIFF
--- a/apple/RNCWebViewImpl.mm
+++ b/apple/RNCWebViewImpl.mm
@@ -117,6 +117,7 @@ NSString *const CUSTOM_SELECTOR = @"_CUSTOM_SELECTOR_";
 @interface RNCWebViewImpl () <WKUIDelegate, WKNavigationDelegate, WKScriptMessageHandler, WKHTTPCookieStoreObserver,
 #if !TARGET_OS_OSX
 UIScrollViewDelegate,
+UIGestureRecognizerDelegate,
 #endif // !TARGET_OS_OSX
 RCTAutoInsetsProtocol>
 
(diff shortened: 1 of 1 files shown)


Titozzz:
Can you give more context on why this change fixes the issue ?

Francesco-Voto:
@Titozzz   the `RCT_NEW_ARCH_ENABLED` is not correctly read in the objective-c file, while it exists and with the right value in the objective-c++.

Francesco-Voto:
this means that this check with the new architecture return the wrong value

```
#ifdef RCT_NEW_ARCH_ENABLED
- (void)destroyWebView
#else
- (void)removeFromSuperview
#endif
```

and when the view is unmounted  a inexistent `destroyWebView` is called

Titozzz:
Can you reproduce with a version >= [13.6.4](https://github.com/react-native-webview/react-native-webview/releases/tag/v13.6.4) ?

Francesco-Voto:
I just tried with the latest version (13.8.1) and I still experiencing the crash (the problem is the one described in the issue).

I am using RN 0.73.2 and the new architecture enabled

farwayer:
I can confirm that the proble does exist (also with 13.8.1) and this PR fixes it
For anyone who tries to use this fix: please note that without renaming the file from '.m' to '.mm' it will not work!

cipolleschi:
Hi there! I'm Riccardo, from the React Native core team.

Thank you so much for raising this issue. The problem should not be addressed in WebView as this issue revealed a broader problem in React Native Core.
We are going to fix it there with [this pr](https://github.com/facebook/react-native/pull/43088).

farwayer:
@cipolleschi thanks. Will the problem only be fixed starting with 0.73.x or will the fix be backported to previous RN versions?

Francesco-Voto:
thanks @cipolleschi I will close this PR waiitng for the real fix to come

cipolleschi:
@farwayer, we will back port it.

PR #3307, opened 2024-01-28
Fixes https://github.com/react-native-webview/react-native-webview/is…

### LINKED ISSUE #3305: [Android / New Arch] `injectJavaScript` and presumably other commands get executed twice
**Bug description:**

I found that `injectJavaScript` code gets executed twice on Android with the New Arch. It happens because once it gets executed here (as a consequence of that command processing pattern):
https://github.com/react-native-webview/react-native-webview/blob/6960a1903c3c32e1e77be11e6370f1c384ed7ee4/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt#L334

then another time it gets executed here:
https://github.com/react-native-webview/react-native-webview/blob/6960a1903c3c32e1e77be11e6370f1c384ed7ee4/android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java#L444-L447

I guess, similar error might be there with other actions handled via that commands pattern.

**To Reproduce:**

**Expected behavior:**

**Screenshots/Videos:**

**Environment:**
 - OS:
 - OS version:
 - react-native version:
 - react-native-webview version:

birdofpreyru:
Fixes #3305.

### DIFF
--- a/android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -420,39 +420,46 @@ public class RNCWebViewManager extends ViewGroupManager<RNCWebViewWrapper>
         mRNCWebViewManagerImpl.setUserAgent(view, value);
     }
 
-    // These will never be called because we use the shared impl for now
   @Override
   public void goBack(RNCWebViewWrapper view) {
-    view.getWebView().goBack();
+    // NOTE: It is handled via receiveCommand().
+    // view.getWebView().goBack();
   }
 
   @Override
   public void goForward(RNCWebViewWrapper view) {
-    view.getWebView().goForward();
+    // NOTE: It is handled via receiveCommand().
+    // view.getWebView().goForward();
   }
 
   @Override
   public void reload(RNCWebViewWrapper view) {
-    view.getWebView().reload();
+    // NOTE: It is handled via receiveCommand().
+    // view.getWebView().reload();
   }
 
   @Override
   public void stopLoading(RNCWebViewWrapper view) {
-    view.getWebView().stopLoading();
+    // NOTE: It is handled via receiveCommand().
+    // view.getWebView().stopLoading();
   }
 
   @Override
   public void injectJavaScript(RNCWebViewWrapper view, String javascript) {
-      view.getWebView().evaluateJavascriptWithFallback(javascript);
+    // NOTE: It is handled via receiveCommand().
+    // view.getWebView().evaluateJavascriptWithFallback(javascript);
   }
 
   @Override
   public void requestFocus(RNCWebViewWrapper view) {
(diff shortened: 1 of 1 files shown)


Titozzz:
Hello I haven't made extensive testing on new arch, I'll hopefully review this soon!

android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
```diff
@@ -420,39 +420,46 @@ public void setUserAgent(RNCWebViewWrapper view, @Nullable String value) {
...
-      view.getWebView().evaluateJavascriptWithFallback(javascript);
+    // NOTE: It is handled via receiveCommand().
+    // view.getWebView().evaluateJavascriptWithFallback(javascript);
   }
 
   @Override
   public void requestFocus(RNCWebViewWrapper view) {
-      view.requestFocus();
+    // NOTE: It is handled via receiveCommand().
+    // view.requestFocus();
```
o-alexandrov, line 456: Why is the purpose of leaving empty overrides?

birdofpreyru: @o-alexandrov I don't have previous experience with native UI components, and I haven't looked much into how exactly that `codegenNativeCommands()` and `.receiveCommands()` stuff works. Just looking at the existing code and behavior it looked to me that if these `.requestFocus()`, _etc._ methods get called directly, perhaps they are the best way to implement related functionality (and probably they are abstract in a parent class, thus have to be overridden anyway &mdash; but I haven't really checked it). Then, there is the backward compatibility to the old arch, I thought if I touch `.receiveCommands()`, perhaps it is needed for the old arch, and it will need a bunch more efforts to ensure it works there.

My immediate interest here was just to hotfix the bug causing me troubles, thus I figured out this is the minimal intrusion way to do it &mdash; comment out the calls that cause duplicated command executions, but otherwise leave these method in the codebase so that somebody later looks second time into it.

Nantris:
This seems like an important issue to get addressed. Is there any progress on this front? Or anything in particular blocking progress?

birdofpreyru:
**bot** go away!

Nantris:
@Titozzz any chance you or anyone else will have a chance to take a look over this in the near future?

Titozzz:
I will!

Nantris:
Friendly bump.

This isn't affecting us (yet) but I noticed some more activity on the associated issue #3305. I hope this can be fixed soon.

@birdofpreyru, are you using this via `patch-package`, I assume? Any issues you've encountered?

birdofpreyru:
@Nantris I’ve just started my fork of the library, with this and few others issues important to me fixed, the latest updates from this upstream regularly pulled in, and released to npm as https://www.npmjs.com/package/@dr.pogodin/react-native-webview

PR #3314, opened 2024-01-31
mediaPlaybackRequiresUserAction defaults to false on iOS

Gregoirevda:
iOS: mediaPlaybackRequiresUserAction has no default value in JS and defaults to false in UIView


Gregoirevda:
@Titozzz I think this doc would help

PR #3309, opened 2024-01-30
[Windows][WebView2] Add support for custom headers, POST requests and `onOpenWindow` to Windows

jfkm69:
## Motivation
In a bid to create feature parity this PR looks to add support for several missing features on the Windows implementation of WebView with a focus on the future facing WebView2 component including custom headers, POST requests and link handling.

### Added
- Added `requestFocus` to WebView2 implementation
- Added `onOpenWindow` callback to allow for custom link handling
- Added `onSourceChanged` callback to allow consumer to request made by the browser
- Added missing Frame Navigation event listeners and missing commands (`postMessage`, `requestFocus` and `loadUrl`)

### Fixed
- Fixes to cookie management

PR #3319, opened 2024-02-06
[Windows][WebView2] Refactor ReactWebView2.cpp to handle optional 'method' parameter

jfkm69:
## Motivation
PR #3309 (https://github.com/react-native-webview/react-native-webview/pull/3309) introduced a bug whereby the call to create a web resource request would directly access the method property of the request object, however, in Typescript the method object is optional thus people that didn't include it implicitly experienced crashes due to the null reference.

### Fixed
- Refactor ReactWebView2.cpp to handle optional 'method' parameter


rafaelmonroy:
the crash i reported is no longer an issue with these updates

PR #3341, opened 2024-02-27
fix(ios): Allow keyboard accessory view to be toggled after initialization using `hideKeyboardAccessoryView` prop

### LINKED ISSUE #3339: Changing the `hideKeyboardAccessoryView` prop doesn't work
**Bug description:**
The `hideKeyboardAccessoryView` prop only seems to care about the initial value that is used when the webview is mounted. Changing it later using a `useState` does not do anything.

**To Reproduce:**
1. Have a useState like `const [hideKeyboardAccessoryView, setHideKeyboardAccessoryView] = useState(true)`
2. Use that state for the `hideKeyboardAccessoryView` prop
3. Change the value some time later using `setHideKeyboardAccessoryView`

Since the initial value was `true`, the keyboard accessory view will stay hidden even if set the state to `false` later.

**Expected behavior:**
The keyboard accessory view should be toggled according to the value of the `hideKeyboardAccessoryView` prop if it is changed after the initial value.

**Environment:**
 - OS: iOS
 - OS version: 16
 - react-native version: 0.72.5
 - react-native-webview version: 13.6.0

amanharwara:
fixes #3339

### DIFF
--- a/apple/RNCWebViewImpl.m
+++ b/apple/RNCWebViewImpl.m
@@ -910,12 +910,14 @@ RCTAutoInsetsProtocol>
 
 -(void)setHideKeyboardAccessoryView:(BOOL)hideKeyboardAccessoryView
 {
+  _savedHideKeyboardAccessoryView = hideKeyboardAccessoryView;
+
   if (_webView == nil) {
-    _savedHideKeyboardAccessoryView = hideKeyboardAccessoryView;
     return;
   }
 
   if (_savedHideKeyboardAccessoryView == false) {
+    [self __addInputAccessoryView];
     return;
   }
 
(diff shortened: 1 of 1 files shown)


jamonholmgren:
Does this work? I'm not sure how the view gets toggled in this code, although I could be missing something obvious.

amanharwara:
It does. I did create an example page locally, but thought maybe it wouldn't make sense to add it to the PR. Should I push that?

amanharwara:
> I'm not sure how the view gets toggled in this code

In the original code the view is hidden by returning `nil` from a `inputAccessoryView` method from a SwizzleHelperWK class that is created from the superclass of the [required subview](https://github.com/standardnotes/react-native-webview/blob/hideKeyboardAccessoryView-prop/apple/RNCWebViewImpl.m#L926-L929) and [set as the class for that subview](https://github.com/standardnotes/react-native-webview/blob/hideKeyboardAccessoryView-prop/apple/RNCWebViewImpl.m#L947).

The `__addInputAccessoryView` method added in this PR toggles the view back by basically undoing the above process by finding the created SwizzleHelperWK subview and then setting the class for that subview back to the superclass.

jamonholmgren:
@amanharwara Makes sense. Usually I'd include a GIF or screen recording showing it working, but that's not critical I don't think.

jamonholmgren (approved):
Seems straightforward to me.

amanharwara:
@jamonholmgren @Titozzz hello, pinging because this got labelled as stale. can it be merged given that there are no issues with it?

PR #3352, opened 2024-03-07
fix(android): Fix bridgeless mode support

Kudo:
# Why

react-native-webview does not work on react-native 0.74 with bridgeless mode

# How

there were two errors actually:
1. build error from override `shouldStartLoadWithLockIdentifier()`. the [newer codegen supports Int32](https://github.com/facebook/react-native/commit/ccd3b047709fda00123e8aec4294dce024b80244) and generates java code as int type. the current code is using double type. the pr tries to use double type as the implementation using double under the hood.
2.  on bridgeless mode, we cannot access `CatalystInstance`. it seems react-native-webview used it for sending direct message. BridgelessReactContext does not have equivalent `callFunction(moduleName: string, ...)`. this pr tries to register a single `RNCWebViewMessagingModule` module and check the underlying `messagingModuleName` inside the event. each webview will only receive events belong its `messagingModuleName`.

# Test Plan

- test example on android
- test example on android (with bumping react-native 0.74 and newArchEnabled=true)
- test on react-native 0.71 old architecture to make sure no regression

### DIFF
--- a/src/WebView.android.tsx
+++ b/src/WebView.android.tsx
@@ -30,9 +30,12 @@ import styles from './WebView.styles';
 
 const { resolveAssetSource } = Image;
 
-const registerCallableModule: (name: string, module: Object) => void =
-  require('react-native').registerCallableModule ??
-  BatchedBridge.registerCallableModule.bind(BatchedBridge);
+const registerCallableModule: (name: string, module: Object) => void
+  // `registerCallableModule()` is available in React Native 0.74 and above.
+  // Fallback to use `BatchedBridge.registerCallableModule()` for older versions.
+  // eslint-disable-next-line global-require
+  = require('react-native').registerCallableModule
+  ?? BatchedBridge.registerCallableModule.bind(BatchedBridge);
 
 /**
  * A simple counter to uniquely identify WebView instances. Do not use this for anything else.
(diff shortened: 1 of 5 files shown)


src/WebView.android.tsx
```diff
@@ -20,13 +20,23 @@ import {
...
 import styles from './WebView.styles';
 
 const { resolveAssetSource } = Image;
 
+const registerCallableModule: (name: string, module: Object) => void
+  // `registerCallableModule()` is available in React Native 0.74 and above.
+  // Fallback to use `BatchedBridge.registerCallableModule()` for older versions.
+  // eslint-disable-next-line global-require
+  = require('react-native').registerCallableModule
+  ?? BatchedBridge.registerCallableModule.bind(BatchedBridge);
```
RSNara, line 38: @Kudo, in 0.73, we also have access to global.RN$registerCallableModule in bridgeless mode. If you also include that here, then we could make this library bridgeless mode compatible in 0.73 as well.

Kudo: thanks @RSNara for taking a look and having the feedback. i'm lean toward to claim the bridgeless support for react-native-webview from 0.74, because i didn't run any bridgeless test on on 0.73.

Titozzz (approved):

Kudo:
late response but much appreciated for the help from @Titozzz to get the fix published.

andreialecu:
I believe this may have introduced a regression. #3368

src/WebView.android.tsx
```diff
@@ -119,12 +129,26 @@ const WebViewComponent = forwardRef<{}, AndroidWebViewProps>(({
...
+      if (event.messagingModuleName === messagingModuleName) {
+        const { messagingModuleName: _, ...rest } = event;
+        onMessage(rest);
+      }
+    },
+  }), [messagingModuleName, onMessage, onShouldStartLoadWithRequest]);
 
   useEffect(() => {
-    BatchedBridge.registerCallableModule(messagingModuleName, directEventCallbacks);
+    registerCallableModule('RNCWebViewMessagingModule', directEventCallbacks);
```
andreialecu, line 151: Should this first parameter be static and not `messagingModuleName`? Perhaps this is what may cause #3368

PR #3373, opened 2024-04-02
feat: Add visionOS support

Saadnajmi:
## Summary 

Add visionOS support, and an example test app via react-native-test-app. 

As it turns out, most of where we ifdeffed stuff out for macOS, is also where we need to ifdef out for visionOS. So a simple conversion of the ifdefs from "if !macOS" to "if iOS" does the job.

Additionally, we needed to port uses of `UIMenuController` to to `UIEditMenuInteraction`. This is only available on iOS 16+, so I did my best to guard behind availability checks and compiler ifdefs. I am not sure what our minimum iOS/Xcode version we want to support in this repo is, so I would appreciate a sanity check. 

## Test Plan

Test app works. Custom menus work on both iOS 16 and visionOS 1.0
Screenshot 2024-04-01 at 11 23 23 PM
(Sorry for the compressed video)

https://github.com/react-native-webview/react-native-webview/assets/6722175/fbab0541-c47c-4ae9-bbc3-84057605101e


apple/RNCWebViewImpl.m
```diff
@@ -170,10 +170,10 @@ - (instancetype)initWithFrame:(CGRect)frame
     _autoManageStatusBarEnabled = YES;
     _contentInset = UIEdgeInsetsZero;
     _savedKeyboardDisplayRequiresUserAction = YES;
-#if !TARGET_OS_OSX
+#if TARGET_OS_IOS
     _savedStatusBarStyle = RCTSharedApplication().statusBarStyle;
     _savedStatusBarHidden = RCTSharedApplication().statusBarHidden;
-#endif // !TARGET_OS_OSX
+#endif // !TARGET_OS_IOS
```
TheSavior, line 176: Does this need to remove the `!` to match?

apple/RNCWebViewImpl.m
```diff
@@ -650,7 +651,7 @@ -(void)keyboardDisplacementFix
     }];
   }
 }
-#endif // !TARGET_OS_OSX
+#endif // !TARGET_OS_IOS
```
TheSavior, line 654: same

TheSavior (commented):
I don't know this stuff very well, so drive by comment on something that looks funky

apple/RNCWebViewImpl.m
Saadnajmi: Good call, I missed it

Saadnajmi:
This PR works as far as I can tell, save custom menus. The method for showing custom actions in a context menu on iOS 16 / visionOS 1.0 changed (use `UIEditMenuConfiguration` instead of `UIMenuController`), so I'm seeing If I can refactor that bit. Otherwise, this change might be able to land without custom menu support

apple/RNCWebViewImpl.m
```diff
@@ -246,29 +252,51 @@ - (void)startLongPress:(UILongPressGestureRecognizer *)pressSender
...
         return;
     }
-    // When a long press ends, bring up our custom UIMenu if defined
-    if (self.menuItems.count == 0) {
+    if (@available(iOS 16.0, *)) {
+      CGPoint location = [pressSender locationInView:self];
+      UIEditMenuConfiguration *config = [UIEditMenuConfiguration configurationWithIdentifier:nil sourcePoint:location];
+      [_editMenuInteraction presentEditMenuWithConfiguration:config];
+    } else {
+      // When a long press ends, bring up our custom UIMenu if defined
```
Saadnajmi, line 260: If we need to support older versions of Xcode, I can put this behind a compiler macro like we do elsewhere, along with any other references.

Saadnajmi:
Hello @Titozzz / @jamonholmgren , I was wondering if there was anything I could do to help move this PR forward? Appreciate your time spent maintaining this repo :)

Titozzz (approved):

PR #3403, opened 2024-04-19
Support for visionOS

TrevorDev:
This was not compatible with https://github.com/callstack/react-native-visionos 

- Marked package as supporting visionOS
- Avoid api's that xcode says are not available on visionos

Tested by following:
https://callstack.github.io/react-native-visionos-docs/docs/getting-started/create-first-app


TrevorDev:
Oh dang, looks like this is already supported via https://github.com/react-native-webview/react-native-webview/pull/3373/files

Saadnajmi:
@TrevorDev the PR hasn't merged yet, but I suppose you can locally use patch-package to get my changes

PR #3411, opened 2024-04-23
Adds support for compiling Windows module for WinAppSDK

rozele:
Similar to other packages (react-native-svg,
@react-native-picker/picker, etc.), this change adds WinAppSDK support for apps that compile react-native-windows from source targeting WinAppSDK.


chiaramooney (approved):
👍

Titozzz (approved):

Issue #3412, opened 2024-04-24
App crashes on iOS when opening websites with Cloudflare protection


opensourcex123:
**Bug description:**
On the ios platform, when i open the webview with uri: https://rocketpool.net, the app crashed, and when i open it on the website, it gave me a 403 forbidden error. All console were not printed according to the code below.

**To Reproduce:**
```  return (
    <WebView
      onError={syntheticEvent => {
        const {nativeEvent} = syntheticEvent;
        console.warn('WebView error: ', nativeEvent);
      }}
      onLoadProgress={({type, nativeEvent}) => {
        console.log('load', nativeEvent);
      }}
     onHttpError={syntheticEvent => {
        const {nativeEvent} = syntheticEvent;
        console.warn(
          'WebView received error status code: ',
          nativeEvent.statusCode,
        );
      }}
      onLoadEnd={event => {
        console.log('load end', event);
      }}
      onContentProcessDidTerminate={syntheticEvent => {
        const {nativeEvent} = syntheticEvent;
        console.warn('Content process terminated, reloading', nativeEvent);
        this.refs.webview.reload();
      }}
      renderError={errorName => <Text>{errorName}</Text>}
      source={{
        uri: 'https://rocketpool.net',
        headers: {
          'User-Agent':
            'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36',
        },
      }}
    />
  );
```

**Expected behavior:**
Al least show me some cloudflare website, like when i open it on the pc.
![image](https://github.com/react-native-webview/react-native-webview/assets/73230562/cfef6ca0-eeb8-4dfd-8b39-825123979998)

**Screenshots/Videos:**

**Environment:**
 - OS:ios
 - OS version:16.7.7
 - react-native version:0.73.5
 - react-native-webview version:13.6.4

PR #3416, opened 2024-04-25
Fix crash when opening 4xx pages with no description.

bimusiek:
Hey, at https://freeyourmusic.com we have a weird case, where user is redirected to a page with 404 error that crashes on iOS. After further investigation, we found that this is because of `description=NULL` as seen on screenshot:
CleanShot 2024-04-25 at 15 24 05@2x

This PR fixes this issue and possibly https://github.com/react-native-webview/react-native-webview/issues/3412

### DIFF
--- a/apple/RNCWebView.mm
+++ b/apple/RNCWebView.mm
@@ -109,7 +109,7 @@ auto stringToOnLoadingFinishNavigationTypeEnum(std::string value) {
                     .lockIdentifier = [[dictionary valueForKey:@"lockIdentifier"] doubleValue],
                     .title = std::string([[dictionary valueForKey:@"title"] UTF8String]),
                     .code = [[dictionary valueForKey:@"code"] intValue],
-                    .description = std::string([[dictionary valueForKey:@"description"] UTF8String]),
+                    .description = std::string([[dictionary valueForKey:@"description"] UTF8String] ?: ""),
                     .canGoBack = static_cast<bool>([[dictionary valueForKey:@"canGoBack"] boolValue]),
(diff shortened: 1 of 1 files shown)


apple/RNCWebView.mm
```diff
@@ -109,7 +109,7 @@ - (instancetype)initWithFrame:(CGRect)frame
                     .lockIdentifier = [[dictionary valueForKey:@"lockIdentifier"] doubleValue],
                     .title = std::string([[dictionary valueForKey:@"title"] UTF8String]),
                     .code = [[dictionary valueForKey:@"code"] intValue],
-                    .description = std::string([[dictionary valueForKey:@"description"] UTF8String]),
+                    .description = std::string([[dictionary valueForKey:@"description"] UTF8String] ?: ""),
```
DorianMazur, line 112: You need to do the same on line 233.

DorianMazur:
@Titozzz @jamonholmgren Can you check this? This is important PR and I think should be merged after applying change suggested in my comment.

apple/RNCWebView.mm
bimusiek: Done.

bimusiek:
@chiaramooney @Titozzz I would love to have this merged. We still use simple patch now, but this is quite common crash for any non 200 page that does not return description. Also, as reported by someone else, this is also crashing for Cloudflare protected pages in case of 403 there.

Seemed like quite nice bug fix, but it is more than a month now without any response or comment? :)

TheAlmightyBob (approved):

TheAlmightyBob:
@bimusiek It's saying "This branch is out-of-date with the base branch" but not giving me any option to rebase it (don't think I've seen GitHub do that before...?). Are you able to update this branch?

PR #3417, opened 2024-04-25
fix(Android): Trimming file names for download to fix File Name too long error

salla-andre:
When the download file name size + the download directory path size exceeds the max file path size, the app will crash with a "File Name too long" exception.

To avoid that, I'm splitting the file name to get the extension, trimming the file name (w/o extension) and then putting the extension back to the file name, so it doesn't break while requesting the download.

IMG_0389

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
@@ -28,8 +28,10 @@ import java.io.UnsupportedEncodingException
 import java.net.MalformedURLException
 import java.net.URL
 import java.util.Locale
+import kotlin.math.min
 
 val invalidCharRegex = "[\\\\/%\"]".toRegex()
+val splitExtensionRegex = "\\.(?=[^\\.]+$)".toRegex()
 
 class RNCWebViewManagerImpl {
     companion object {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
```diff
@@ -49,6 +51,7 @@ class RNCWebViewManagerImpl {
     private val HTML_ENCODING = "UTF-8"
     private val HTML_MIME_TYPE = "text/html"
     private val HTTP_METHOD_POST = "POST"
+    private val MAX_FILE_PATH_SIZE = 255
```
TheAlmightyBob, line 54: How was this number determined? I have not found any information about there being such a small limit on Android.

TheAlmightyBob (changes requested):

android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
salla-andre: @TheAlmightyBob it's the `_POSIX_PATH_MAX` value. Actually, `_POSIX_PATH_MAX` is 256, but we need to consider the null ending byte in the char byte array, so we have 255 chars left for path name.

PR #3548, opened 2024-09-04
fix(android): Support video upload with capture attribute

StephanCodes:
My first time contributing, so any feedback is welcome!

---

Currently, when you use a file input for videos with the `capture` attribute in the webview, it does not appear to work as intended.

In my specific situation, I have a file input that look like this:
```html
<input type="file" accept="video/*" capture="user">
```

On Android 14, this will result in the `there is no Camera permission` error, since it will look for a `photoIntent, but no photoIntent is requested.

I made a fairly simple change that will first check for a `photoIntent` and then check for the `videoIntent`.


escherlies (approved):

StephanCodes:
@Titozzz If possible, I would still like to implement this fix.

PR #3552, opened 2024-09-09
fix(android, new arch): Do not execute commands twice

### LINKED ISSUE #3305: [Android / New Arch] `injectJavaScript` and presumably other commands get executed twice

Titozzz:
No need to manually call the methods, the generated delegate already does that for us.

Fixes #3305

### DIFF
--- a/android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
+++ b/android/src/newarch/com/reactnativecommunity/webview/RNCWebViewManager.java
@@ -426,7 +426,6 @@ public class RNCWebViewManager extends ViewGroupManager<RNCWebViewWrapper>
         mRNCWebViewManagerImpl.setUserAgent(view, value);
     }
 
-    // These will never be called because we use the shared impl for now
   @Override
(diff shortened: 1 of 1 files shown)


17Amir17:
🎇

PR #3564, opened 2024-09-20
fix-android-capture:fix Video Capture is not supported on Android.

Yorick32:
When I have to capture video in Webview, like this:

`<input type="file" accept="video/*" capture="user">`

I could't get any response when i click "Choose file" on Android. Then i get "there is no Camera permission" by checking  the Logcat. So I submit some code to fix this issue.


escherlies (approved):

frankfancode:
Thanks to @Yorick32
This pull request addresses video capture support on Android.
It functions correctly, as I have tested it.

PR #3575, opened 2024-10-04
[#3014] Fix for the issue 3014

birdofpreyru:
- #3014:
  **Android:** Adds `requestFocus()` and `onFocusChanged()` methods to the native WebView implementation, to only accept input focus when a focusable element inside the WebView is hit by a touch, and to correctly reset the hit test when WebView loses the focus. It fixes the issue with ScollView jump when a WebView is touched &mdash; the ScrollView automatically scrolls to its child when the child accepts input focus, and before this patch the WebView was accepting the focus on touch anywhere within WebView rect.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
@@ -341,6 +341,14 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
         return true;
     }
 
+    @Override
+    public boolean requestFocus(int direction, Rect previouslyFocusedRect) {
+      var hit = this.getHitTestResult();
+      if (hit.getType() == HitTestResult.EDIT_TEXT_TYPE) {
+        return super.requestFocus(direction, previouslyFocusedRect);
+      } else return false;
+    }
+
     protected void onScrollChanged(int x, int y, int oldX, int oldY) {
         super.onScrollChanged(x, y, oldX, oldY);
 
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
@@ -2,6 +2,8 @@ package com.reactnativecommunity.webview;
 
 import android.annotation.SuppressLint;
 import android.graphics.Rect;
+import android.os.Handler;
+import android.os.SystemClock;
 import android.text.TextUtils;
 import android.view.ActionMode;
 import android.view.Menu;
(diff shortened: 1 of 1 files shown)


steffenkleinle:
Hey @Titozzz, could you perhaps have a look at this? It fixes a long-lasting bug and problem we had with react-native-webview and is a valuable addition in my opinion. Thank you for your work!

cguino:
lastFocusResetTimestamp is not used, except to set : lastFocusResetTimestamp = 0

I don't think overriding onFocusChanged is usefull

You should replace 
- `var hit = this.getHitTestResult();` with `WebView.HitTestResult hit = this.getHitTestResult();` 
- `var now = SystemClock.uptimeMillis();` with `long now = SystemClock.uptimeMillis();`

birdofpreyru:
@cguino 
> _lastFocusResetTimestamp is not used, except to set : lastFocusResetTimestamp = 0_

wrong :) it is used to abort the overridden `onFocusChanged()` operation, if it is triggered within 100ms from the previous call, as there is `return` statement on the line 377.

> _I don't think overriding onFocusChanged is usefull_

Ogh, it is useful :) Without it, the "jumping issue" will manifest itself on subsequent clicks on the WebView if a previous click led to an input element inside the WebView gaining the focus.

cguino:
@birdofpreyru tkx for reply

> @cguino
> 
> > _lastFocusResetTimestamp is not used, except to set : lastFocusResetTimestamp = 0_
> 
> wrong :) it is used to abort the overridden `onFocusChanged()` operation, if it is triggered within 100ms from the previous call, as there is `return` statement on the line 377.
> 

I dont see where `lastFocusResetTimestamp` has a value that is not 0 :
 
Capture d’écran 2024-12-16 à 22 12 02

So, l.393 the if statement is always false : `now - lastFocusResetTimestamp < 100`

Maybe I missed something ?
Or Maybe you intend to set `lastFocusResetTimestamp = now;` at l.401 ? elsewhere ?


> > _I don't think overriding onFocusChanged is usefull_
> 
> Ogh, it is useful :) Without it, the "jumping issue" will manifest itself on subsequent clicks on the WebView if a previous click led to an input element inside the WebView gaining the focus.
> 

I tried to comment out the whole `onFocusChanged` function and it didn't make any difference so I was just wondering 
what it was supposed to do 

> > You should replace...
> 
> Yeah, that's a good suggestion :)

;)

birdofpreyru:
@cguino

> _I dont see where lastFocusResetTimestamp has a value that is not 0 :_

Ogh... that's a great catch! The intention was to have line 383 to read `lastFocusResetTimestamp = now;`. Not a big deal, as it is a precaution against a very unfortunate edge-case when the page point `(-12345, -12345)`, where a click is emulated, contains a focus-able element.

> _I tried to comment out the whole onFocusChanged function and it didn't make any difference so I was just wondering
what it was supposed to do_

Yeah, you can see the difference in the test I added to the example app in my fork (https://github.com/birdofpreyru/react-native-webview/blob/master/example/examples/WebViewInsideScrollView.tsx); if you click on the input element inside the WebView, then on the input element outside the WebView, and then on empty place in the WebView &mdash; without `onFocusChanged()` the scrolling position should jump in the end, as the hit test inside `requestFocus()` will return stale hit state, corresponding to the first click in the WebView.

PR #3585, opened 2024-10-15
React native webview shared instance

Jaybhade:


Titozzz:
Hello, I'm planning on rewriting webview to be able to have instances separate from the rendering of the webview. This PR will probably not make it as is, but this idea behind it is good

Jaybhade:
Hi @Titozzz , what's missing in this PR? What can I improve in it?

Jaybhade:
Hi @Titozzz , I am using this implementation in one of our production app. Is anything wrong with it or anything that I can improve?

Titozzz:
I don't think it's necessarily broken, it just has many other features that would use this. Pre loading, messaging with the instance even when not shown. Stuff like that but the API will be quite different

Issue #3439, opened 2024-05-14
Use WebViewCompat.addWebMessageListener instead of @JavascriptInterface to prevent origin spoofing


tian000:
Currently if the web app redirects quickly after doing `window.ReactNativeWebview.postMessage` you could receive a spoofed origin on the react native side.

We should use WebViewCompat.addWebMessageListener to prevent origin spoofing.

This is critical for security sensitive apps like crypto wallets. 

I can put up a PR for this if maintainers a willing to accept a fix for this issue.

otech47:
could you elaborate further how an attacker would exploit this? would an attacker aiming to spoof a domain need control over the web app that is redirecting quickly to then trick a user into thinking they are on google.com when really they are using maliciouswebsite.com?

tian000:
Exactly the malicious attacker controls the malicious site (usually a phishing site) and tricks the user into thinking it is a legitimate site via fast redirects.

It works something like this:
1. User lands on malicioussite.com
2. malicious site redirects to `google.com`, and sends a request to the mobile app with a delay e.g. 
    ```
    setTimeout(() => window.ReactNativeWebview.postMessage("I am google"), 5)
    ```
3. The mobile app thinks that the request is coming from google.com even though it is coming from malicioussite.com

tian000:
@otech47 what do you think?

otech47:
> @otech47 what do you think?

I didn't attempt to reproduce. I just wanted to confirm the low risk to a use case in which the user specifies which URL they access so the security impact for that is minimal imo

But I'm sure it is more dangerous in plenty of other use cases and should be fixed

rafinskipg:
Yeah, we see it being reported several times as a very dangerous origin spoofing issue. Since it can fake the website that is emitting the messages and make the users believe that a legitimate site is requesting them to sign some transaction etc

tian000:
@otech47 i've put up a PR for the fix here: https://github.com/react-native-webview/react-native-webview/pull/3609

tian000:
Fixed in https://github.com/react-native-webview/react-native-webview/pull/3609

PR #3609, opened 2024-11-14
fix(android,ios): Sending correct url in onMessage callback

tian000:
https://github.com/react-native-webview/react-native-webview/issues/3439

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
@@ -56,7 +56,8 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
     protected static final String JAVASCRIPT_INTERFACE = "ReactNativeWebView";
     protected @Nullable
     RNCWebViewBridge fallbackBridge;
-    protected WebViewCompat.WebMessageListener bridgeListener = null;
+    protected @Nullable
+    WebViewCompat.WebMessageListener bridgeListener = null;
 
     /**
      * android.webkit.WebChromeClient fundamentally does not support JS injection into frames other
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
@@ -251,13 +251,6 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
     }
 
     protected void createRNCWebViewBridge(RNCWebView webView) {
-        if (getSettings().getJavaScriptEnabled() &&
-          injectedJavaScriptObject != null &&
-          !TextUtils.isEmpty(injectedJavaScriptObject)) {
-          evaluateJavascriptWithFallback("(function(){\n" +
-            "    window.ReactNativeWebView.injectedObjectJson = \"" + injectedJavaScriptObject + "\";\n" +
-            "})()");
-        }
         if (WebViewFeature.isFeatureSupported(WebViewFeature.WEB_MESSAGE_LISTENER)){
           if (this.bridgeListener == null) {
             this.bridgeListener = new WebViewCompat.WebMessageListener() {
@@ -279,6 +272,19 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
             addJavascriptInterface(fallbackBridge, JAVASCRIPT_INTERFACE);
           }
         }
+        injectJavascriptObject();
+    }
+
+    private void injectJavascriptObject() {
+      if (getSettings().getJavaScriptEnabled() &&
+        injectedJavaScriptObject != null &&
+          !TextUtils.isEmpty(injectedJavaScriptObject)) {
+        String js = "(function(){\n" +
+          "    window." + JAVASCRIPT_INTERFACE + " = window." + JAVASCRIPT_INTERFACE + " || {}; \n" +
+          "    window." + JAVASCRIPT_INTERFACE + ".injectedObjectJson = function () { return `" + injectedJavaScriptObject + "`; };\n" +
+          "})();";
+        evaluateJavascriptWithFallback(js);
+      }
     }
 
     @SuppressLint("AddJavascriptInterface")
@@ -303,6 +309,7 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
                 injectedJS != null &&
                 !TextUtils.isEmpty(injectedJS)) {
             evaluateJavascriptWithFallback("(function() {\n" + injectedJS + ";\n})();");
+            injectJavascriptObject(); // re-inject the Javascript object in case it has been overwritten.
         }
     }
 
@@ -311,16 +318,15 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
                 injectedJSBeforeContentLoaded != null &&
                 !TextUtils.isEmpty(injectedJSBeforeContentLoaded)) {
             evaluateJavascriptWithFallback("(function() {\n" + injectedJSBeforeContentLoaded + ";\n})();");
+            injectJavascriptObject();  // re-inject the Javascript object in case it has been overwritten.
         }
     }
 
-    protected String injectedJavaScriptObject = null;
+    protected String injectedJavaScriptObject = "null";
 
     public void setInjectedJavaScriptObject(String obj) {
-        if (getSettings().getJavaScriptEnabled()) {
-            this.injectedJavaScriptObject = obj;
-            this.createRNCWebViewBridge(this);
-        }
+      this.injectedJavaScriptObject = obj;
+      injectJavascriptObject();
     }
 
     public void onMessage(String message, String sourceUrl) {
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
@@ -54,6 +54,9 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
     protected @Nullable
     String injectedJSBeforeContentLoaded;
     protected static final String JAVASCRIPT_INTERFACE = "ReactNativeWebView";
+    protected @Nullable
+    RNCWebViewBridge fallbackBridge;
+    protected WebViewCompat.WebMessageListener bridgeListener = null;
 
     /**
      * android.webkit.WebChromeClient fundamentally does not support JS injection into frames other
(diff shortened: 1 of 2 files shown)


tian000:
@Titozzz @jamonholmgren could you take a look

tomByrer:
Thanks for a fix!

tian000:
@brentvatne @Titozzz @jamonholmgren bump!

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -242,12 +247,31 @@ public boolean getMessagingEnabled() {
...
+            this.bridgeListener = new WebViewCompat.WebMessageListener() {
+                @Override
+                public void onPostMessage(@NonNull WebView view, @NonNull WebMessageCompat message, @NonNull Uri sourceOrigin, boolean isMainFrame, @NonNull JavaScriptReplyProxy replyProxy) {
+                    RNCWebView.this.onMessage(message.getData(), sourceOrigin.toString());
+                }
+            };
+            WebViewCompat.addWebMessageListener(
+                    webView,
+                    JAVASCRIPT_INTERFACE,
+                    Set.of("*"),
```
Kudo, line 264: having a question here. if we put `*` here, what's the difference between original `addJavascriptInterface` approach?
maybe we can reuse the [`originWhitelist`](https://github.com/react-native-webview/react-native-webview/blob/4733fe7988f1a4489f0d430ffcbe152416fe1b0e/docs/Reference.md#originwhitelist) prop?

tian000: This PR does not solve the fact that messages are coming from unwanted origins, i.e. origins not on a whitelist, but rather messages are coming from origin A but this library says that its coming from Origin B. 

The `*` here is just saying any origin can send messages over the bridge. The difference between the previous implementation is how we retrieve the origin of the message.

In the previous implementation we retrieve the origin of the message using `webView.getUrl()` **after** the message has been sent. The webview could be on a different page by the time we call `webView.getUrl()`.

By using a `WebMessageListener` in this PR, we get the origin **with** the message. Therefore, even if the site navigates away the instant after sending the message - the origin will still be the origin of the original sender.

tian000:
@Kudo could i get a re-review when you have time!

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
Kudo: thanks for the great explanation. so to achieve the spoofing prevention as you described in #3439, the callsite (app code) have to further check the `url` from the `onMessage` callback? this pr just to make sure the url passing into onMessage is correct. is my understanding right?

tian000: Yup! exactly. 
This PR just makes sure that the `url` on the `onMessage` is correct. 
Each app can then decide in its own way - whether they want to process or ignore the messages.

Kudo: cool! thanks for confirming that. i think the pr title may be too strong. i was scared by the title. 😅 that's not a real vulnerability, or at least we still need a further check to prevent the vulnerability.
how about changing the title like `fix(android,ios): Sending correct url in onMessage callback`

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -242,12 +247,31 @@ public boolean getMessagingEnabled() {
...
 
-    protected RNCWebViewBridge createRNCWebViewBridge(RNCWebView webView) {
-        if (bridge == null) {
-            bridge = new RNCWebViewBridge(webView);
-            addJavascriptInterface(bridge, JAVASCRIPT_INTERFACE);
+    protected WebViewCompat.WebMessageListener bridgeListener = null;
+
+    protected void createRNCWebViewBridge(RNCWebView webView) {
+        if (this.bridgeListener == null &&
+                WebViewFeature.isFeatureSupported(WebViewFeature.WEB_MESSAGE_LISTENER)) {
```
Kudo, line 254: should we also support fallback if the feature is not available? maybe we should keep original code as a fallback path.

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
tian000: Updated the title!

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
tian000: Added the previous code path as fallback!

tian000:
@Kudo Responded to the comments!

tian000:
@Kudo happy new year! could i get a re-review on this

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -242,12 +250,35 @@ public boolean getMessagingEnabled() {
...
-    protected RNCWebViewBridge createRNCWebViewBridge(RNCWebView webView) {
-        if (bridge == null) {
-            bridge = new RNCWebViewBridge(webView);
-            addJavascriptInterface(bridge, JAVASCRIPT_INTERFACE);
+    protected void createRNCWebViewBridge(RNCWebView webView) {
+        if (getSettings().getJavaScriptEnabled() &&
+          injectedJavaScriptObject != null &&
+          !TextUtils.isEmpty(injectedJavaScriptObject)) {
+          evaluateJavascriptWithFallback("(function(){\n" +
+            "    window.ReactNativeWebView.injectedObjectJson = \"" + injectedJavaScriptObject + "\";\n" +
```
Kudo, line 258: `window.ReactNativeWebView.injectedObjectJson` should be a function returning json.
please see equivalent ios implementation:  https://github.com/react-native-webview/react-native-webview/blob/4733fe7988f1a4489f0d430ffcbe152416fe1b0e/apple/RNCWebViewImpl.m#L1692-L1695

please also test your change on the [examples/Injection](https://github.com/react-native-webview/react-native-webview/blob/master/example/examples/Injection.tsx) test case.

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -283,14 +314,16 @@ public void callInjectedJavaScriptBeforeContentLoaded() {
...
     }
 
+    protected String injectedJavaScriptObject = null;
+
     public void setInjectedJavaScriptObject(String obj) {
         if (getSettings().getJavaScriptEnabled()) {
-            RNCWebViewBridge b = createRNCWebViewBridge(this);
-            b.setInjectedObjectJson(obj);
+            this.injectedJavaScriptObject = obj;
+            this.createRNCWebViewBridge(this);
```
Kudo, line 322: not a regression but an existing bug.
when testing on `examples/Injection` test case. i noticed that `setJavaScriptEnabled` coming after `setInjectedJavaScriptObject` (because we cannot ensure the order of prop updates). in this case, the injectedJavaScriptObject will not be updated because the `getJavaScriptEnabled()` is false here.

but previously it's just `window.ReactNativeWebView.injectedObjectJson()` returning null. now it would be `window.ReactNativeWebView.injectedObjectJson === undefined`

Kudo (commented):
sorry for late. having some regression from `injectedObjectJson()`. please take a look when you get a chance

tian000:
@Kudo I've pushed some changes based on your suggestions!

Here are the test results!
Screenshot_1735913868
Screenshot_1735913872

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
tian000: updated!

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
tian000: makes sense, i've defaulted injectedObjectJson to returning null.

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -48,7 +55,8 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
     String injectedJSBeforeContentLoaded;
     protected static final String JAVASCRIPT_INTERFACE = "ReactNativeWebView";
     protected @Nullable
-    RNCWebViewBridge bridge;
+    RNCWebViewBridge fallbackBridge;
+    protected WebViewCompat.WebMessageListener bridgeListener = null;
```
Kudo, line 59:
```suggestion
    protected @Nullable
    WebViewCompat.WebMessageListener bridgeListener = null;
```

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -242,12 +250,41 @@ public boolean getMessagingEnabled() {
...
-        return bridge;
+        injectJavascriptObject();
+    }
+
+    private void injectJavascriptObject() {
+      if (getSettings().getJavaScriptEnabled() &&
+        injectedJavaScriptObject != null &&
+          !TextUtils.isEmpty(injectedJavaScriptObject)) {
+        String js = "(function(){\n" +
+          "    window." + JAVASCRIPT_INTERFACE + " = window." + JAVASCRIPT_INTERFACE + " || {}; \n" +
```
Kudo, line 283:
```suggestion
          "    window." + JAVASCRIPT_INTERFACE + " = window." + JAVASCRIPT_INTERFACE + " || {};\n" +
```

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -242,12 +250,41 @@ public boolean getMessagingEnabled() {
...
+          "    window." + JAVASCRIPT_INTERFACE + ".injectedObjectJson = function () { return `" + injectedJavaScriptObject + "`; };\n" +
```
Kudo, line 284: note that when `injectedJavaScriptObject` is null. the `window.ReactNativeWebView.injectedObjectJson()` returns `"null"` (a "null" string).

would be good to return true null

Kudo (commented):
much better now. just a small nit remaining. thanks again for the great work

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
tian000: ✅

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
tian000: ✅

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
tian000: :whi

tian000:
@Kudo i've addressed the nits and the null issue

tian000:
@Kudo could i get a re-review when you have time

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
Kudo: hey @tian000, the last nit comment here is that it seems no Nullable here in the latest pr changes. could you double check again?

tian000: oops missed this one - added nullable!

tian000:
@Kudo i've added the nullable annotation

Kudo (approved):
sorry for having the back-and-forth in the pr. thanks again for the pr!

tian000:
@jamonholmgren @Titozzz could I get your reviews as codeowners.

tian000:
@jamonholmgren @Titozzz bump!

Titozzz:
Hey this is the next thing on my to-do but I have a lot of work related things m I need to do first! Hang in there!

tian000:
@Titozzz Just wanted to quickly follow up to see if you can review this PR this week!

Titozzz (approved):

Titozzz:
LGTM ! Thanks for the work, and sorry for the delay 😓

aliceni81:
Hi, onMessage for Android does not work for v13.13.0, v13.13.1 and v13.13.2 but works on iOS.

aliceni81:
@tian000 Hi, onMessage does not work for Android ar... It works on Android with v13.12.5 indeed.

Kudo:
@aliceni81 could you elaborate the problem and maybe share a repro? that would help for further investigation. did that happen to all android versions or just specific android versions?

ifsnow:
@Kudo Please note that onMessage does not work properly on Android 9.

ifsnow:
@Kudo The issue with Android 9 has been resolved in version 13.13.5. Thank you for the prompt action.

android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
```diff
@@ -242,12 +251,39 @@ public boolean getMessagingEnabled() {
...
-        if (bridge == null) {
-            bridge = new RNCWebViewBridge(webView);
-            addJavascriptInterface(bridge, JAVASCRIPT_INTERFACE);
+    protected void createRNCWebViewBridge(RNCWebView webView) {
+        if (WebViewFeature.isFeatureSupported(WebViewFeature.WEB_MESSAGE_LISTENER)){
+          if (this.bridgeListener == null) {
+            this.bridgeListener = new WebViewCompat.WebMessageListener() {
+              @Override
+              public void onPostMessage(@NonNull WebView view, @NonNull WebMessageCompat message, @NonNull Uri sourceOrigin, boolean isMainFrame, @NonNull JavaScriptReplyProxy replyProxy) {
+                RNCWebView.this.onMessage(message.getData(), sourceOrigin.toString());
```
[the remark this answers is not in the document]
otech47: > Expected behavior:
> It should be the full url

why?

postmessage mdn [docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) don't specify that the full URL must be passed into the message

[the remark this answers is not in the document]
Thristhart: Every other WebViewNativeEvent has the full URL in the `url` property. Before this change, so did this one. (And, if you're targeting a webview that doesn't have WEB_MESSAGE_LISTENER, it's still the full URL rather than the origin.)

IMO, this should be a new `origin` property, and the `url` should remain the full page URL.

Issue #3578, opened 2024-10-07
[v13.12.3][iOS] Broken loading of JS (and probably some other) assets by web pages in WebView


birdofpreyru:
**Bug description:**

Web pages in WebView fail to load JavaScript (and perhaps some other) assets on iOS. Just commenting out `didMoveToSuperview` implementation makes it work fine.

I guess, if initialization is done inside `didMoveToSuperview`, the page load begins while the webview is not visible on the screen, and in that case it just does not bother to load JS, as some sort of optimization.

**To Reproduce:**

**Expected behavior:**

**Screenshots/Videos:**

**Environment:**
 - OS:
 - OS version:
 - react-native version:
 - react-native-webview version:

nullCode666:
me too...

0xFA11:
+1, on latest RN0.76 (RNVW13.12.3) this issue still persists on iOS.

EliasVal:
+1, WebView half-loads and gets fully loaded on state refresh (i.e adding an element to trigger HMR)
Is there any workaround till this gets fixed?

birdofpreyru:
@EliasVal you may switch to my fork of the library &mdash; https://www.npmjs.com/package/@dr.pogodin/react-native-webview

craiganderson-iotv:
@EliasVal 

You can trigger a full reload by just changing the `key` prop:

```tsx
const [key, setKey] = useState('webview')

<Webview
  key={key}
  ...
/>

// call setKey() with a new value when the issue occurs
```

There doesn't seem to be a mechanism to detect when it fails to load properly, so for now, you can always do a full reload:

```tsx
const [key, setKey] = useState('webview')

<Webview
  key={key]
  onLoad={() => {
    if (key === 'webview')
      setKey('webview_reloaded')
  }
/>
```

The specific values for `key` are not important, as long as they are unique

EliasVal:
> @EliasVal you may switch to my fork of the library — https://www.npmjs.com/package/@dr.pogodin/react-native-webview

Amazing work!

holdenchristiansen:
> @EliasVal you may switch to my fork of the library — https://www.npmjs.com/package/@dr.pogodin/react-native-webview

This worked for me! All assets load and JS executes on initial render. And no errors on injectJavaScript either. Thanks @birdofpreyru

rodmastro32:
Hi, how can I reproduce this?

fondue-tech:
[...]

PR #3559, opened 2024-09-13
fix(ios): Fix webview not loading in background

sheat-git:
Hi there!

Sorry for the lack of details in this PR. I'm not great with English, and this is my first time with Objective-C. This change is related to issues we've been discussing.

Feel free to ask questions or suggest changes. I'm open to any feedback and eager to learn.

---

PRs:
- https://github.com/react-native-webview/react-native-webview/pull/2930
- https://github.com/react-native-webview/react-native-webview/pull/3041

Issues:
- https://github.com/react-native-webview/react-native-webview/issues/2431
- https://github.com/react-native-webview/react-native-webview/issues/3481

PR #3615, opened 2024-11-21
revert (e8f3c0e) fix(iOS): preload WebView on initial render in background tabs

0xFA11:
PR #3559 which was committed by e8f3c0e causes a regression in a patch release (v13.12.3).
Issue #3578 is open for more than a month and there's even another patch release released after that (v13.12.4).
I'm worried that this regression would be buried and stay unnoticed if we continue ignoring it and don't address.

For more context, Expo DOM (use dom) just started to gain some traction and it relies on RNWV.
Expo managed version of RNWV is 13.12.4 and it'd definitely hit much more people very soon when people start to use it more often.
Maybe pre-13.12.3, this regression wasn't a wild-spread issue but I think we're about to see a lot more people facing it.

Let's revert this relatively small and easy to undo PR and try again another day in another attempt to fix the original issue?

### DIFF
--- a/apple/RNCWebViewImpl.m
+++ b/apple/RNCWebViewImpl.m
@@ -504,12 +504,9 @@ RCTAutoInsetsProtocol>
   return wkWebViewConfig;
 }
 
-- (void)initializeWebView
+- (void)didMoveToWindow
 {
-  @synchronized (self) {
-    if (_webView != nil) {
-      return;
-    }
+  if (self.window != nil && _webView == nil) {
     WKWebViewConfiguration *wkWebViewConfig = [self setUpWkWebViewConfig];
     _webView = [[RNCWKWebView alloc] initWithFrame:self.bounds configuration: wkWebViewConfig];
     [self setBackgroundColor: _savedBackgroundColor];
@@ -561,23 +558,6 @@ RCTAutoInsetsProtocol>
     [self setHideKeyboardAccessoryView: _savedHideKeyboardAccessoryView];
     [self setKeyboardDisplayRequiresUserAction: _savedKeyboardDisplayRequiresUserAction];
     [self visitSource];
-  }
-}
-
-// react-native-mac os does not support didMoveToSuperView
-#if !TARGET_OS_OSX
-- (void)didMoveToSuperview
-{
-  if (self.superview != nil && _webView == nil) {
-    [self initializeWebView];
-  }
-}
-#endif // !TARGET_OS_OSX
-
-- (void)didMoveToWindow
-{
-  if (self.window != nil && _webView == nil) {
-    [self initializeWebView];
   }
 
 #if !TARGET_OS_OSX


inkless:
I would second here, any idea when we could either revert it (merge this PR) or provide a proper fix?

mlazari:
Please consider this PR first: https://github.com/react-native-webview/react-native-webview/pull/3631, hopefully that would be a proper fix.

mlazari:
@0xFA11 Sorry, I didn't mean that my change is more correct, just wanted to let whoever reviews this PR to let them know about my change too, maybe I didn't use the right wording there. I'm ok with whatever the maintainers decide. I added there a repro project with 3 different branches to help with testing - one using version 13.12.2 (the one before this bug was introduced, but had the issue with the WebView not loading in background tab), one using the latest version 13.12.4 (loads the WebView in background, but javascript is disabled) and one with a patch with my changes on top of 13.12.4 (that doesn't have any of those 2 issues from my testing), and also added a video recording for each to help see the difference in behaviour.
I think the only difference between this PR and mine is that I left the initialization code extracted in a separate `initializeWebView` method and that method is also called from `setSource:` beside the original call in `didMoveToWindow` (which ensures the WebView is loaded when it's in a background tab, because `didMoveToWindow` is called only when the user switches to that tab). Also I left the body of `initializeWebView` wrapped in `@synchronized (self) { ... }`.

Titozzz:
Hey all, a little bit of context, I've been super busy with work these past months, and had less time to review webview stuff. The original PR was not merged by me but by another maintainer that did not measure the impact of the change. Thank y'all for pushing on those issue and staying civil, you are the best.

Regarding more general initialization issues: new arch introduced quite a few subtle changes and I wanted to rewrite a lot of things and drop support for old arch. I want to support preloading, keeping instances alive and more, but that takes time. In the meantime thanks for the help keeping it as functional as possible!

ladp333:
Hi @mlazari @0xFA11,

Apologies, I’m a bit unclear on the status of the “preload WebView on initial render in background tabs” issue. Are we planning to fix this? If so, do we have an estimated timeline?

Thanks!

0xFA11:
@ladp333 are you able to reproduce the issue on latest `v13.15.0`?

ladp333:
> @ladp333 are you able to reproduce the issue on latest `v13.15.0`?

@0xFA11 Thanks for your prompt reply. Yes, I can reproduce the issue. I’m using:
```
react-native-webview: "^13.15.0",
react-native": "0.74.5",
@react-navigation/bottom-tabs: ^6.6.1  (with lazy set to false)
```

I’m still on the old architecture. I’ve hacked my app to demonstrate the behavior and attached a video.

https://github.com/user-attachments/assets/99bac7f6-0403-408d-8422-ff57d89121ba

mlazari:
@ladp333 That is the default behaviour for TabNavigator. You need to set lazy: false if you want to render the screen on the on the initial render of the navigator rather than only after the first time it is accessed. https://reactnavigation.org/docs/bottom-tab-navigator/#lazy

ladp333:
@mlazari Thanks for your response. I have already done that and it didn't work. Hence my question. Have you ever tried the latest react-native-webview and @react-navigation/bottom-tabs 6.X (6.6.1) or 7.X with lazy set to false ? Did it work ?

mlazari:
@ladp333 Sorry, you're right, it currently does not load the WebView content when the screen is in a background tab (a tab that was not open yet). It passed some time and I forgot what happened here. I had a different PR https://github.com/react-native-webview/react-native-webview/pull/3631 that should have fixed the issue with the WebView loading with JS disabled, but looks like the mantainers decided it would be safer to just revert the change that was fixing the WebView not loading in a background tab (which is also fixing the disabled JS issue).

ladp333:
@mlazari Appreciate your reply. Btw, did your PR fix the preloading without causing any other issues? As you said, I found out yday that the PR which was merged and reverted added preloading in iOS but caused other issues unfortunately.

mlazari:
@ladp333 I don't think it was ever reviewed by someone else, but as far as I remember it was not causing other issues.

ladp333:
@0xFA11 Any thoughts on this issue?

Issue #3616, opened 2024-11-21
WebView displays 'JavaScript not enabled' error on first load (iOS only)


fabiovieira1901:
**Bug description:**  
When opening the WebView for the first time in the app on **iOS**, the following error is displayed:  
_"We're sorry but App doesn't work properly without JavaScript enabled. Please enable it to continue."_  
This happens despite having `javaScriptEnabled={true}` set in the WebView configuration.

**To Reproduce:**  
1. Create a React Native project using Expo with the following `package.json` configuration:  
   (see below)  

2. Implement the WebView component with the following code:  
   <WebView
       androidHardwareAccelerationDisabled={true}
       cacheEnabled={false}
       cacheMode={'LOAD_NO_CACHE'}
       incognito={true}
       javaScriptEnabled={true}
       bounces={false}
       automaticallyAdjustContentInsets={false}
       scalesPageToFit={false}
       source={{ uri: <url> }}
   />

3. Start the application on an iOS device or simulator and open the WebView for the first time. The error appears.
Expected behavior:
The WebView should load the content correctly on the first attempt on iOS without showing JavaScript-related errors.

Screenshot:
![36e2486f-49cf-480f-9779-48537ca5a8c5](https://github.com/user-attachments/assets/ce35c640-dc41-4640-9e5b-ee3c31f1c8f4)

Environment:
- OS: iOS
- OS version: Tested on iOS 16.2
- react-native version: 0.76.2
- react-native-webview version: 13.12.4
- Expo version: 52.0.9

package.json:
{
  "name": "name_test",
  "scripts": {
    "start": "expo start",
    "android": "expo start --android",
    "ios": "expo start --ios",
    "eject": "expo eject",
    "test": "node --inspect ./node_modules/jest/bin/jest.js --runInBand --watchAll",
    "lint": "eslint ."
  },
  "dependencies": {
    "@expo/react-native-action-sheet": "^4.1.0",
    "@react-native-async-storage/async-storage": "1.23.1",
    "@react-native-community/datetimepicker": "^8.2.0",
    "@react-native-community/netinfo": "11.4.1",
    "@react-native-masked-view/masked-view": "0.3.2",
    "@react-navigation/bottom-tabs": "^6.0.9",
    "@react-navigation/drawer": "^6.6.15",
    "@react-navigation/native": "^6.0.2",
    "@react-navigation/stack": "^6.1.9",
    "axios": "^0.19.2",
    "dotenv": "^16.4.5",
[...]

Issue #3624, opened 2024-11-26
Communicating between JS and Native not working on iOS 


javbk201:
**Bug description:**

**To Reproduce:**
1. Create a new app with expo or cli.
2. Install react-native-webview
3. install pods 
4. copy and paste the code below in App.tsx

```
import { StyleSheet, View } from 'react-native'
import React from 'react'
import WebView from 'react-native-webview'

const App = () => {
  const html = `
      <html>
      <head></head>
      <body>
        <script>
          setTimeout(function () {
            window.ReactNativeWebView.postMessage("Hello!")
          }, 2000)
        </script>
      </body>
      </html>
    `;
  return (
    <View style={styles.container}>
      <WebView
        source={{html}}
        onMessage={event => {
          alert(event.nativeEvent.data);
        }}
      />
    </View>
  );
};

export default App

const styles = StyleSheet.create({
  container: {
    flex: 1,
  },
})
```

**Expected behavior:**
This code should to show an Alert after 2 seconds. But in iOS does not show it.

**Environment:**
 - OS: Mac OS Sequoia
 - OS version: 15.1
 - react-native version: 0.76.2
 - react-native-webview version: 13.12.4
 - iOS Version: 18.0

mehmetsalihyaldiz:
same here

mlazari:
> If source={{ html }}, iOS is fine

@mehmetsalihyaldiz As I understand from the provided example, this is the problematic case actually - html source on iOS.

I see the same issue. My WebView is in a Modal. First time when I open the modal it doesn't work on iOS. The second and next times already works.

I wrote a small repro similar to my case here: https://snack.expo.dev/@mlazari/webview-repro?platform=ios

As you can see the first time I open the modal, the alert is not shown:
https://github.com/user-attachments/assets/ca70c88d-f8e1-42f7-b3bf-98fd5b383a25

mlazari:
Downgrading to 13.12.2 fixes the issue for me. It's strange that in my expo snack it's also 13.12.2, but maybe the Expo Go app that I have installed uses a newer version (I am not sure how dependency versions work with Expo Go).

mlazari:
This [patch](https://www.npmjs.com/package/patch-package) for 13.12.4 works for me too:
https://github.com/user-attachments/files/17937022/react-native-webview%2B13.12.4.patch
```
[...]

Issue #3612, opened 2024-11-20
React-native-webview in ios emulator first loads a local link on a white screen, then reloads it the second time?


xiaowen123456789:
**Bug description:**

  const localHtmlFile = {uri: 'http://172.25.10.39:3000/'};
  return (
    <View style={styles.container}>
      <SafeAreaWarper style={styles.webview}>
        <WebView
          ref={webViewRef}
          source={localHtmlFile}
          style={styles.webview}
          originWhitelist={['*']}
          scalesPageToFit={false}
          useWebKit={true}
          javaScriptEnabled={true}
          domStorageEnabled={true}
          allowFileAccess={true}
          allowUniversalAccessFromFileURLs={true}
          allowFileAccessFromFileURLs={true}
          allowingReadAccessToURL='*'
          mediaPlaybackRequiresUserAction={true}
          onHttpError={e => console.log(e)}
          onLoadProgress={e => console.log(e.nativeEvent, 'progress')}
          onError={e => console.log(e, 'error')}
          onMessage={e => console.log(e, 'message')}
        />
      </SafeAreaWarper>
    </View>
  );

**To Reproduce:**

**Expected behavior:**

**Screenshots/Videos:**

**Environment:**
 - OS:
 - OS version:15.1
 - react-native version:0.76.1
 - react-native-webview version:13.12.4

inkless:
Same here, seems relevant to https://github.com/react-native-webview/react-native-webview/issues/3578

RohanRajGautam:
I encountered the same issue and spent about an hour debugging it. After exploring various solutions, the most effective fix for now is to downgrade the version to:

`"react-native-webview": "13.12.2"`

It seems that version [13.12.3](https://github.com/react-native-webview/react-native-webview/pull/3559) contains some buggy code that requires attention from the maintainers. If you're facing similar problems, I recommend trying this version as a workaround until the issue is resolved in a future release.

mlazari:
Anyone willing to test a fix for this issue, please try these steps:
- replace `"react-native-webview": "13.12.4"` with `"react-native-webview": "github:react-native-webview/react-native-webview#573ce151683296f469de74fb21aa6de8c58d1dce"` in your package.json file
- install npm dependencies with `npm install` / `yarn install` (depending on which package manager you are using)
- reinstall pods (`pod install` in the `ios` folder) or if you are using expo run `npx expo prebuild`
[...]

Issue #3595, opened 2024-10-29
RN 0.76 iOS Javascript Injection Error


ChanghyeonYoon:
**Bug description:**

Only iOS, below error is appear.
```
 Error evaluating injectedJavaScript: This is possibly due to an unsupported return type. Try adding true to the end of your injectedJavaScript string. Error Domain=WKErrorDomain Code=4 "A JavaScript exception occurred" UserInfo={WKJavaScriptExceptionLineNumber=0, WKJavaScriptExceptionMessage=Cannot execute JavaScript in this document, WKJavaScriptExceptionColumnNumber=0, NSLocalizedDescription=A JavaScript exception occurred}
```

**To Reproduce:**

```
const injectedJavaScriptBeforeContentLoaded = `
    window.isWebView = true;
    window.systemLanguage = ${JSON.stringify(RNLocalize.getLocales()[0].languageCode)};
    window.appVersion = ${JSON.stringify(VersionCheck.getCurrentVersion())};
    true;
`;

<WebView
      injectedJavaScriptBeforeContentLoaded={injectedJavaScriptBeforeContentLoaded}
      bounces={false}
      onMessage={onMessageFromWeb}
      onError={errorWebView}
      onNavigationStateChange={(event) => {
        onNavigationStateChange?.(event);
        onChangeUrl(event);
      }}
      onLoadStart={onLoadStart}
      onLoadEnd={onLoadEnd}
      downloadingMessage={undefined}
      javaScriptCanOpenWindowsAutomatically
      allowsBackForwardNavigationGestures={enableSwipe}
      onOpenWindow={onOpenNewWindow}
      onContentProcessDidTerminate={onContentProcessDidTerminate}
      onRenderProcessGone={onRenderProcessGone}
      allowsInlineMediaPlayback
      onShouldStartLoadWithRequest={onShouldStartLoadWithRequest}
      renderToHardwareTextureAndroid
      webviewDebuggingEnabled={!BuildEnv.IS_PRODUCTION}
      sharedCookiesEnabled
      thirdPartyCookiesEnabled
      domStorageEnabled
      cacheEnabled
      javaScriptEnabled
      injectedJavaScriptBeforeContentLoaded={injectedJavaScriptBeforeContentLoaded}
      androidLayerType={'hardware'}
      decelerationRate={0.998}
      mixedContentMode={'compatibility'}
      textZoom={100}
/>
```

**Expected behavior:**

**Screenshots/Videos:**

**Environment:**
[...]

PR #3631, opened 2024-11-29
fix(iOS): WebView initialized with JavaScript disabled

mlazari:
The change in https://github.com/react-native-webview/react-native-webview/pull/3559 added WebView initialization to didMoveToSuperview, but that method seems to be called too early, before properties like `javascriptEnabled` are set from WebView props.

This change moves WebView initialization from didMoveToSuperview to `setSource` which is called from `updateProps:oldProps:` after all RNCWebViewImpl properties are set from WebView props.

Repro:

- with react-native-webview 13.12.2 - https://github.com/mlazari/WebViewRepro/tree/13.12.2 - WebView does not load in the background tab, it loads only when that tab becomes active, despite the `lazy` tab bar navigator option being set to false
https://github.com/user-attachments/assets/cc3a760a-58ff-4636-b47f-1eb6eef895fb

- with react-native-webview 13.12.4 - https://github.com/mlazari/WebViewRepro/tree/13.12.4 - the WebView loads in the background tab (because of the change in https://github.com/react-native-webview/react-native-webview/pull/3559), but with javascript disabled (it should be enabled by default and enabling it explicitly with `javascriptEnabled={true}` does not change anything)
https://github.com/user-attachments/assets/03ab24c8-b62d-4381-803b-aa4b47d93219

- with react-native-webview 13.12.4 plus this fix (applied with a patch) - https://github.com/mlazari/WebViewRepro/tree/13.12.4-fixed - the WebView load in the background tab with Javascript enabled as expected
https://github.com/user-attachments/assets/172ae56f-9d51-4c48-b9cc-65c817db8164

Fixes https://github.com/react-native-webview/react-native-webview/issues/3616, https://github.com/react-native-webview/react-native-webview/issues/3624 and probably others like https://github.com/react-native-webview/react-native-webview/issues/3595, https://github.com/react-native-webview/react-native-webview/issues/3578, https://github.com/react-native-webview/react-native-webview/issues/3612

### DIFF
--- a/apple/RNCWebViewImpl.m
+++ b/apple/RNCWebViewImpl.m
@@ -507,7 +507,7 @@ RCTAutoInsetsProtocol>
 - (void)initializeWebView
 {
   @synchronized (self) {
-    if (_webView != nil) {
+    if (self.superview == nil || _webView != nil) {
       return;
     }
     WKWebViewConfiguration *wkWebViewConfig = [self setUpWkWebViewConfig];
@@ -564,21 +564,9 @@ RCTAutoInsetsProtocol>
   }
 }
 
-// react-native-mac os does not support didMoveToSuperView
-#if !TARGET_OS_OSX
-- (void)didMoveToSuperview
-{
-  if (self.superview != nil && _webView == nil) {
-    [self initializeWebView];
-  }
-}
-#endif // !TARGET_OS_OSX
-
 - (void)didMoveToWindow
 {
-  if (self.window != nil && _webView == nil) {
-    [self initializeWebView];
-  }
+  [self initializeWebView];
 
 #if !TARGET_OS_OSX
   // Allow this object to recognize gestures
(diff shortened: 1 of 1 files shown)


apple/RNCWebViewImpl.m
```diff
@@ -564,21 +564,9 @@ - (void)initializeWebView
...
-  }
-}
-#endif // !TARGET_OS_OSX
-
 - (void)didMoveToWindow
 {
-  if (self.window != nil && _webView == nil) {
-    [self initializeWebView];
-  }
+  [self initializeWebView];
```
mlazari, line 569: I am not sure if this is needed or not with the call from `setSource:` below since from my testing that seems to be called first so here it doesn't have any effect. I thought it would be safer to leave it here too for now.

mlazari:
Closing this since the change that was causing the bug was reverted in #3615 (together with the change that preloads the WebView in a background tab).

yunyu:
Any chance of reopening or resubmitting this? The behavior here is "more correct" than the behavior in the current prod release

PR #3649, opened 2024-12-19
feat(ios): scrollIndicatorInsets

### LINKED ISSUE #817: Not possible to set scrollIndicatorInsets
As of, I believe, `6.2.1` it is now possible to correctly set `contentInset` on iOS. However, it is not possible to set `scrollIndicatorInsets`, which will quite often go hand in hand with the content insets.

It's also not possible to use `contentContainerStyle` to get around it that way.

Gregoirevda:
Adds support for `scrollIndicatorInsets` in iOS, which is optional and defaults to `UIEdgeInsetsZero`. Added a test in the example app which dynamically updates the content and scroll insets.

Fixes #817


Gregoirevda:
@Titozzz @jamonholmgren Could I please get a review

PR #3650, opened 2024-12-19
fix(ios): refresh control

Gregoirevda:
Assign refresh control to the correct property instead of adding a subview (which causes a jump when insets are set and the user pulls to refresh). 

PS: there's an issue with the reload function when source is inline html, it calls `webview.reload()` which only executes a new navigation (url is `about:blank`) and nothing will happen. It should call visitSource, but separating concerns and creating small PR's


Gregoirevda:
Ping @Titozzz please re open

PR #3693, opened 2025-01-31
chore(apple): fix some Xcode warnings caught internally

Saadnajmi:
At my side of Microsoft, we have recently "vendored" a couple of libraries. Which means, we started building them from source using our own generated Xcode project (not the one cocoapods generates), which has much stricter build flags / treats warnings as errors. In doing so, I caught a couple of issues in React Native Webview I'd like to push back up!

Namely:
- `RNCWebViewImpl` should implement the protocol `UIGestureRecognizerDelegate`. It already was, but it wasn't listed so at the header.
  - Without this,I got the error `Assigning to 'id<UIGestureRecognizerDelegate> _Nullable' from incompatible type 'RNCWebViewImpl *const __strong' ` on line 565 when we try to set `self` as a `UILongPressGestureRecognizer` delegate. 
- Move some macOS ifdefs around so that RNWWebViewImpl conforms to the `RCTAutoInsetsProtocol`
  - I need to look into more whether RCTView on macOS should maybe not conform to this protocol. It looks iOS specific. 
- Add some `__unused` cland macros to unused arguments     
- Fix a couple of "declaration of variable shadows a local variable" by either moving the variable declation or renaming the variable so it longer conflicts
- Fixed a couple of "setter doesn't set the autogenerated property" issues by renaming `_savedHideKeyboardAccessoryView` and `_savedKeyboardDisplayRequiresUserAction` to `_hideKeyboardAccessoryView` and `_keyboardDisplayRequiresUserAction`

### DIFF
--- a/apple/RNCWebViewImpl.h
+++ b/apple/RNCWebViewImpl.h
@@ -45,7 +45,7 @@ shouldStartLoadForRequest:(NSMutableDictionary<NSString *, id> *)request
 @end
 
 #if !TARGET_OS_OSX
-@interface RNCWebViewImpl : RCTView <UIEditMenuInteractionDelegate>
+@interface RNCWebViewImpl : RCTView <UIEditMenuInteractionDelegate, UIGestureRecognizerDelegate>
 
 @property (nonatomic, nullable) UIEditMenuInteraction *editMenuInteraction API_AVAILABLE(ios(16.0));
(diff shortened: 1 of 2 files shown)


Titozzz (approved):

Saadnajmi:
Thanks!

Issue #3600, opened 2024-11-06
[iOS] Expose filename on onFileDownload


roniemartinez:
**Is your feature request related to a problem? If so, Please describe.**

On desktop browsers, the file are saved with file name taken from the `Content-Disposition` header. This does not seem to be obtainable in the `onFileDownload` property. Please expose this as we cannot always predict the file type when the `downloadUrl` does not end with an file extension.

```
Content-Disposition: attachment; filename="myfile.xlsx"
```

**Describe the solutions you came up with**

Add the filename in the `FileDownloadEvent`

**Additional context**

roniemartinez:
It's understandable that no one can pick it up... but why auto-close a valid feature request?

zabojad:
@roniemartinez still no solution on this ?

zabojad:
@roniemartinez I've just created a PR that solves it if you are still looking for a solution...

roniemartinez:
Awesome! Thank you. Hope the maintainers will review your PR 👍

gaeldb:
PR has been closed, can we reopen with code from @zabojad ?

roniemartinez:
Just got back here after a long while. We did our own patch on the `react-native-webview` code by making sure that all attachments will download (`decisionHandler(WKNavigationResponsePolicyDownload)`). We based our implementation from https://dev.to/gualtierofr/download-files-in-a-wkwebview-boo
I honestly think `onFileDownload` is no longer needed.

PR #3701, opened 2025-02-06
[iOS] Extract filename from Content-Disposition header

zabojad:
…and expose it to `onFileDownload` callback.

This PR aims to solve issue #3600.


roniemartinez:
Just leaving a comment here to prevent stale :)

zabojad:
Hi there @Titozzz can this be reopened? And merged?

PR #3741, opened 2025-03-26
fix(android): onMessage exception on fallback path

Kudo:
# Why

fix a regression from #3609, not sure if it is the android 9 issue.
on the fallback onMessage path, the `mWebView.getUrl()` will throw exception because it's not on main thread.

# How

post onMessage execution on main thread

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
@@ -449,7 +449,8 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
         @JavascriptInterface
         public void postMessage(String message) {
             if (mWebView.getMessagingEnabled()) {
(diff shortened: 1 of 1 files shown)


Titozzz (approved):

dkahdwk:
Issue still occurs in version 13.13.5.
OnMssage function is not being called at all on Android device.

Kudo:
@dkahdwk happy to help investigation if you can elaborate the problem and share a repro.

dkahdwk:
@Kudo It's a project that I can't share, so I'll create a test project and share it with you when I have time. For now, I downgraded it to version 13.10.5 and solved it.

Kudo:
oh if the case is to pass non-string data in postMessage, i would recommend to transform arguments as string like i mentioned at https://github.com/react-native-webview/react-native-webview/pull/3609#issuecomment-2754947629.

PR #3834, opened 2025-08-22
Add sub resource SSL error handling on Android

coolsoftwaretyler:
This PR adds a new event handler to the `WebView` component for Android: `onLoadSubResourceError`, which is fired when the package receives SSL errors where the failing URL does not match the top window URL.

Currently, `react-native-webview` simply logs these out. But I had a use case where I needed to specifically handle these events. This patch works for my use case, and it seems reasonable other people might want something like this as well.

A simpler change might be to change `TopLoadingErrorEvent` to have some value indicating that `failingUrl !== topWindowUrl`, and just route everything through the existing handler. As a starting proposal, I chose to add a separate handler for this because I figured it was more backwards compatible. Happy to iterate on that if the project prefers a different design.

I also added a test case that points to `https://badssl.com/`. If you run `yarn android` to see the Example app and tap around, you'll see the logs coming through:

https://github.com/user-attachments/assets/ccf59104-fb83-4816-9ba3-b57e7f7afdc2


Titozzz (approved):
Good PR

PR #3845, opened 2025-09-05
Avoid exception by normalizing failingUrl on SSL error, and check topWindowUrl for null-ness.

MegaManSec:
Both failingUrl and topWindowUrl can be null, resulting in a null pointer exception.

Also fix the formatting of this code.


MegaManSec:
ping @Titozzz @jamonholmgren

PR #3884, opened 2025-11-11
Support new event constructor for new architecture

### LINKED ISSUE #3883: (Android) Replace deprecated constructors when creating webview events

- In Android, the constructor of the new `Event` class requires two parameters: `surfaceId` and `viewId`. The former represents the ID of the ReactSurfaceView, and the latter refers to the ID of the view that triggered the event.
- However, the current implementation defines the `Event` using the constructor that only takes the `viewId` parameter.
  + In this case, the surfaceId is automatically set to -1, which may cause issues where events are not properly received when using the new architecture.

- Use new constructor of the `Event` class

**Additional context**

- https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/Event.kt

l2hyunwoo:
## Issue

- close #3883

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebChromeClient.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebChromeClient.java
@@ -1,11 +1,9 @@
 package com.reactnativecommunity.webview;
 
 import android.Manifest;
-import android.annotation.TargetApi;
 import android.app.Activity;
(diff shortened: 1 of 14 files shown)


l2hyunwoo:
@Titozzz @jamonholmgren Hi, could you take a look at this PR when you get a chance?
Thanks!

Issue #3814, opened 2025-07-30
Unable to convert string to floating point value: "normal"


anigenero:
**Bug description:**
On Android, when using the WebView component with new architecture enabled, the `decelerationRate` property causes the component to error with the following:

```
This error is located at:
 ERROR  Warning: Error: Exception in HostFunction: Unable to convert string to floating point value: "normal"
```

**To Reproduce:**

Create a WebView with `decelerationRate` set to `normal`.

```typescript
<WebView decelerationRate="normal" />
```

In order to circumvent the error, you must use the numeric value. Simply removing the property does not resolve the issue either

```typescript
<WebView decelerationRate={0.998} />
```

**Expected behavior:**
The property value should not cause a conversion error on Android, since it's iOS specific. Assuming the error is from this line in `RNCWebViewManager:375`.

```java
   @Override
    public void setDecelerationRate(RNCWebViewWrapper view, double value) {}
```

If I can get to it, I will try to open a PR today for it.

**Environment:**
 - OS: Android 
 - OS version: 15-16
 - react-native version: `0.79.5`
 - react-native-webview version: `13.15.0`

Appsmen:
@anigenero Please let this be merged asap

anigenero:
> [@anigenero](https://github.com/anigenero) Please let this be merged asap

This requires a code review. We're waiting on acknowledgement of the issue and review of the code

thedivac:
This issue just cause me 3 hours of debugging, thanks for this issue, otherwise I would never have found the fix.

harsman:
This is the same issue as #3617 which was opened in November 2024 but was closed for some reason. If you don't want to hard code the deceleration rate you can just do:

`decelerationRate={Platform.OS === 'ios' ? 'normal' : undefined}`

Panda-ref:
Not stale

darkonetconomy:
This is still causing issues on Android

PR #3815, opened 2025-07-30
Android decelerationRate property fix

anigenero:
Fixes an issue where setting `decelerationRate` on Android fails because the property passes as a string and not a float. The underlying `RNCWebViewManager.java` accepts this property but it's not transformed in the Android React component.

```typescript
public void setDecelerationRate(RNCWebViewWrapper view, double value) {}
```

See https://github.com/react-native-webview/react-native-webview/issues/3814

PR #3885, opened 2025-11-12
fix decelerationRate property for Android

anigenero:
Fixes an issue where setting `decelerationRate` on Android fails because the property passes as a string and not a float. The underlying `RNCWebViewManager.java` accepts this property but it's not transformed in the Android React component.

```typescript
public void setDecelerationRate(RNCWebViewWrapper view, double value) {}
```

See https://github.com/react-native-webview/react-native-webview/issues/3814


anigenero:
Re-opening stale PR for https://github.com/react-native-webview/react-native-webview/pull/3815

johanholm (approved):

johanholm:
I just got this bug in my app and patched with this, worked fine after that.

anigenero:
This is not stale

anigenero:
@Titozzz @jamonholmgren

Issue #3894, opened 2025-11-24
[iOS] Crash in RNCWebView.mm when converting NSString to std::string (nullptr passed to basic_string)


MODORIAPPS:
**Bug description:**

On iOS, the app crashes natively when `react-native-webview` handles some navigation actions.

In `RNCWebView.mm`, some `NSString` values are converted to `std::string` without checking for `nil`. When a `nil` gets passed down as a `nullptr` into `std::basic_string`, `_LIBCPP_ASSERT_NON_NULL` aborts the app. Since this happens on the native side, the crash cannot be caught in JavaScript.

A patch-package workaround on our side prevents the crash. I can share the patch file if needed...!

**To Reproduce:**

I don’t have a minimal public repo yet, but the crash is reproducible in cases where `navigationAction` is missing some expected string fields.

In our app it happens on flows like:

- redirects
- `window.open`
- special/custom schemes (e.g. `about:blank`, `javascript:`)
- navigation events that fire before the document has fully loaded

Rough steps:

1. Load a WebView that triggers one of the flows above.
2. A navigation event fires before `onShouldStartLoadWithRequest`.
3. The app terminates with a native crash in `RNCWebView` (stack trace below).

A patch-package fix was applied locally and confirmed to prevent crashes. Patch file can be shared upon request.

**Expected behavior:**

`react-native-webview` should safely handle `nil` / missing `NSString` values and avoid passing `nullptr` into `std::string`, so the app does not terminate with a native abort.

**Screenshots/Videos:**

real error log below:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x0000000231e3e0cc libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x00000001e455c810 libsystem_pthread.dylib`pthread_kill + 268
    frame #2: 0x0000000193d8ff64 libsystem_c.dylib`abort + 124
  * frame #3: 0x0000000104b88c6c MyApp.debug.dylib`std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>::basic_string[abi:de200100]<0>(this="", __s=0x0000000000000000) at string:1083:5
    frame #4: 0x0000000104b88828 MyApp.debug.dylib`std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>::basic_string[abi:de200100]<0>(this="", __s=0x0000000000000000) at string:1082:87
[...]

PR #3917, opened 2026-01-30
fix(ios): prevent SIGABRT from nil NSString -> std::string conversion

MODORIAPPS:
On iOS, some navigation/event payload dictionaries may contain missing string fields. When an NSString is nil, UTF8String can yield a nullptr, and constructing std::string with it triggers _LIBCPP_ASSERT_NON_NULL, causing a native SIGABRT.

Introduce nullSafeString / nullSafeStringWithLength helpers and use them when building event payload structs to avoid passing nullptr into std::string.

> This could not be reliably mitigated at the JS layer, as the crash occurs on the native side, and was observed when loading certain legacy / non-standard websites.

Fixes react-native-webview/react-native-webview#3894

Changes:
- Add nullSafeString() with class check + UTF8String nullptr guard
- Add nullSafeStringWithLength() for the 2-arg std::string constructor
- Replace unsafe std::string([value UTF8String]) conversions across event payloads
- Remove redundant (?: "") fallbacks now covered by the helpers

Thanks for considering this contribution 🚀

### DIFF
--- a/apple/RNCWebView.mm
+++ b/apple/RNCWebView.mm
@@ -39,6 +39,26 @@ auto stringToOnLoadingFinishNavigationTypeEnum(std::string value) {
     return RNCWebViewEventEmitter::OnLoadingFinishNavigationType::Other;
 }
 
+static inline std::string nullSafeString(id value) {
+    if (![value isKindOfClass:[NSString class]]) {
+        return std::string();
+    }
+
+    const char *utf8String = [static_cast<NSString *>(value) UTF8String];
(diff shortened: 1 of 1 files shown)


wooller:
We are also experiencing this issue in our release binaries now and confirmed that this PR does fix the issue.

It would be great if there is any time for a review by the maintainers @Titozzz @jamonholmgren ❤️

Titozzz:
Will have a look !

Titozzz:
Thanks for the contribution

PR #3929, opened 2026-02-25
fix(android): fix injectedJavaScriptBeforeContentLoaded and injectedJavaScriptObject

### LINKED ISSUE #3326: injectedJavaScriptObject prop (sometimes) does not work on Android
**Bug description:** When the object from the `injectedJavaScriptObject` prop is to be injected into the page, [this condition](https://github.com/react-native-webview/react-native-webview/blob/4197bb42ce79406ff20c7e5637f78e44ca45474c/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java#L293) fails resulting in `window.ReactNativeWebView.injectedObjectJson()` returning `undefined` in the injected js.

**To Reproduce:** I've been unable to reproduce the issue with the example app, even when loading the same webpage the issue occurs with in my project. Unfortunately I also can't share my code. Can anyone think of a reason why this might happen only in my project? Is it possible the object is being injected before the page is ready currently? It seems to be a problem related to the order in which the native side loads the props.

**Expected behavior:** I expect the injected js to get the injected object from `window.ReactNativeWebView.injectedObjectJson()` instead of `undefined`.

**Environment:**
 - OS: macOS
 - OS version: 14.2.1
 - react-native version: 0.73.4
 - react-native-webview version: 13.8.1

Kudo:
# Why

Fixing two issues on Android:
- Unreliable `injectedJavaScriptBeforeContentLoaded` on Android. This is a known issue as #1609 and #1099.
- `injectedJavaScriptObject` doesn't work in some cases. fixes #3326.

# How

## injectedJavaScriptObject — Revamp bridge architecture

The old bridge architecture used `addWebMessageListener` (WML) on `"ReactNativeWebView"` for messaging and `addJavascriptInterface` on a separate name as fallback. `injectedObjectJson()` was injected via `evaluateJavascript` with backtick interpolation, which was fragile and not running synchronously.

The new design flips the relationship:

- `addJavascriptInterface` is **always** registered on `"ReactNativeWebView"`, making `injectedObjectJson()` a native `@JavascriptInterface` method — available before DOMContentLoaded, no JS aliasing needed.
- WML is registered on an internal name (`"__RNCWebViewMessageListener"`) for origin security.
- When WML is available, `window.ReactNativeWebView` is replaced with a wrapper (via `Object.assign`) that routes `postMessage` through WML while preserving all other properties including `injectedObjectJson()`.

| Android WebView feature | postMessage | injectedObjectJson | beforeContentLoaded |
|------------------------|------------|-------------------|-------------------|
| WML + DSS (modern) | WML (sourceOrigin) | RNCWebViewBridge | DSS |
| WML only (mid) | WML after override runs; @JavascriptInterface before that | RNCWebViewBridge | evaluateJavascript (unreliable) |
| Neither (old) | @JavascriptInterface | RNCWebViewBridge | evaluateJavascript (unreliable) |

## injectedJavaScriptBeforeContentLoaded — Document Start Scripts

Use `WebViewCompat.addDocumentStartJavaScript()` API to inject `injectedJavaScriptBeforeContentLoaded` before any page content loads. This is similar to `WKUserScriptInjectionTimeAtDocumentStart` on iOS but only available in newer WebView (when `WebViewFeature.isFeatureSupported(WebViewFeature.DOCUMENT_START_SCRIPT)` is available).

For older Android WebView (e.g. factory Android 9 WebView), DSS is not supported and `injectedJavaScriptBeforeContentLoaded` falls back to the existing `evaluateJavascript` path (which remains unreliable on those devices — a known platform limitation).

# Test Plan

A new **InjectedObjectJson** example has been added to the example app to verify `injectedObjectJson()` availability at both DOMContentLoaded and afterContentLoaded. Also verified the existing **Messaging** example to ensure `postMessage` continues to work correctly.

- **Modern Android (WML + DSS)**: Run the Injection and InjectedObjectJson examples
  - `injectedObjectJson()` should be available at both DOMContentLoaded and afterContentLoaded
  - `postMessage` should work (messaging example)
  - `beforeContentLoaded` injection should succeed (injection example)

- **Android 9 / old WebView (no WML, no DSS)**: Run same examples
  - `injectedObjectJson()` should work (RNCWebViewBridge, always available)
  - `postMessage` should work via `@JavascriptInterface`
  - `beforeContentLoaded` may be unreliable (known platform limitation)

---

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java
@@ -18,6 +18,7 @@ import android.webkit.WebViewClient;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.webkit.JavaScriptReplyProxy;
+import androidx.webkit.ScriptHandler;
 import androidx.webkit.WebMessageCompat;
 import androidx.webkit.WebViewCompat;
 import androidx.webkit.WebViewFeature;
@@ -59,6 +60,10 @@ public class RNCWebView extends WebView implements LifecycleEventListener {
     RNCWebViewBridge bridge;
     protected @Nullable
     WebViewCompat.WebMessageListener webMessageListener = null;
+    protected @Nullable
+    ScriptHandler documentStartScriptHandler = null;
+    private String lastDocumentStartScript = "";
+    protected boolean useDocumentStartScripts = false;
 
     /**
      * android.webkit.WebChromeClient fundamentally does not support JS injection into frames other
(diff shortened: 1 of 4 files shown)


kelset:
mmm looks like the CI is failing potentially because iPhone 15 is old (?)

but thanks @Kudo for working on this! it'd be great to get it in, would unlock a lot on my end :)

versten1uk:
Any updates here guys?

kewdex (approved):

caueobici:
Is there any updates here? This fixes a code injection security issue so i think it should be prioritized

seanleecoder:
Another downstream data point, since this PR fixes two bugs that turn out to be load-bearing *simultaneously* in the same code path.

**Context:** Expo SDK 57 DOM components (`"use dom"`) rendered through `react-native-webview` 13.16.1 (`useExpoDOMWebView: false`), on Android. This is the same failure as [expo/expo#47373](https://github.com/expo/expo/issues/47373), where @Kudo pointed at this PR, the reporter confirmed it resolved the crash, and the issue was then closed as third-party. So this PR is currently the *only* home for that bug — there is no open Expo-side issue tracking it.

The Android symptom is a hard crash on mount:

```
Error: Top OS ($$EXPO_DOM_HOST_OS) is not defined. This is a bug in the DOM Component runtime.
  at registerDOMComponent (expo/src/dom/dom-entry.tsx)
```

because `window.ReactNativeWebView.injectedObjectJson` is installed asynchronously via `evaluateJavascript` from `onPageStarted`, so it can land *after* the DOM bundle has already been evaluated. Near-certain in dev, where the bundle `<script>` is parser-blocking, and still reachable in production where it is `defer`red. iOS is unaffected — it installs the same helper as a document-start `WKUserScript`.

**What that costs downstream: two interlocking workarounds, one per bug this PR fixes.**

1. A package patch to `expo/src/dom/dom-entry.tsx` that polls for `injectedObjectJson` (16ms, 10s ceiling) and replays the bootstrap, instead of throwing. Turns an intermittent hard crash into a few frames of delay.
2. A deliberately **non-empty** `injectedJavaScriptBeforeContentLoaded`. `callInjectedJavaScriptBeforeContentLoaded()` no-ops on an empty string, and on Android it is the only thing that re-runs `injectJavascriptObject()` after navigating to the DOM page — i.e. the only way `injectedObjectJson` ever comes back at all. With an empty string, workaround (1) has nothing to wait for.

That second one is the part worth flagging for prioritisation: it means the `injectedJavaScriptBeforeContentLoaded` unreliability (#1609, #1099) and the `injectedJavaScriptObject` bug (#3326) are not independent in practice. Fixing either alone would leave this path broken, which is presumably why the revamped bridge architecture here — `addJavascriptInterface` always on `ReactNativeWebView`, making `injectedObjectJson()` a native `@JavascriptInterface` available before `DOMContentLoaded` — is the right shape rather than a targeted patch.

Happy to test this branch against a real Expo SDK 57 app with DOM components on Android and report back, if that would help move the review along.

Issue #3877, opened 2025-10-29
iOS Detox test stuck when WebView is opened


nlpulido:
**Bug description:**
On iOS, if a Detox test opens a screen with `react-native-webview`, it will hang indefinitely causing tests to timeout.

**To Reproduce:**
Write a simple Detox test that opens a screen with the WebView.

This started breaking from 12.0.0 onwards when the new arch support was introduced.

```
12:55:40.615 detox[1109] i The app is busy with the following tasks:
• Run loop "Main Run Loop" is awake.
• There are 2 work items pending on the dispatch queue: "Main Queue (<OS_dispatch_queue_main: com.apple.main-thread>)".
12:55:46.905 detox[1109] i The app has not responded to the network requests below:
  (id = 132) invoke: {"type":"action","action":"tap","predicate":{"type":"text","value":"Completed","isRegex":false}}
```

**Expected behavior:**
Detox test should not hang and test should continue.

**Screenshots/Videos:**
N/A

**Environment:**
 - OS: iOS (old architecture, new architecture is disabled).
 - OS version: 17.5
 - react-native version: 0.76.6
 - react-native-webview version: 13.16.0, fails from 12.0.0 onwards.

csath:
I am also experiencing the same. I noticed that setting `sharedCookiesEnabled` to true causes this detox hang. 

**Environment**:

OS: iOS (old architecture, new architecture is disabled).
react-native version: 0.79.5
react-native-webview version: 13.16.0/13.13.5

Should this issue be reported here or in Detox? 

PS: Found this already on detox > https://github.com/wix/Detox/issues/4438

LuanNguyen-HnL:
Any update on this? thanks

huszzsolt:
We have the same issue with detox 

`sharedCookiesEnabled={true}` makes detox hang / wait
`sharedCookiesEnabled={false}` detox is not working fine.

I suspect some threads keeps detox hanging.

Also we managed to get temporarily around with:
`  await device.detoxSynchronization();`, however this makes app testing flaky.

huszzsolt:
This is still an issue.

huszzsolt:
Hopefully https://github.com/react-native-webview/react-native-webview/pull/3934 provides a fix for iOS detox tests. It did for us.

huszzsolt:
I think this is still an issue

PR #3934, opened 2026-03-18
fix (iOS): sharedCookies bug when webview is nil. Race condition. Bug making Detox tests to wait indefinitely

huszzsolt:
# Issue

Detox identified the issue. Detox keeps waiting/synchronising with 

```
Run loop "Main Run Loop" is awake.
• There are 2 work items pending on the dispatch queue: "Main Queue (<OS_dispatch_queue_main: com.apple.main-thread>)".
12:55:46.905 detox[1109] i The app has not responded to the network requests below:
  (id = 132) invoke: {"type":"action","action":"tap","predicate":{"type":"text","value":"Completed","isRegex":false}}
```
when `sharedCookiesEnabled=true`

https://github.com/react-native-webview/react-native-webview/issues/3877

# Description 

This fix is based on some manual debugging and AI analysis. Here is the AI explanation:

What’s causing the Detox hang
Detox can hang because navigation is gated on the cookie-sync completion, and that completion can never fire.

In visitSource, the actual loadRequest happens only inside the completion passed to syncCookiesToWebView:

```
RNCWebViewImpl.m
Lines 861-884
  [self syncCookiesToWebView:^{
    // ...
    if (request.URL.host) {
      [webView loadRequest:request];
    }
    // ...
  }];
````

But syncCookiesToWebView calls writeCookiesToWebView, which (before the fix) did:
- capture `__weak WKWebView *webView = _webView;`
- `dispatch_async` to main
- `dispatch_group_enter(group)` for each cookie
- call `[webView.configuration.websiteDataStore.httpCookieStore setCookie:... completionHandler:^{ dispatch_group_leave }]`

If webView becomes nil by the time the async block runs (weak ref), setCookie is sent to nil, so the completion handler is never invoked, so the group never leaves, so the dispatch_group_notify never runs, so the navigation completion never runs → Detox waits forever.

# Note
- I am not familiar with the codebase and iOS webviews 
- I was wondering if this also causes some memory leaks/hangs in non-detox, production scenarios

### DIFF
--- a/apple/RNCWebViewImpl.m
+++ b/apple/RNCWebViewImpl.m
@@ -1872,7 +1872,9 @@ didFinishNavigation:(WKNavigation *)navigation
       if(!_incognito && !_cacheEnabled) {
         wkWebViewConfig.websiteDataStore = [WKWebsiteDataStore nonPersistentDataStore];
       }
-      [self syncCookiesToWebView:^{}];
+      if (_webView != nil) {
+        [self syncCookiesToWebView:^{}];
+      }
     } else {
       NSMutableString *script = [NSMutableString string];
 


huszzsolt:
I think this is still an issue @jamonholmgren . If needed, we I catch up with any maintainer at Apps.js next week?

PR #3939, opened 2026-03-24
fix(ios): call clearCache from webview implementation in new arch

tectonick:
It seems that after clearCache support was added for iOS in #3119, the new architecture file RNCWebView.mm wasn't updated to call the clearCache implementation in RNCWebViewImpl. It still performs a no-op with the android-only comment from before the change was introduced
Issue

After patch:
Fix

Post-fix implementation called


tectonick:
Hey @Titozzz, could you take a look when you have a moment? So far we are fine with a patch, but it'd be great to get this merged

Issue #2530, opened 2022-06-07
Getting user location fails when user only grants approximate location on Android


ayau:
**Bug description:**
onGeolocationPermissionsShowPrompt is currently hardcoded to request `ACCESS_FINE_LOCATION` only. Even if `<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />` is defined in the Manifest, Android still asks if the user wants to give Approximate (coarse) or precise (fine) location. If the user chooses approximate, the location request fails, saying user denied permission.

**To Reproduce:**
* Add `<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />` to Android Manifest.
* Set `geolocationEnabled={true}` on webview
* Trigger getting location in javascript with `window.navigator.geolocation.getCurrentPosition((position) => {})`
* Android will prompt user to select either approximate or precise location. Select approximate location
* `getCurrentPosition` fails with the error that user denied permission.

**Expected behavior:**
onGeolocationPermissionsShowPrompt should support both ACCESS_FINE_LOCATION and ACCESS_COARSE_LOCATION depending on which one is granted

**Screenshots/Videos:**
<img width="286" alt="image" src="https://user-images.githubusercontent.com/1381132/172445589-076791ba-0562-4382-8e45-3c3caeb021b5.png">

**Environment:**
 - OS: Android
 - OS version: Android 12 (API 31)
 - react-native version: 0.68.1
 - react-native-webview version: 11.18.1

liquidvisual:
Did you ever figure this out? I've been banging my head over this bug for years.

PR #3947, opened 2026-04-20
fix(android): handle ACCESS_COARSE_LOCATION Android permission

### LINKED ISSUE #2481: can't get geolocation with react-native-webview and android
**Bug description:**
Hello guys, thnaks for this great package.
I cant get geolocation with android 
and it is working fine with ios.

I've set :
`<WebView geolocationEnabled={true} .../>`

I'm using expo managed and I've added permissions 
```
"ACCESS_FINE_LOCATION",
"ACCESS_COARSE_LOCATION",
"FOREGROUND_SERVICE "
```

I've tested this with android physical device and also with android emulator, 
I've make sure the location is enabled .
I also tried to get permission with in the app before the webview is loaded, but not working 

**Environment:**
 - react-native version: 0.68.1
 - react-native-webview version: 11.18.1

Thanks

robwalkerco:
Fixes #2481, #2530 

Many apps legitimately only request `ACCESS_COARSE_LOCATION` for privacy reasons or to comply with Play Store policy, such as the [Financial Services Loans policy](https://support.google.com/googleplay/android-developer/answer/9876821#), so this PR updates the Android location permissions handling to support both `ACCESS_COARSE_LOCATION` and `ACCESS_FINE_LOCATION`. 

We now check if `ACCESS_FINE_LOCATION` is available in the permissions list, and request it, otherwise we fallback to `ACCESS_COARSE_LOCATION`. This fixes the scenario where the app does not declare `ACCESS_FINE_LOCATION` in the permissions manifest, either by choice or because of a Play Store policy.

When checking if the user already allows the permission request (in `webviewPermissionsListener`), we treat the user providing `ACCESS_COARSE_LOCATION` as a successful permission request (which it is!) - this fixes the scenario described in #2530

The WebView geolocation API doesn't distinguish between fine and coarse location - it simply needs some location permission to be granted. The native platform then returns location data with accuracy based on the permission level. By checking `ACCESS_FINE_LOCATION` first and falling back to `ACCESS_COARSE_LOCATION`, the code correctly handles both scenarios:

 - Apps that have declared precise location permissions
 - Apps that have opted for coarse-only location for privacy or policy compliance

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebChromeClient.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebChromeClient.java
@@ -195,18 +195,23 @@ public class RNCWebChromeClient extends WebChromeClient implements LifecycleEven
 
     @Override
     public void onGeolocationPermissionsShowPrompt(String origin, GeolocationPermissions.Callback callback) {
+        String permissionToRequest = null;
 
         if (ContextCompat.checkSelfPermission(this.mWebView.getThemedReactContext(), Manifest.permission.ACCESS_FINE_LOCATION)
                 != PackageManager.PERMISSION_GRANTED) {
+            permissionToRequest = Manifest.permission.ACCESS_FINE_LOCATION;
+        } else if (ContextCompat.checkSelfPermission(this.mWebView.getThemedReactContext(), Manifest.permission.ACCESS_COARSE_LOCATION)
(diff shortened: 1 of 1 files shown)


robwalkerco:
Not stale - just in need of a review from @Titozzz 🙇

Titozzz:
👋 I've started digging PRs ! Hold on

PR #3950, opened 2026-04-29
feat(android): add `profile` prop for isolated cookie jars

SMJ93:
## Summary

Adds a `profile` prop on Android that binds a WebView to an [androidx.webkit `Profile`](https://developer.android.com/reference/androidx/webkit/Profile), giving it an isolated cookie jar, web storage, geolocation permissions, and service workers - separate from the default profile shared by the rest of the app's WebViews.

Useful for any case where a WebView's cookies need to be independent from everything else. Two `<WebView>`s with different `profile` values can be logged into the same site as different users at the same time.

```jsx
<WebView source={{ uri: 'https://example.com' }} profile="profile-a" />
<WebView source={{ uri: 'https://example.com' }} profile="profile-b" />
```

## Behavior

- **Android with `MULTI_PROFILE` support** (androidx.webkit 1.9+, Android 10+ with a recent WebView): the named profile is created on first use via `ProfileStore.getOrCreateProfile` and attached with `WebViewCompat.setProfile`. Profiles persist across app launches.
- **Android without `MULTI_PROFILE` support**: the prop is ignored, the WebView falls back to the default profile, and a warning is logged via Logcat. No crash.
- **iOS**: the prop is typed as Android-only and is a no-op natively. The codegen field exists in the shared spec but is never read by the iOS view.

The Android Profile API does not allow reassigning a WebView's profile after it has loaded, so changing the prop value after mount is ignored with a Logcat warning. This is documented on the prop.

## Changes

- `android/.../RNCWebViewManagerImpl.kt` - `setProfile` with feature gate, get-or-create, and idempotency check
- `android/.../RNCWebViewManager.java` (newarch + oldarch) - `@ReactProp("profile")` wiring
- `src/RNCWebViewNativeComponent.ts`, `src/WebViewTypes.ts` - codegen + TS types (Android-only at the public API)
- `docs/Reference.md` - prop documentation
- `example/examples/Profile.tsx` + `example/App.tsx` - demo screen with multiple WebViews on different profiles, plus controls to set/get/clear cookies so you can verify isolation against `httpbin.org/cookies`

## Recording

https://github.com/user-attachments/assets/14274b7f-c474-4919-8456-8fa97297feee

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManagerImpl.kt
@@ -15,7 +15,6 @@ import android.webkit.CookieManager
 import android.webkit.DownloadListener
 import android.webkit.WebSettings
 import android.webkit.WebView
-import androidx.webkit.Profile
 import androidx.webkit.ProfileStore
 import androidx.webkit.WebSettingsCompat
 import androidx.webkit.WebViewCompat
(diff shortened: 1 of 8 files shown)


SMJ93:
Hey @Titozzz, let me know if you need to me to provide anything else 🙏

SMJ93:
Hey @Titozzz @jamonholmgren @MODORIAPPS, I can see you guys merged most recently. 

Please can you take a look 🙏

Issue #3942, opened 2026-04-08
Android injectedJavaScriptObject returns invalid JSON when payload contains escaped control characters


hoywu:
**Bug description:**

On Android, `injectedJavaScriptObject` can become invalid JSON when the injected payload contains escaped control characters such as `\n`, causing `JSON.parse(window.ReactNativeWebView.injectedObjectJson())` to fail with:

`SyntaxError: Bad control character in string literal in JSON`

This seems to happen because the Android implementation injects the already-stringified JSON into a JavaScript template literal:

https://github.com/react-native-webview/react-native-webview/blob/eb8ccacd35740af39993725ad1b592d45364a510/android/src/main/java/com/reactnativecommunity/webview/RNCWebView.java#L275-L283

If the JSON string contains escaped characters like `\n`, they are interpreted by the template literal and become real control characters before being returned to JavaScript. As a result, `injectedObjectJson()` no longer returns valid JSON.

**To Reproduce:**

A minimal reproduction should be possible with a payload containing a newline in a string value.

Example:

```tsx
import React from 'react';
import WebView from 'react-native-webview';

const html = `
<!doctype html>
<html>
  <body>
    <script>
      setTimeout(() => {
        try {
          const raw = window.ReactNativeWebView.injectedObjectJson();
          console.log('raw injectedObjectJson:', raw);
          const parsed = JSON.parse(raw);
          document.body.innerText = 'OK: ' + JSON.stringify(parsed);
        } catch (e) {
          document.body.innerText = 'ERROR: ' + e.message;
          console.error('Failed to parse injected object JSON', e, window.ReactNativeWebView.injectedObjectJson());
        }
      }, 0);
    </script>
  </body>
</html>
`;

export default function App() {
  return (
    <WebView
      source={{ html }}
      onMessage={() => {}}
      injectedJavaScriptBeforeContentLoaded={';'}
      injectedJavaScriptObject={{
        text: 'line1\nline2',
      }}
    />
  );
}
```

Steps:
1. Render a `WebView` on Android
2. Pass an object through `injectedJavaScriptObject`
3. Ensure one string field contains a newline, for example `'line1\nline2'`
4. In the page, call `window.ReactNativeWebView.injectedObjectJson()`
5. Run `JSON.parse(...)` on the returned value

Observed result:
- Android throws `SyntaxError: Bad control character in string literal in JSON`

**Expected behavior:**

[...]

PR #3958, opened 2026-05-31
fix(ios): avoid template literal in injectedObjectJson() to preserve escapes

### LINKED ISSUE #3957: iOS injectedJavaScriptObject corrupts JSON when value contains escaped quotes (template literal unescape)
**Bug description:**

On iOS, `injectedJavaScriptObject` can return invalid JSON when the injected payload contains a string value whose content is itself JSON-encoded (i.e. nested escaped double quotes such as `\"`). `JSON.parse(window.ReactNativeWebView.injectedObjectJson())` then throws:

```
SyntaxError: Expected '}'
```

The root cause is the same shape as #3942 (Android), but on a different platform: the iOS implementation injects the already-stringified JSON into a **JavaScript template literal**:

https://github.com/react-native-webview/react-native-webview/blob/eb8ccacd35740af39993725ad1b592d45364a510/apple/RNCWebViewImpl.m#L1721-L1740

```objc
// apple/RNCWebViewImpl.m
[NSString stringWithFormat:
 @"window.%@ = window.%@ || {};"
  "window.%@.injectedObjectJson = function () {"
  "  return `%@`;"          // <-- template literal
  "};", MessageHandlerName, MessageHandlerName, MessageHandlerName, source]
```

Template literals unescape `\"` to `"`. So a payload that contains a nested JSON string like `{"state":"{\"root\":...}"}` is injected as:

```js
return `{"state":"{\"root\":...}"}`;
```

…and the returned string ends up as `{"state":"{"root":...}"}` — broken JSON. The same problem also affects `\n` / `\r` / `\t` (as reported for Android in #3942).

**To Reproduce:**

A minimal reproduction:

```tsx
import React from 'react';

const html = `
<!doctype html>
<html>
  <body>
    <script>
      setTimeout(() => {
        try {
          const raw = window.ReactNativeWebView.injectedObjectJson();
          const parsed = JSON.parse(raw);
          document.body.innerText = 'OK: ' + JSON.stringify(parsed);
        } catch (e) {
          document.body.innerText = 'ERROR: ' + e.message;
        }
      }, 0);
    </script>
  </body>
</html>
`;

export default function App() {
  // The value `state` is itself a JSON string, which uses escaped quotes.
  // This is how editor frameworks (e.g. Lexical) typically serialize state.
  return (
    <WebView
      source={{ html }}
      onMessage={() => {}}
      injectedJavaScriptBeforeContentLoaded={';'}
      injectedJavaScriptObject={{
        state: JSON.stringify({ root: { children: [{ type: 'text', text: 'hello' }] } }),
      }}
    />
  );
}
```

Steps:
1. Render the above `WebView` on iOS.
2. Inspect the page text.

Observed: `ERROR: Expected '}'` (or a similar `JSON.parse` failure).

Expected: `OK: { ... }` — `injectedObjectJson()` should always return valid JSON for any payload that was accepted by `injectedJavaScriptObject`, including nested-JSON values and escaped control characters.

**Real-world impact:**

This breaks Lexical editor (and any other rich-text editor / framework that serializes state as JSON) when mounted inside an Expo DOM Component on iOS. The Expo DOM bootstrap (`expo/src/dom/dom-entry.tsx`) reads `JSON.parse(injectedObjectJson())` to recover `$$EXPO_DOM_HOST_OS` and `$$EXPO_INITIAL_PROPS`; when that throws, the React tree never mounts and the WebView renders blank.

**Environment:**

- OS: iOS (any version that supports `WKUserScript` — affects all iOS WebViews here)
- react-native-webview version: 13.16.1 (also present in master at `eb8ccac`)

**Possibly relevant implementation:**

`apple/RNCWebViewImpl.m` — `setInjectedJavaScriptObject:` builds the bootstrap script via `[NSString stringWithFormat:@"... return \`%@\`; ..."]`. The Android counterpart of this bug is #3942, and PR #3929 fixes Android by switching to `@JavascriptInterface` (which sidesteps the JS-string-escape problem entirely).

I'm opening a PR that fixes the iOS side with a minimal change — embedding the JSON as a JS expression rather than a template literal, and re-emitting it via `JSON.stringify(...)` so existing callers see no behavior change.

azuma317:
## Why

Fixes #3957 — the iOS counterpart of #3942.

On iOS, `setInjectedJavaScriptObject:` injects the already-stringified JSON into a JS **template literal**:

https://github.com/react-native-webview/react-native-webview/blob/eb8ccacd35740af39993725ad1b592d45364a510/apple/RNCWebViewImpl.m#L1721-L1740

```objc
"  return `%@`;"
```

Template literals unescape `\"` to `"`, so payloads that contain nested-JSON string values (e.g. `{ state: '{"root":...}' }`) come back from `injectedObjectJson()` with broken quoting. Consumers calling `JSON.parse(window.ReactNativeWebView.injectedObjectJson())` then throw `SyntaxError: Expected '}'`. Control characters like `\n` / `\r` / `\t` hit the same class of bug (as reported for Android in #3942).

This breaks, among other things, **Lexical editor mounted inside an Expo DOM Component on iOS**: `expo/src/dom/dom-entry.tsx` reads `JSON.parse(injectedObjectJson())` to recover `$$EXPO_DOM_HOST_OS` / `$$EXPO_INITIAL_PROPS`; when that throws, the React tree never mounts and the WebView renders blank.

## How

Minimal one-line change: embed the JSON as a **JS expression** (JSON is a subset of JS object-literal syntax) and re-emit it via `JSON.stringify(...)`:

```diff
-      "  return `%@`;"
+      "  return JSON.stringify(%@);"
```

Because the JSON is parsed by the JS engine as an object/array/string expression rather than scanned as a template literal, the escape handling matches what the JSON spec already guarantees. `JSON.stringify(...)` then re-serializes to the canonical JSON string form, so existing callers of `injectedObjectJson()` (which typically `JSON.parse` the result) keep working without any consumer-side change.

## Relationship to #3929

#3929 fixes Android for the same root cause by switching to `@JavascriptInterface` (a native bridge method) — that approach sidesteps JS string escaping entirely and is the more thorough fix. iOS has no exact equivalent without raising the deployment target to introduce `WKScriptMessageHandlerWithReply` (iOS 14+) and refactoring the bridge architecture, so this PR keeps the existing `WKUserScript` injection but plugs the escape bug with a minimal, reviewable change.

If a follow-up wants to mirror #3929's architecture on iOS, this PR is forward-compatible with that — and in the meantime restores correctness for users on the current code path.

## Test Plan

Manual repro from #3957: render a `WebView` whose `injectedJavaScriptObject` contains a nested-JSON string value, then `JSON.parse(window.ReactNativeWebView.injectedObjectJson())` from the page.

- **Before**: `SyntaxError: Expected '}'`
- **After**: parses successfully, value round-trips intact.

Also verified via the real-world Lexical-inside-Expo-DOM-Component case (Expo SDK 56, iOS 26.3 simulator): editor was rendering blank before this fix, mounts normally after.

### DIFF
--- a/apple/RNCWebViewImpl.m
+++ b/apple/RNCWebViewImpl.m
@@ -1721,6 +1721,12 @@ didFinishNavigation:(WKNavigation *)navigation
 - (void)setInjectedJavaScriptObject:(NSString *)source
 {
   _injectedJavaScriptObject = source;
+  // Embed `source` as a JS expression (JSON is a subset of JS object-literal
+  // syntax) rather than inside a template literal. Template literals unescape
+  // `\"` -> `"`, which corrupts payloads that contain already-stringified JSON
+  // (e.g. `{"state": "{\"k\":\"v\"}"}`). `JSON.stringify(...)` then re-emits a
+  // valid JSON string so existing callers of `injectedObjectJson()` (which
+  // typically `JSON.parse` the result) keep working unchanged.
   self.injectedObjectJsonScript = [
     [WKUserScript alloc]
     initWithSource: [
@@ -1728,7 +1734,7 @@ didFinishNavigation:(WKNavigation *)navigation
       stringWithFormat:
        @"window.%@ = window.%@ || {};"
       "window.%@.injectedObjectJson = function () {"
-      "  return `%@`;"
+      "  return JSON.stringify(%@);"
       "};", MessageHandlerName, MessageHandlerName, MessageHandlerName, source
     ]
     injectionTime:WKUserScriptInjectionTimeAtDocumentStart


azuma317:
Hi! Just a friendly follow-up on this PR.

This fixes the iOS escape-handling issue reported in #3957 and is the iOS counterpart to the Android-side work around #3942 / #3929.

I’ve verified the fix with:

- a nested JSON string passed through `injectedJavaScriptObject`
- `JSON.parse(window.ReactNativeWebView.injectedObjectJson())`
- the original Expo DOM Component + Lexical reproduction on iOS

The PR is intentionally small, but I’d also be happy to add a regression test or adjust the implementation if there are concerns about evaluating the serialized JSON as a JavaScript expression.

Thanks for maintaining the project!

PR #3962, opened 2026-06-09
fix(android): request CAMERA permission for capture file inputs

keyurgovrani:
## Summary

On Android, a capture-enabled file input silently does nothing when the host app **declares** the `CAMERA` permission but it **hasn't been granted** yet.

### Repro
- App's `AndroidManifest` declares `android.permission.CAMERA` (very common).
- CAMERA is not granted at runtime.
- Web content triggers `<input type="file" accept="image/*" capture>` (or `capture="environment"`).

**Result:** nothing opens. Logcat: `RNCWebViewModule: there is no Camera permission`.

### Root cause
`RNCWebViewModuleImpl.startPhotoPickerIntent` calls `needsCameraPermission()`, which returns `true` precisely when CAMERA is declared-but-not-granted. In that case the camera intent is **not built** (`photoIntent` stays `null`), so for a capture input `chooserIntent` becomes `null` and the picker never launches. The component *detects* it needs the permission but never *requests* it — unlike downloads, which already request `WRITE_EXTERNAL_STORAGE` via `grantFileDownloaderPermissions`.

### Fix
When a capture is requested and `needsCameraPermission()` is `true`, request `CAMERA` through the existing `PermissionAwareActivity` plumbing and launch the picker once the user responds:
- **Granted** → the capture intent is built and the camera opens.
- **Denied** → `needsCameraPermission()` stays `true`, the existing "no Camera permission" branch runs → **no re-request loop**.
- Host `Activity` not `PermissionAwareActivity` → falls back to the previous behaviour (logged, no crash).

The picker-launch body is extracted into `launchPhotoPickerIntent(...)` so it runs both directly and from the permission callback. Mirrors the existing `grantFileDownloaderPermissions` storage-permission flow; no JS API change, Android-only.

## Test Plan
Root cause confirmed on an Android emulator: with an app declaring CAMERA but not granted, tapping `<input capture>` is a no-op and logs `there is no Camera permission`; granting CAMERA out-of-band makes the same input open the camera capture.

With this change, on first capture tap:
1. **Before** — nothing happens; `there is no Camera permission` in logcat.
2. **After** — OS CAMERA prompt appears → **Allow** → camera capture opens. **Deny** → no crash, no loop. Already-granted → opens directly.

Non-capture file inputs and the download permission flow are unaffected.

## Changelog

[Android] [Fixed] - Request the CAMERA permission for `<input capture>` file inputs instead of silently dropping the camera intent

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModuleImpl.java
+++ b/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewModuleImpl.java
@@ -50,6 +50,7 @@ public class RNCWebViewModuleImpl implements ActivityEventListener {
     public static final int PICKER = 1;
     public static final int PICKER_LEGACY = 3;
     public static final int FILE_DOWNLOAD_PERMISSION_REQUEST = 1;
+    public static final int CAMERA_PERMISSION_REQUEST = 2;
 
     final private ReactApplicationContext mContext;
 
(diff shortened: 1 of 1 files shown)


Smankusors:
Isn't "does nothing" is the intended behavior according to this documentation?

[https://github.com/react-native-webview/react-native-webview/blob/master/docs/Guide.md#camera-option-availability-in-uploading-for-android](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Guide.md#camera-option-availability-in-uploading-for-android)

> Normally, apps that do not have permission to use the camera can prompt the user to use an external app so that the requesting app has no need for permission. However, Android has made a special exception for this around the camera to reduce confusion for users. If an app can request the camera permission because it has been declared, and the user has not granted the permission, it may not fire an intent that would use the camera (`MediaStore.ACTION_IMAGE_CAPTURE` or `MediaStore.ACTION_VIDEO_CAPTURE`). **In this scenario, it is up to the developer to request camera permission before a file upload directly using the camera is necessary.**

If the responsibility for requesting camera permission shifts from the app developer to the library, we should either update the documentation accordingly, or provide an opt-in/opt-out option so developers can choose whether the library should request camera permission automatically.

That said, this PR does make the behavior more consistent with Google Chrome on Android though. And it also supersedes the previous PR https://github.com/react-native-webview/react-native-webview/pull/944 and fixes https://github.com/react-native-webview/react-native-webview/issues/942.

PR #3971, opened 2026-06-28
fix(types): restore default generic for WebView

### LINKED ISSUE #3970: ### Bug: `index.d.ts` declares `WebView<P = undefined>`, causing TypeScript to reject all JSX props
When <WebView> is used in JSX without explicitly specifying the generic parameter P, it defaults to undefined. The props type then becomes WebViewProps & undefined, which TypeScript resolves to never.

This causes TypeScript to reject every JSX attribute passed to <WebView> — source, ref, onNavigationStateChange, onError, javaScriptEnabled, domStorageEnabled, sharedCookiesEnabled, cacheEnabled, bounces, etc. — with errors like:

No overload matches this call.
  Overload 2 of 2, '(props: never, context: any): WebView<undefined>', gave the following error.
    Type '{ uri: string; }' is not assignable to type 'never'.
    Type 'true' is not assignable to type 'never'.
    Type 'false' is not assignable to type 'never'.
To Reproduce

Create a React Native project with TypeScript.
Install react-native-webview@14.0.1 (or any version ≥ 13.17.0).
Try to use <WebView> with any props:

<WebView source={{ uri: 'https://example.com' }} javaScriptEnabled={true} />
Run tsc --noEmit — the compiler will report a type error for every prop.
Expected behavior

The component should accept props as defined by WebViewProps. The generic default should be P = {} so that WebViewProps & {} resolves to WebViewProps:

declare class WebView<P = {}> extends Component<WebViewProps & P> {
This was the declaration before commit b094f92 (PR #3931).

Environment

react-native-webview version: >= 13.17.0, 14.0.0, 14.0.1
React Native version: any
TypeScript version: any (confirmed on TS 5.x and 6.x)
Platform: any where the non-platform-specific index.d.ts is resolved (e.g., Linux, CI)
Additional context

The bug was introduced in commit b094f92 (PR #3931, "chore: tools update"), likely as an unintended side effect of the toolchain migration (Bun + oxfmt). Prior to that commit, index.d.ts used P = {}, which worked correctly. The platform-specific type definitions (WebView.ios.d.ts, WebView.android.d.ts) are unaffected because they use ForwardRefExoticComponent instead of the class declaration. ```

MinsikJeong:
Fixes #3970

## Summary

Restores the default generic parameter (`P = {}`) for the `WebView` class declaration in `index.d.ts`.

The current declaration causes `WebViewProps & undefined` to resolve to `never` when `WebView` is used in JSX without explicitly specifying `P`, making all props fail type checking.

This restores the behavior that existed before commit b094f92.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,7 +5,7 @@ export { FileDownload, WebViewMessageEvent, WebViewNavigation } from './lib/WebV
 
 export type WebViewProps = IOSWebViewProps & AndroidWebViewProps & WindowsWebViewProps;
 
-declare class WebView<P = undefined> extends Component<WebViewProps & P> {
+declare class WebView<P = {}> extends Component<WebViewProps & P> {
   /**
    * Go back one page in the webview's history.
    */


Titozzz:
Thanks solved in latest version (@next if you need it now)

PR #3973, opened 2026-07-02
feat(windows)!: New Architecture support + update all platforms to latest React Native

Titozzz:
## Summary

Updates every platform to its latest React Native version while keeping the `react-native-test-app` example harness, and brings Windows onto the New Architecture (Fabric).

### Version updates

| Package | Before | After |
|---|---|---|
| react-native (Android/iOS) | 0.73.5 | **0.86.0** |
| react | 18.2.0 | **19.2.7** |
| react-native-windows | 0.73.8 | **0.83.2** |
| react-native-macos | 0.73.17 | **0.81.8** |
| @callstack/react-native-visionos | 0.73.8 | **0.79.6** |
| react-native-test-app | 3.7.2 | **5.4.1** |

Per-platform ceilings come from react-native-test-app 5.4.1's support matrix (core 0.76–0.86, windows ≤0.83, macos ≤0.81, visionos ≤0.79). The platform forks are intentionally **not** in the checked-in `package.json` (kept aligned with core react-native only); each CI lane installs its own via the `add:macos` / `add:visionos` / `add:windows` scripts.

### Windows New Architecture

Integrates the library-side WebView2 Fabric implementation from #3952 (which builds on #3908), adapted to react-native-windows **0.83.2 stable**:

- `RCTWebView2ComponentView` hosts the WinUI 3 WebView2 control in a `XamlIsland` connected through `ContentIslandComponentView`
- Fabric component + `RNCWebViewModule` TurboModule registered from `ReactPackageProvider` (`RNW_NEW_ARCH` only)
- Hand-written codegen headers and a Win32 DLL project replace the old WinRT/UWP projects
- JS side registers `RCTWebView2` via `NativeComponentRegistry` (`src/RCTWebView2NativeComponent.windows.ts`, flagged Windows-only) and reuses the shared `NativeRNCWebViewModule` spec
- Fixes on top of #3952: removed the WinAppSDK override (RNW 0.83.2 ships a newer one), dropped the stale `packages.lock.json`, activated compile-time TurboModule spec validation, run the `alert()` `MessageBoxW` on a detached thread (it froze the app's UI thread), and dispose the `XamlIsland`/WebView2 on the component view's `Destroying` event (previously nothing was ever torn down)

**BREAKING (Windows):** the old-architecture Paper/XAML `WebView`/`WebView2` implementations are removed; Windows now requires New Architecture.

### Windows e2e tests (restored)

[...]

PR #3974, opened 2026-07-05
feat!: remove legacy architecture (Paper) support

### LINKED ISSUE #3814: Unable to convert string to floating point value: "normal"

### LINKED ISSUE #3883: (Android) Replace deprecated constructors when creating webview events

### LINKED ISSUE #3900: [Android] Deprecation warnings: Library uses legacy RCTEventEmitter instead of RCTModernEventEmitter (New Architecture)
Bug description: When building an Android app using React Native 0.81.1 (with New Architecture enabled), the Kotlin compiler emits multiple deprecation warnings regarding react-native-webview.

It appears the library is still relying on the legacy RCTEventEmitter interface instead of the RCTModernEventEmitter required for full Fabric/TurboModule compatibility. This suggests the library is running through the Interop Layer rather than being fully native to the New Architecture, causing build noise and potential performance overhead on event emission.

Build Log Output:

```Plaintext
w: .../node_modules/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/events/TopLoadingErrorEvent.kt:5:8 'interface RCTEventEmitter : JavaScriptModule' is deprecated. Use [RCTModernEventEmitter] instead.
w: .../node_modules/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/events/TopLoadingErrorEvent.kt:22:42 'interface RCTEventEmitter : JavaScriptModule' is deprecated. Use [RCTModernEventEmitter] instead.
w: .../node_modules/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/events/TopLoadingErrorEvent.kt:23:21 'fun receiveEvent(targetTag: Int, eventName: String, params: WritableMap?): Unit' is deprecated. Use [RCTModernEventEmitter.receiveEvent] instead.
(Note: This warning repeats for TopHttpErrorEvent, TopLoadingFinishEvent, TopMessageEvent, and other event classes).
```
To Reproduce:

Initialize a React Native 0.81.1 project (New Architecture enabled).

Install react-native-webview (latest version).

Run ./gradlew assembleRelease (or assembleDebug) inside the android folder.

Observe the Kotlin compilation warnings in the build output.

Expected behavior: The library should implement RCTModernEventEmitter to be fully compatible with the New Architecture (Fabric/Bridgeless) without relying on deprecated interfaces, ensuring a warning-free build and optimal performance.

Screenshots/Videos: N/A (See Build Logs above).

Environment:

OS: Linux (GitHub Actions Runner / CI)

OS version: Ubuntu Latest

react-native version: 0.81.1

react-native-webview version: 13.16.0

### LINKED ISSUE #3970: ### Bug: `index.d.ts` declares `WebView<P = undefined>`, causing TypeScript to reject all JSX props

### LINKED ISSUE #3977: [v14] TypeScript regression: every WebView prop errors with "not assignable to type 'never'"
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `react-native-webview@14.0.1` for the project I'm working on.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/react-native-webview/index.d.ts b/node_modules/react-native-webview/index.d.ts
index 10b9c14..626a67a 100644
--- a/node_modules/react-native-webview/index.d.ts
+++ b/node_modules/react-native-webview/index.d.ts
@@ -5,7 +5,7 @@ export { FileDownload, WebViewMessageEvent, WebViewNavigation } from './lib/WebV
 
 export type WebViewProps = IOSWebViewProps & AndroidWebViewProps & WindowsWebViewProps;
 
-declare class WebView<P = undefined> extends Component<WebViewProps & P> {
+declare class WebView<P = {}> extends Component<WebViewProps & P> {
   /**
    * Go back one page in the webview's history.
    */
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

Titozzz:
This PR removes all legacy (Paper) architecture support and makes the New Architecture (Fabric/TurboModules) mandatory. It also fixes several latent Fabric bugs that were previously masked by the Paper implementations and that would have become universal regressions with this change.

**BREAKING CHANGE:** the legacy architecture is no longer supported. Requires React Native 0.76+ (New Architecture enabled) and iOS 15.1+; react-native-macos 0.76+ on macOS (10.15+).

## Android

- Deleted the `android/src/oldarch/` and `android/src/newarch/` source sets and merged the per-architecture wrappers with the shared `*Impl` classes into single classes in `src/main`: `RNCWebViewManagerImpl.kt` + the Fabric manager became `RNCWebViewManager.kt` (Kotlin), and `RNCWebViewModuleImpl.java` + the TurboModule wrapper became `RNCWebViewModule.java`.
- `build.gradle` always applies the `com.facebook.react` plugin, drops the conditional source sets and the `IS_NEW_ARCHITECTURE_ENABLED` build-config field, and fails fast if `newArchEnabled=false` is set explicitly.
- `RNCWebViewPackage` registers the module as a TurboModule unconditionally.
- The `loadUrl` command now resets the progress-changed filter — previously only the Paper `receiveCommand` path did, so an intercepted navigation could leave progress events suppressed on Fabric.
- **Event system migrated off the legacy dispatch path** (fixes #3883, fixes #3900): all 11 event classes now use the `(surfaceId, viewId)` constructor and expose their payload via `getEventData()` instead of overriding the deprecated `dispatch(RCTEventEmitter)`; every dispatch site obtains the surface ID via `UIManagerHelper.getSurfaceId()`, and the `ScrollEvent`/`ContentSizeChangeEvent` call sites use the `surfaceId` overloads.

## iOS / macOS / visionOS

- Deleted the Paper `RNCWebViewManager.h/.mm` and the `RCTConvert+WKDataDetectorTypes` category it used; removed all `RCT_NEW_ARCH_ENABLED` guards. `RNCWebViewImpl` stays as the Fabric component's content view (same pattern RN core uses for ScrollView/InputAccessory).
- Teardown now always goes through `destroyWebView` via `prepareForRecycle`, which is compiled for macOS too (Paper's `removeFromSuperview` cleanup is gone), and it deregisters the `WKHTTPCookieStoreObserver` before releasing the `WKWebView`.
- Fabric prop/command parity fixes (previously handled only by the removed Paper manager):
  - `dataDetectorTypes` conversion read the wrong variable (always `none`) and applied at most one type; it now reads the incoming prop, OR-combines types like the old `RCT_MULTI_ENUM_CONVERTER`, and the ivar defaults to `phoneNumber` to match the codegen default.
  - `decelerationRate` ivar now defaults to `UIScrollViewDecelerationRateNormal` (was 0, which killed scroll momentum when the prop was omitted).
  - `automaticallyAdjustsScrollIndicatorInsets` added to the codegen spec and mapped (it existed only on Paper).
  - The `clearCache` command is forwarded to the implementation (was a no-op stub on Fabric).
- Version-guard cleanup for the new floors: SDK `#if __IPHONE_OS_VERSION_MAX_ALLOWED` guards became explicit `!TARGET_OS_OSX` guards (their only remaining effect), below-floor `@available` checks were unwrapped, and the dead pre-iOS-11 cookie-injection fallbacks were removed.
- Podspec simplified to a plain `install_modules_dependencies(s)`; minimum iOS raised to 15.1 and macOS to 10.15 (matching React Native 0.76 / react-native-macos 0.76).

## TypeScript / JavaScript

- The legacy `source` native prop is no longer passed (only the codegen `newSource`); deleted the dead Paper macOS component wrapper — macOS uses the shared codegen component.
- Android and macOS drop the iOS-only `decelerationRate` before the native component, so the `"normal"`/`"fast"` string shortcuts can no longer hit the Fabric `Double` conversion (fixes #3814).
- `messagingModuleName` and `onLoadingSubResourceError` are now optional in the codegen spec (Android-only; other platforms never provided them — an element-level type error previously masked by a `@ts-expect-error` on the legacy `source` prop).
- macOS now JSON-stringifies `injectedJavaScriptObject` like iOS/Android.
- `index.d.ts` no longer collapses `WebViewProps` to `never`: the `WebView<P = undefined>` generic default became `P = {}` again (with an explicit oxlint suppression so the toolchain can't reintroduce the regression from b094f92/#3931), since intersecting with `undefined` reduces to `never` under strictNullChecks (fixes #3970, fixes #3977).

## Examples, docs

- The macOS and visionOS example Podfiles opt into the New Architecture explicitly (react-native-test-app only forces it for react-native ≥ 0.82, and react-native-macos 0.81 / react-native-visionos 0.79 would otherwise fall back to Paper).
- README + Getting Started state the requirement and minimum versions; the `nativeConfig` reference and the custom-Android/iOS guides document that custom components receive the codegen props with the source delivered as `newSource` (`headers` as an array of `{ name, value }` entries).

Windows is untouched — it is already New-Architecture-only since #3973.

## Linked issues

- Fixes #3814 — `Unable to convert string to floating point value: "normal"` on Android with the New Architecture.
- Fixes #3883 — replace deprecated `Event(viewId)` constructors (surfaceId was always `-1`).
- Fixes #3900 — deprecation warnings from the legacy `RCTEventEmitter` event path.
- Fixes #3970 — `index.d.ts` props collapse to `never` (regression from the toolchain migration in #3931).
- Fixes #3977 — duplicate report of #3970 (patch-package workaround).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RXhcR2ZgZBVnaaatD3Ad56


Titozzz:
Thanks for working on this. I found two issues that should be fixed before merging:

1. **macOS teardown is missing.** `RNCWebViewImpl` no longer cleans itself up from `removeFromSuperview`, while `RNCWebView::prepareForRecycle` is compiled only when `TARGET_OS_OSX` is false. This leaves macOS without a path that calls `destroyWebView`, so recycled/unmounted WebViews may retain their KVO observer, script message handlers, media playback, and underlying `WKWebView`. Please add an equivalent macOS teardown/recycling hook.

2. **`dataDetectorTypes` is always converted to `WKDataDetectorTypeNone` on iOS.** In the Fabric prop conversion, the local `dataDetectorTypes` variable is initialized to `WKDataDetectorTypeNone` and then used as the input to every bit test. Those tests need to inspect `newViewProps.dataDetectorTypes` instead. Since this PR removes the Paper implementation that converted this prop correctly, the bug becomes an upgrade regression for all supported iOS users.

The removal of compatibility with React Native <= 0.70 through the unconditional `install_modules_dependencies` call is acceptable for this breaking release, so that does not need to block the PR. It would still be helpful to document the resulting minimum React Native version.

Titozzz:
Both issues fixed in 8d95f92:

1. **macOS teardown**: `prepareForRecycle` is now compiled for macOS as well — react-native-macos ships it in `RCTComponentViewProtocol` (verified against 0.81.8), so the `!TARGET_OS_OSX` guard was unnecessary and macOS now reaches `destroyWebView` on unmount/recycle.
2. **`dataDetectorTypes`**: the bit tests now inspect `newViewProps.dataDetectorTypes` instead of the local accumulator. While fixing it I also replaced the `else if` chain with independent `if`s so multiple detector types combine, matching the `RCT_MULTI_ENUM_CONVERTER` semantics of the removed Paper converter.

Also added the minimum version to the README and Getting Started guide: React Native 0.71+ with the new architecture enabled (`install_modules_dependencies` is the hard floor), react-native-macos 0.76+ on macOS.

---
_Generated by [Claude Code](https://claude.ai/code)_

Titozzz:
Thanks for addressing the previous findings. I found three more Fabric regressions that should be fixed before merging:

1. **The default `dataDetectorTypes="phoneNumber"` still resolves to `none`.** The conversion in `RNCWebView::updateProps` only runs when the old and new codegen props differ. The component's old props are initialized with the codegen default `phoneNumber`, while `RNCWebViewImpl` leaves `_dataDetectorTypes` initialized to zero. Consequently, omitting the prop—or explicitly passing `phoneNumber`—skips conversion and creates the `WKWebView` with `WKDataDetectorTypeNone`. Please initialize the implementation to `WKDataDetectorTypePhoneNumber` or apply the initial value unconditionally.

2. **`automaticallyAdjustsScrollIndicatorInsets` is lost when Paper is removed.** It remains a documented public iOS prop in `WebViewTypes.ts`, but is absent from the codegen `NativeProps` and has no mapping in `RNCWebView.mm`. Paper previously exported it directly, so after this PR setting the prop is silently ignored. Please add it to `NativeProps` and map it to `setAutomaticallyAdjustsScrollIndicatorInsets`.

3. **The default iOS deceleration rate becomes zero.** `decelerationRate` is an optional `Double` without a codegen default, and JS passes `undefined` when it is omitted. Both the default codegen value and the Objective-C ivar are therefore zero, so the initial prop comparison skips assignment. `scrollViewWillBeginDragging` later writes that zero value to the scroll view, causing scrolling to stop immediately instead of using the normal/fast rate. Paper supplied `UIScrollViewDecelerationRateNormal` when the prop was absent. Please initialize the native ivar or provide an explicit JS/codegen default.

Titozzz:
All three fixed in baf4f6d:

1. **Default `dataDetectorTypes`**: `RNCWebViewImpl` now initializes `_dataDetectorTypes = WKDataDetectorTypePhoneNumber`, matching the codegen default, so omitting the prop (or passing `phoneNumber` explicitly) creates the `WKWebView` with the documented default instead of `none`.
2. **`automaticallyAdjustsScrollIndicatorInsets`**: added to the codegen `NativeProps` (`WithDefault<boolean, false>`, matching the ivar's `NO` initialization) and mapped in `RNCWebView::updateProps` under the same iOS 13 guard as the implementation's setter, which is now declared in `RNCWebViewImpl.h`.
3. **`decelerationRate`**: `_decelerationRate` is now initialized to `UIScrollViewDecelerationRateNormal`, so omitting the prop keeps normal deceleration instead of writing `0` to the scroll view on the first drag.

Both new ivar defaults live in an `#if !TARGET_OS_OSX` block in `initWithFrame:` alongside the other prop defaults, keeping the general rule: with Fabric, `updateProps` only converts on change, so the implementation's initial state must equal the codegen defaults.

---
_Generated by [Claude Code](https://claude.ai/code)_

Titozzz:
Two more iOS lifecycle/API regressions to address:

1. **`clearCache()` becomes a no-op on iOS.** `WebView.ios.tsx` still exposes the imperative method and invokes the generated `clearCache` command, but `RNCWebView::clearCache` currently contains only an “android only” comment. The removed Paper manager previously forwarded this command to the existing `RNCWebViewImpl::clearCache:` implementation. Please make the Fabric command call `[_view clearCache:includeDiskFiles]`.

2. **Recycling leaves a dangling `WKHTTPCookieStoreObserver`.** Each loaded WebView registers `RNCWebViewImpl` on its cookie store in `visitSource`. The now-universal `destroyWebView` path clears `_webView` without first calling `removeObserver:self` on that store. The later `dealloc` cleanup accesses `self.webView`, which is already nil, so it cannot remove the old registration. Apple documents that `WKHTTPCookieStore` does not retain observers and requires callers to remove them before they become invalid, so this can leave callbacks targeting a recycled or deallocated object. Please remove the observer from `_webView.configuration.websiteDataStore.httpCookieStore` inside `destroyWebView`, before setting `_webView = nil`.

Titozzz:
Both fixed in 70df4d2:

1. **`clearCache`**: the Fabric command now forwards to `[_view clearCache:includeDiskFiles]` like the Paper manager did.
2. **Cookie store observer**: `destroyWebView` now calls `removeObserver:self` on `_webView.configuration.websiteDataStore.httpCookieStore` (same `@available(iOS 11.0, *)` guard as `dealloc`) before the KVO removal path drops `_webView`, so recycling no longer leaves a dangling `WKHTTPCookieStoreObserver` registration.

---
_Generated by [Claude Code](https://claude.ai/code)_

Titozzz:
I found two more consequences of making Fabric mandatory:

1. **Android must not forward the string form of `decelerationRate`.** `WebView.android.tsx` currently leaves this iOS-only prop in `...otherProps`, while the shared codegen spec declares it as `Double`. With `decelerationRate="normal"` (or `"fast"`), Fabric attempts the string-to-number conversion before reaching Android's no-op setter and throws `Unable to convert string to floating point value`. This is the existing failure reported in #3814, and this PR makes that Fabric path universal. Please destructure/drop `decelerationRate` on Android (or otherwise normalize it before the native component).

2. **Remove the legacy `source` contract from the custom-component documentation.** This PR intentionally stops passing `source={sourceResolved}` and keeps only the codegen `newSource` representation. The `nativeConfig` reference and custom WebView guides still promise that a custom native component can use the same JavaScript interface, which implies receiving `source`. Please remove/update that documentation and examples so they no longer advertise the removed `source` behavior, and document `newSource` if custom components remain supported.

Titozzz:
Both addressed in f763ef7:

1. **`decelerationRate` on Android**: the Android wrapper now explicitly passes `decelerationRate={undefined}` after the props spread, so the `"normal"`/`"fast"` string shortcuts arriving via untyped props can no longer reach the Fabric `Double` conversion (#3814). The macOS wrapper had the identical exposure (same shared spec, prop not destructured), so it got the same guard; iOS keeps its existing string→number normalization.
2. **Custom-component docs**: the `nativeConfig` reference and the custom-Android/iOS guides now state that custom components receive the codegen props, with the source delivered as the processed `newSource` prop (`headers` as an array of `{ name, value }` entries) and the legacy `source` prop no longer passed.

---
_Generated by [Claude Code](https://claude.ai/code)_

Titozzz:
One more New Architecture issue surfaced from the existing issue history:

**Android events still use the legacy, surface-less dispatch path.** All of the WebView event classes currently call the deprecated `Event(viewId)` constructor and override `dispatch(RCTEventEmitter)`. Under Fabric that constructor assigns `surfaceId = -1`, so events can be routed incorrectly and continue to depend on the interop/legacy emitter path that this PR is intended to leave behind.

This was already identified in #3883, and #3900 documents the resulting deprecation warnings and the risk once interop is disabled. Because this PR makes the New Architecture mandatory, it should migrate these event classes as part of the removal:

- construct events with both `surfaceId` and `viewId`;
- obtain the surface ID through `UIManagerHelper` at the dispatch sites;
- expose the payload through `getEventData()` instead of overriding the legacy `dispatch(RCTEventEmitter)`.

This applies to the loading, progress, HTTP/error, message, open-window, render-process, custom-menu, and should-start-load event classes—not only one callback.

Titozzz:
Done in 45c1fd6 — all 11 event classes (loading start/finish/error/progress, sub-resource error, HTTP error, message, open-window, render-process-gone, custom-menu, should-start-load) now:

- take `(surfaceId, viewId)` in the constructor instead of the deprecated `Event(viewId)`;
- expose the payload through `getEventData()` — the `dispatch(RCTEventEmitter)` overrides and the `RCTEventEmitter` imports are gone;
- every dispatch site (7 in `RNCWebViewClient`, 4 in `RNCWebView`, 2 in `RNCWebChromeClient`) obtains the surface ID via `UIManagerHelper.getSurfaceId()`.

I also migrated the two RN-core event call sites in `RNCWebView` (`ScrollEvent.obtain` and `ContentSizeChangeEvent`) to their `surfaceId` overloads, since they were on the same deprecated path. All the APIs used exist back to well before the documented RN 0.71 floor.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3976, opened 2026-07-10
fix(android): ignore iOS decelerationRate prop

### LINKED ISSUE #3935: decelerationRate crashes work when fast, normal is provided on Android
**Bug description:**
Currently on Android, if the `decelerationRate` is set to one of the constants like fast or normal it causes the application to crash.

Set `decelerationRate="fast"`.

**Expected behavior:**
It should be silently ignored, as other iOS-only props do.

**Screenshots/Videos:**

Image

**Environment:**

- OS: Android
- OS version: 16
- react-native version: 0.81.5
- react-native-webview version: 13.15.0

MarshallBear1:
## Summary

- Stop the Android wrapper from forwarding the iOS-only `decelerationRate` prop to the native component.
- Prevent the generated Fabric delegate from casting the string aliases `fast` and `normal` to `Double`.
- Add Android regression coverage for both string aliases and a numeric value while checking that unrelated native props are preserved.

Fixes #3935

## Why

The public cross-platform WebView props accept the iOS `decelerationRate` values, and the Android wrapper previously included the value in its rest-prop spread. Under Fabric, the generated delegate recognizes the native spec field as a `Double` and performs that cast before the Android no-op setter can ignore it. Filtering the iOS-only prop at the Android JavaScript boundary makes the behavior match the expected platform-specific no-op.

## Test plan

- `bun run lint:ci`
- `bun run jest src --runInBand` (3 suites, 17 tests)
- `bun run prepare`
- `bun pm pack --dry-run`
- `./gradlew --no-daemon clean build check test` from `example/android` (356 tasks, including debug, debugOptimized, release, codegen, CMake, lint, and tests)
- Android API 34 emulator using the new architecture (`fabric: true`) with the exact `decelerationRate="fast"` repro: the WebView rendered and logcat contained no `ClassCastException`, fatal exception, or Fabric property-update error.

### DIFF
--- a/src/WebView.android.tsx
+++ b/src/WebView.android.tsx
@@ -25,6 +25,7 @@ import {
 import {
   AndroidWebViewProps,
   WebViewSourceUri,
+  type IOSWebViewProps,
   type WebViewMessageEvent,
   type ShouldStartLoadRequestEvent,
 } from './WebViewTypes';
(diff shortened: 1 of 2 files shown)


Titozzz:
fixed on latest (@next on npm for now)

Issue #3980, opened 2026-07-11
Windows (New Architecture): production feedback on the WebView2 implementation + offer of hardening PRs


FaithfulAudio:
We ship a production react-native-windows 0.83.2 New-Architecture app (Win32 composition, WinAppSDK 1.8, ARM64) with an independently-built WebView2 Fabric component that converged on exactly the architecture #3973 landed: WinUI 3 `WebView2` → `Microsoft.UI.Xaml.XamlIsland` → `ContentIsland` → RNW `ContentIslandComponentView.Connect()`. It renders real web content in the field today, so first: the XamlIsland/ContentIsland approach in #3973 is solid — we can confirm first pixel, load events, postMessage round-trips, and teardown behavior (including the "orphaned island breaks window-wide input" trap your comment in `RCTWebView2ComponentView.cpp` also documents) on independent code.

Reading `windows/ReactNativeWebView/RCTWebView2ComponentView.cpp` against our production component, we found a handful of places where our implementation had to harden the same code paths, and we'd like to upstream them as a small PR series.

1. **No `onError` when the WebView2 Runtime is missing.** `EnsureCoreWebView2Async()` is called without an environment and without failure handling (`InitializeContentIsland`). On machines without the Evergreen Runtime (stripped/managed images, Server SKUs) the view stays blank with no JS-visible signal. We create the environment explicitly (`CoreWebView2Environment::CreateWithOptionsAsync`) with a writable, packaged-safe user-data folder, catch failures, and emit `onLoadingError` with the HRESULT — both when environment creation itself throws (Runtime entirely missing) and via `CoreWebView2InitializedEventArgs.Exception()` (failure after an environment was created).
2. **Failed navigations emit `onLoadingFinish`.** `OnNavigationCompleted` doesn't consult `args.IsSuccess()`/`WebErrorStatus()`, so DNS failures, TLS errors, and offline states look like successful loads to JS. We branch and emit `onLoadingError` with the `CoreWebView2WebErrorStatus` instead. While in there: `title` (`DocumentTitle()`) already exists as a field on every nav-event struct in the codegen (`RCTWebView2.g.h`) but is never populated on Windows — we fill it in on `onLoadingStart`/`onLoadingFinish`/`onLoadingError`/`onSourceChanged`.
[...]

PR #3982, opened 2026-07-11
fix(windows): honor navigation success/failure + include title in nav events

FaithfulAudio:
Part of the hardening series offered in #3980. This is **WV-2**: honor navigation success/failure, and populate `title` in nav events.

### Problem

`OnNavigationCompleted` never reads `args.IsSuccess()` / `args.WebErrorStatus()` — it unconditionally dispatches `onLoadingFinish`. Separately, every nav-event struct in the codegen (`RCTWebView2.g.h`) already has a `title` field, but nothing on Windows ever sets it.

### Real-world failure mode

A DNS failure, TLS error, or the device going offline mid-navigation all report as `onLoadingFinish` to JS — a page that failed to load looks identical to one that succeeded. Apps that branch on `onLoadingError` to show a retry/offline UI (a common pattern, and one this monorepo's own offline-first FIT app relies on) never see it fire on Windows. Apps that display `nativeEvent.title` (e.g. for a browser-chrome-style header) get an empty string forever on Windows.

### Fix

- `OnNavigationCompleted` branches on `args.IsSuccess()`. On failure it emits `onLoadingError` with `event.code` set from `args.WebErrorStatus()` (cast to `int32_t`) instead of `onLoadingFinish`, and skips the message-bridge injection that follows.
- `CurrentTitle()` (new, best-effort, empty string if `CoreWebView2` isn't ready or the read throws) is wired into `onLoadingStart`, `onLoadingFinish`, `onLoadingError`, and `onSourceChanged`.

### Validation status

Both behaviors — checking `IsSuccess()`/`WebErrorStatus()` on nav-completed, and populating `title` from `DocumentTitle()` — are production-proven in a parallel, independently-built Windows new-arch WebView2 Fabric component running in Facilitron FIT (react-native-windows 0.83.2, WinAppSDK 1.8, ARM64) — see the production twin: [FacilitronWorks/react-native-webview-windows](https://github.com/FacilitronWorks/react-native-webview-windows). Renamed symbols, hand-ported to this repo's codegen/class shape — not a copy-paste.

Related: #3980

### DIFF
--- a/windows/ReactNativeWebView/RCTWebView2ComponentView.cpp
+++ b/windows/ReactNativeWebView/RCTWebView2ComponentView.cpp
@@ -319,9 +319,19 @@ bool RCTWebView2ComponentView::Is17763OrHigher() {
     return hasUniversalAPIContract_v7.value();
 }
 
+std::string RCTWebView2ComponentView::CurrentTitle() const noexcept {
+    try {
+        if (m_webView && m_webView.CoreWebView2()) {
+            return winrt::to_string(m_webView.CoreWebView2().DocumentTitle());
+        }
+    } catch (...) {
+    }
+    return {};
+}
+
 void RCTWebView2ComponentView::OnNavigationStarting(
     winrt::Microsoft::Web::WebView2::Core::CoreWebView2NavigationStartingEventArgs const& /*args*/) {
-    
+
     try {
         if (auto eventEmitter = EventEmitter()) {
             RNCWebViewCodegen::RCTWebView2EventEmitter::OnLoadingStart event;
@@ -329,6 +339,7 @@ void RCTWebView2ComponentView::OnNavigationStarting(
                 event.url = winrt::to_string(m_webView.Source().AbsoluteCanonicalUri());
             }
             event.loading = true;
+            event.title = CurrentTitle();
             event.canGoBack = m_webView ? m_webView.CanGoBack() : false;
             event.canGoForward = m_webView ? m_webView.CanGoForward() : false;
             event.navigationType = "other";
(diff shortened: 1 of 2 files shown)


windows/ReactNativeWebView/RCTWebView2ComponentView.cpp
```diff
@@ -319,16 +319,27 @@ bool RCTWebView2ComponentView::Is17763OrHigher() {
...
-    
+
     try {
         if (auto eventEmitter = EventEmitter()) {
             RNCWebViewCodegen::RCTWebView2EventEmitter::OnLoadingStart event;
             if (m_webView && m_webView.Source()) {
                 event.url = winrt::to_string(m_webView.Source().AbsoluteCanonicalUri());
             }
             event.loading = true;
+            event.title = CurrentTitle();
```
acoates-ms: This is an extra string copy, right?
Can we have CurrentTitle return the hstring, then do the to_string here? - I think that avoids the extra string copy.

FaithfulAudio:
Good catch — done. `CurrentTitle()` now returns the `hstring` straight from `CoreWebView2::DocumentTitle()`, and the four call sites do `winrt::to_string()` where they assign to the `std::string` event field.

One honest correction on the saving, since I'd rather you have the real number: on today's four call sites — all of which assign to a `std::string` — it's the same single allocation either way. The old version returned by value, so that was a move, not a copy. What the change actually buys is the right shape for a WinRT accessor: `hstring` is a refcounted handle, so any future caller that only needs to compare, test-empty or forward the title pays no allocation at all, and the conversion happens at the point of use rather than unconditionally inside the getter. Worth doing for that reason — just not a copy I can claim to have removed today.

PR #3984, opened 2026-07-11
fix(windows): fix javaScriptEnabled lost before core-init; stop re-navigating on unrelated prop changes

FaithfulAudio:
Part of the hardening series offered in #3980. This is **WV-4**: fix `javaScriptEnabled` lost before core-init, and stop re-navigating on unrelated prop changes.

### Problem

Two prop-lifecycle bugs in `UpdateProps`:

- `javaScriptEnabled` is only applied when `m_webView.CoreWebView2()` already exists (`if (m_webView.CoreWebView2()) { ...IsScriptEnabled(...); }`). `OnCoreWebView2Initialized` only reapplies `userAgent` and pending HTML once the core becomes ready — never `javaScriptEnabled`.
- `newSource.uri` / `newSource.html` are applied unconditionally whenever present. `UpdateProps` receives the **full current prop set** on every call (not a diff of what changed), so `m_webView.Source(uri)` / `NavigateToString(html)` re-runs on *any* prop change that happens to co-occur with a still-set source — not just an actual source change.

### Real-world failure mode

- `<WebView javaScriptEnabled={false} source={{uri}} />` set on first render: the prop is captured by `UpdateProps` before `CoreWebView2` exists, the branch is a no-op, and `OnCoreWebView2Initialized` never revisits it — JS ends up enabled despite the prop, silently.
- Any subsequent prop change on the component (toggling `javaScriptEnabled`, changing `userAgent`, anything that triggers a Fabric prop update) re-sets `Source(uri)` with the *same* URI it already has — a real navigation/reload as a side effect of an unrelated prop change. This is a plausible perf/UX regression for any app that re-renders a `<WebView>` with other changing props while `source` stays constant.

### Fix

- `javaScriptEnabled` is now stored on the instance (`m_javaScriptEnabled`) and (re)applied both in `UpdateProps` (when the core already exists) and in `OnCoreWebView2Initialized` (covering the case where `UpdateProps` ran first).
- `newSource.uri` / `newSource.html` are now diffed against `oldProps` (already an `UpdateProps` parameter, previously unused) — `Source()`/`NavigateToString()` only run when the value actually changed. `oldProps` is null on first mount, so the initial navigation is unaffected.

### Question for maintainers

We didn't confirm from source whether WinUI3's `WebView2.Source` setter itself no-ops when set to an unchanged URI (i.e. whether this is a double-fix or the actual fix). Either way the `NavigateToString`/html path had the identical unconditional-re-navigate structure, so the diffing is applied uniformly; if `Source()` already no-ops internally, this PR is still a correctness improvement for the html path and cheap insurance for the uri path.

### Validation status

Both the `javaScriptEnabled` lifecycle fix and the applied-source diffing (there tracked via `appliedUri`/`appliedHtml` rather than a raw `oldProps` diff, but same effect) are production-proven in a parallel, independently-built Windows new-arch WebView2 Fabric component running in Facilitron FIT (react-native-windows 0.83.2, WinAppSDK 1.8, ARM64) — see the production twin: [FacilitronWorks/react-native-webview-windows](https://github.com/FacilitronWorks/react-native-webview-windows). Renamed symbols, hand-ported to this repo's codegen/class shape — not a copy-paste.

Related: #3980

### DIFF
--- a/windows/ReactNativeWebView/RCTWebView2ComponentView.cpp
+++ b/windows/ReactNativeWebView/RCTWebView2ComponentView.cpp
@@ -185,35 +185,58 @@ void RCTWebView2ComponentView::UpdateProps(
         }
     }
     
-    // Handle source navigation
+    // Handle source navigation. UpdateProps is called with the FULL current
+    // prop set whenever ANY prop changes (not just source-related ones), so
+    // re-setting Source() / calling NavigateToString() unconditionally here
+    // re-navigates -- for Source() specifically, possibly reloading the
+    // page -- on every unrelated prop change (e.g. toggling
+    // javaScriptEnabled). Diff against oldProps so navigation only happens
+    // when the source actually changed; oldProps is null on first mount, in
+    // which case we always want to navigate.
+    const bool uriChanged = !oldProps ||
+        oldProps->newSource.uri.value_or(std::string{}) != newProps->newSource.uri.value_or(std::string{});
+    const bool htmlChanged = !oldProps ||
+        oldProps->newSource.html.value_or(std::string{}) != newProps->newSource.html.value_or(std::string{});
+
     if (newProps->newSource.uri.has_value() && !newProps->newSource.uri.value().empty()) {
-        try {
-            auto uri = winrt::Windows::Foundation::Uri(winrt::to_hstring(newProps->newSource.uri.value()));
-            m_webView.Source(uri);
-        } catch (...) {
-            // Invalid URI
+        if (uriChanged) {
+            try {
+                auto uri = winrt::Windows::Foundation::Uri(winrt::to_hstring(newProps->newSource.uri.value()));
+                m_webView.Source(uri);
+            } catch (...) {
+                // Invalid URI
+            }
         }
     } else if (newProps->newSource.html.has_value() && !newProps->newSource.html.value().empty()) {
-        if (m_webView.CoreWebView2()) {
-            try {
-                m_webView.NavigateToString(winrt::to_hstring(newProps->newSource.html.value()));
-            } catch (...) {
-                // Navigation failed
+        if (htmlChanged) {
+            if (m_webView.CoreWebView2()) {
+                try {
+                    m_webView.NavigateToString(winrt::to_hstring(newProps->newSource.html.value()));
+                } catch (...) {
+                    // Navigation failed
+                }
+            } else {
+                // CoreWebView2 not ready yet - save HTML for later navigation in OnCoreWebView2Initialized
+                m_pendingHtml = newProps->newSource.html.value();
             }
-        } else {
-            // CoreWebView2 not ready yet - save HTML for later navigation in OnCoreWebView2Initialized
-            m_pendingHtml = newProps->newSource.html.value();
         }
     }
-    
+
     // Apply debugging enabled
     if (newProps->webviewDebuggingEnabled.has_value() && m_webView.CoreWebView2()) {
         m_webView.CoreWebView2().Settings().AreDevToolsEnabled(newProps->webviewDebuggingEnabled.value());
     }
-    
-    // Apply JavaScript enabled
+
+    // Apply JavaScript enabled. Stored on the instance (m_javaScriptEnabled)
+    // so OnCoreWebView2Initialized can (re)apply it once CoreWebView2 exists
+    // -- previously, javaScriptEnabled={false} set before CoreWebView2 was
+    // ready (e.g. on first render) was silently dropped: this branch was a
+    // no-op because CoreWebView2() was still null, and
+    // OnCoreWebView2Initialized never reapplied it (only userAgent and
+    // pending HTML were reapplied there).
+    m_javaScriptEnabled = newProps->javaScriptEnabled;
     if (m_webView.CoreWebView2()) {
-        m_webView.CoreWebView2().Settings().IsScriptEnabled(newProps->javaScriptEnabled);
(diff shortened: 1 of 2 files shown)


manage-some:
Testing `@facilitronworks/react-native-webview-windows@0.1.0-alpha.0` on RNW 0.84.1 (Fabric/Composition).

**What we've confirmed so far:**
- ✅ Builds clean on x64 Debug with VS 2026
- ✅ Explicit user-data folder (PR #3981) fixes the black-screen-on-subsequent-launch bug present in upstream v16
- ✅ URL loads successfully in the WebView
- ✅ `prewindows` cleanup of orphaned `msedgewebview2.exe` processes works

Still to validate: `onError`, message bridge, and `injectedJavaScriptBeforeContentLoaded` — will update once we wire those up.

FaithfulAudio:
Thanks for running this — really useful to have independent confirmation on 0.84.1/Fabric, especially that the explicit user-data folder (#3981) clears the black-screen-on-relaunch you were seeing on upstream v16.

A few pointers for the remaining items, mapped to the PR that carries each fix:

- **`onError`**: #3981 surfaces CoreWebView2 environment/controller init failures to JS, and #3982 makes navigation failures actually fire the failure path (upstream reported success unconditionally). Easiest repro for the nav case: point `source.uri` at an unreachable host and check you get the failure event instead of a phantom success. The nav events should also now carry the document title.

- **Message bridge + `injectedJavaScriptBeforeContentLoaded`**: both come from #3983 — the bridge is registered at document-start, so the `window.chrome.webview` / postMessage wiring exists before your page scripts run. A quick check is an inline script in the page's `<head>` that posts a message immediately: on upstream it races, with #3983 it should arrive reliably. `injectedJavaScriptBeforeContentLoaded` should likewise observe a pre-content DOM.

- **This PR (#3984)** is easiest to see by toggling an unrelated prop (e.g. `style`) on a mounted WebView — upstream re-navigates and blows away page state; with the fix it shouldn't. And `javaScriptEnabled` set at mount now survives the core-init race.

One caveat worth flagging on #3983 while you're testing it: `messagingEnabled` and `injectedJavaScriptBeforeContentLoaded` are currently captured once at CoreWebView2 init, so toggling either *after* mount only takes effect on remount. I'm treating that as something to fix rather than defend (iOS and Android both honor post-mount changes), so if your integration sets those dynamically, that's a known gap rather than a surprise.

If anything misbehaves on 0.84.1 specifically, please share the repro — these were developed against 0.83, and I'd like to know if the 0.84 composition path differs anywhere.

Waleed065:
Hey @FaithfulAudio appreciate your effort for fixing RN windows bugs , i see you've opened 4 PRs

PR1 have conflict merged with PR3

Can you create a single PR instead?

FaithfulAudio:
Thanks @Waleed065 — appreciate you digging into these! 🙏

I split them into four deliberately: each PR fixes one distinct Windows bug and is small enough to review and revert on its own —

- #3981 — explicit CoreWebView2 environment + surface init failure to JS
- #3982 — honor navigation success/failure + include title in nav events
- #3984 — `javaScriptEnabled` lost before core-init + stop re-navigating on unrelated prop changes
- #3983 — register the message bridge at document-start + `injectedJavaScriptBeforeContentLoaded`

Focused PRs like this are usually easier for maintainers to land (and to revert if needed) than one large combined change, so I'd lean toward keeping them separate unless a maintainer prefers otherwise.

On the conflict: all four report `mergeable` against `master` individually — there's no conflict with the base branch. The overlap you're seeing is only because they all touch the same WebView2 files, so whichever lands first will just need the others rebased on top. I'm happy to keep them in a clean stack and rebase as they merge (a natural order is #3981 → #3982 → #3984 → #3983). If a maintainer would rather take them as a single squashed PR, I'll gladly combine them — just say the word.

Waleed065:
Hey @FaithfulAudio Yes I'd really appreciate a single PR because i tried to fork it and merge 1-2-3-4 and 1 and 3 had conflicts. 
BTW this work of yours is really appreciatable. It has enabled me to upgrade my project easily. Thanks you so much.

FaithfulAudio:
Really glad the fixes helped you upgrade your project, @Waleed065 — that's great to hear! 🙌

On combining them into one PR: that's ultimately the maintainers' call, not mine (or ours) to make. How changes are packaged and merged is part of their review process and their purview to decide — so I don't want to collapse four reviewed PRs into one on my own initiative. If a maintainer would prefer a single squashed PR, I'll combine them right away; until then I'll keep each fix isolated and independently reviewable, which is usually what makes them easiest for maintainers to land.

For your immediate case, though, you don't need to wait on any of that. I'm happy to push a combined branch — all four fixes with the #3981/#3983 conflicts already resolved — to our fork, which you can point your project at directly. Want me to set that up? That way you're unblocked regardless of how the PRs are ultimately merged upstream.

Waleed065:
@FaithfulAudio Yes exactly, I've already installed the #4 branch as npm package. So pushing a combined branch would do just fine and I hope the maintainers merge all your work. These are good quality code pieces.

SameerTariq50:
@FaithfulAudio thanks for this

Waleed065:
Hey @FaithfulAudio 
Just to clarify — I didn't test the PR branch itself. I installed the standalone [@facilitronworks/react-native-webview-windows](https://github.com/FacilitronWorks/react-native-webview-windows) package, which is a Windows-only backport that already includes equivalent fixes to PRs #3981–#3984 if i'm not wrong. The PRs themselves haven't been tested on my end.

acoates-ms (approved):

